PR #1, opened 2022-12-05
Start return call

chambart:
Adds the `return_call` instructions. This changes quite a bit the evaluation loop


src/interpret.ml
```diff
@@ -250,9 +246,24 @@ let exec_extern_func stack (f : Value.Func.extern_func) =
...
+  let rec split_args :
+      type f r.
+      Env.t' Stack.t -> (f, r) Value.Func.atype -> Env.t' Stack.t * Env.t' Stack.t =
+   fun stack ty ->
+    match ty with
+    | Value.Func.Arg (_arg, args) ->
+      let elt, stack = Stack.pop stack in
+      let elts, stack = split_args stack args in
+      elt :: elts, stack
+    | NArg (_, _arg, args) ->
```
zapashcanon, line 258: Can you merge the first two cases ?

src/interpret.ml
```diff
@@ -282,63 +294,204 @@ let exec_extern_func stack (f : Value.Func.extern_func) =
...
+  let return (state : exec_state) =
+    let args = Stack.keep state.stack (List.length state.func_rt) in
+    match state.return_state with
+    | None -> raise (Result args)
+    | Some state ->
+      let stack = args @ state.stack in
+      { state with stack }
+
+  let branch (state : exec_state) n =
+    let rec drop_n l n =
```
zapashcanon, line 338: Can you put this in the `Stack` module ?

zapashcanon: Hum, actually you added `let rec drops s n =` in `Stack`, so you can use it directly (but I believe `drop_n` to be a better name)

src/stack.ml
```diff
@@ -156,3 +156,10 @@ let pop_n s n =
   (List.filteri (fun i _hd -> i < n) s, List.filteri (fun i _hd -> i >= n) s)
 
 let keep s n = List.filteri (fun i _hd -> i < n) s
+
+let rec drops s n =
```
zapashcanon, line 160: `drop_n` instead of `drops` ?

src/interpret.ml
chambart: Not really, GADT stuffs...

src/stack.ml
chambart: Yes

PR #19, opened 2023-03-02
allow to link two modules from two different files using their wasm name e.g.: `(module $name)`

zapashcanon:
Otherwise, if you have a file `wasm1.wast` :

```wast
(module $wasm1
  (export ...)
)
```

Which is added to your link state. When you try to link a second module `wasm2.wast`:

```wast
(module $wasm2
  (import "wasm1" ...)
)
```

You get a failure saying `unknown module wasm1`. This was discovered @epatrizio and the current patch seems to fix the issue for him.

@chambart does it looks good to you ? You know this part of the code better than me.


chambart (approved):
Looks good to me

PR #21, opened 2023-03-10
Various improvements

chambart:
* Avoid useless work when logging
* Profile mode
* Build menhir parser in table mode (faster build)


zapashcanon:
Thanks ! :heart:

PR #23, opened 2023-03-15
Optimize module

epatrizio:


src/compile.mli
```diff
@@ -1,9 +1,23 @@
 (** Utility functions to compile a module until a given step. *)
 
+val ( let* ) :
```
zapashcanon, line 3: Could you remove this ?

src/compile.mli
```diff
@@ -1,9 +1,23 @@
...
+    ('a, 'b) result
+ -> ('a -> ('c, 'b) result)
+ -> ('c, 'b) result
+
+val until_simplify :
```
zapashcanon, line 8: This doesn't look like it has been formatted. Could you run `dune fmt` ?

src/optimize.ml
```diff
@@ -0,0 +1,89 @@
...
+  match value with
+  | Imported _ -> f
+  | Local f ->
+    let value = Local (optimize_func f) in
+    { value; index}
+
+let optimize_funcs funs =
+  Named.map optimize_runtime_func funs
+
+let modul (m : modul) : modul =
```
zapashcanon, line 87: Could you write an `optimize.mli` file ? We only want to expose the `modul` function, not all the internal stuff that comes before ;)

src/optimize.ml
```diff
@@ -0,0 +1,89 @@
...
+      optimize_expr (I32_const (Int32.mul x y) :: tl)
+  | I32_const x :: I32_const y :: I_binop (S32, Div U) :: tl ->
+      optimize_expr (I32_const (Int32.div x y) :: tl)
+  | I64_const x :: I64_const y :: I_binop (S64, Add) :: tl ->
+      optimize_expr (I64_const (Int64.add x y) :: tl)
+  | I64_const x :: I64_const y :: I_binop (S64, Sub) :: tl ->
+      optimize_expr (I64_const (Int64.sub x y) :: tl)
+  | I64_const x :: I64_const y :: I_binop (S64, Mul) :: tl ->
+      optimize_expr (I64_const (Int64.mul x y) :: tl)
+  | I64_const x :: I64_const y :: I_binop (S64, Div U) :: tl ->
```
zapashcanon, line 20: I believe it could be possible to do something like:

```ocaml
| ((I32_const _ | I64_const _) as x) :: ((I32_const _ | I64_const _) as y) :: I_binop (nn, op) :: tl ->
  let result = Interpret.exec_ibinop [] nn op in
  begin
    match result with
    | [I32_const _ | I64_const _ as result ] -> result :: optimize_expr tl
    | _ -> assert false
  end
```

Could you try if this works ? This would make the code much shorter as you'll have a single match case for all theses binary operators cases :)

src/optimize.ml
```diff
@@ -0,0 +1,89 @@
...
+      optimize_expr (I64_const (Int64.div x y) :: tl)
+  (* | I32_const (Int32.zero) ::  :: tl -> EQZ
```
zapashcanon, line 22: What is this ?

src/optimize.ml
```diff
@@ -0,0 +1,89 @@
...
+  in
+  let str_idx_param (idx : int) (p : param) : string  =
+    let str_opt, _ = p in
+      match str_opt with
+      | None -> string_of_int idx
+      | Some s -> s
+  in
+  let local_find (idx : int) (p : param) : unit =
+    match Hashtbl.find_opt loc_hashtbl idx with
+    | None -> raise (Failure ("function " ^ str_func_id ^ ": local " ^ (str_idx_param idx p) ^ " not used"))
```
zapashcanon, line 64: I'm not found of failing in this case. I believe the way to go is to define a `val remove_local : int -> expr -> expr` that will remove the local `n`, to do so, you leave all references to local `m` unchanged when `m < n`, and replace `m` by `m - 1` when `m > n`.

You'll need to fold on the body with this function for every unused local.

src/optimize.ml
epatrizio: I tried to build the following optimization :
.eqz instruction after const 0 is always true

src/optimize.ml
```diff
@@ -0,0 +1,123 @@
...
+let locals_func (body : expr) : (int, unit) Hashtbl.t =
+  let locals_hashtbl : (int, unit) Hashtbl.t = Hashtbl.create 10 in
+  let rec aux_instr (i : instr) : unit =
+    match i with
+    | Local_get ind | Local_set ind | Local_tee ind ->
+        Hashtbl.add locals_hashtbl ind ()
+    | Block (_, _, e) -> aux_expr e
+    | Loop (_, _, e) -> aux_expr e
+    | If_else (_, _, e1, e2) -> aux_expr e1; aux_expr e2
+    | _ -> ()
```
zapashcanon, line 68: This is going to be cumbersome, but it would probably be a good idea to list all cases here, so that when new cases are added in the future, we don't forget to go through them if it's needed.

src/optimize.ml
```diff
@@ -0,0 +1,123 @@
...
+  let remove_local (idx : int) (body : expr) : expr =
+    let rec aux_instr (i : instr) : instr =
+      match i with
+      | Local_get ind when nb_params <= idx && idx < ind -> Local_get (ind-1)
+      | Local_set ind when nb_params <= idx && idx < ind -> Local_set (ind-1)
+      | Local_tee ind when nb_params <= idx && idx < ind -> Local_tee (ind-1)
+      | Block (m, t, e) -> Block (m, t, aux_expr e)
+      | Loop (m, t, e) -> Loop (m, t, aux_expr e)
+      | If_else (m, t, e1, e2) -> If_else (m, t, aux_expr e1, aux_expr e2)
+      | _ -> i
```
zapashcanon, line 87: Same as above about listing all cases.

src/optimize.ml
```diff
@@ -0,0 +1,293 @@
...
+  | ((I32_const _ | I64_const _) as x)
+    :: ((I32_const _ | I64_const _) as y)
+    :: I_relop (nn, op)
+    :: tl ->
+    let result =
+      Interpret.exec_irelop [ Value.of_instr y; Value.of_instr x ] nn op
+    in
+    begin
+      match result with
+      | [ ((I32 _ | I64 _) as result) ] ->
```
zapashcanon, line 69: This can only be an `I32 _`.

src/optimize.ml
```diff
@@ -0,0 +1,293 @@
...
+  | ((F32_const _ | F64_const _) as x)
+    :: ((F32_const _ | F64_const _) as y)
+    :: F_relop (nn, op)
+    :: tl ->
+    let result =
+      Interpret.exec_frelop [ Value.of_instr y; Value.of_instr x ] nn op
+    in
+    begin
+      match result with
+      | [ ((F32 _ | F64 _) as result) ] ->
```
zapashcanon, line 82: This is not going to be a float but always an `I_32 _`.

src/optimize.ml
```diff
@@ -0,0 +1,293 @@
...
+        optimize_expr (Value.to_instr result :: tl)
+      | _ -> assert false
+    end
+  | (I32_const _ | I64_const _ | F32_const _ | F64_const _) :: Drop :: tl ->
+    optimize_expr tl
+  | Local_set x :: Local_get y :: tl when x = y ->
+    optimize_expr (Local_tee x :: tl)
+  | (Br _ as br) :: _tl -> [ br ]
+  | (I32_const c as const) :: (Br_if _ as brif) :: tl -> begin
+    match c with 0l -> optimize_expr tl | _ -> [ const; brif ]
```
zapashcanon, line 92: You can replace `[ const; brif ]` by `[ Br label ]` (you'll need to name the label when matching on `Br_if`).

test/main.ml
```diff
@@ -14,7 +14,7 @@ let test_file f =
   Format.printf "testing file     : `%a`... " Fpath.pp f;
   match Owi.Parse.from_file ~filename:(Fpath.to_string f) with
   | Ok script -> begin
-    match Owi.Script.exec script with
+    match Owi.Script.exec script ~optimize:true with
```
zapashcanon, line 17: It would be nice to also launch the test without the optimization.

test/main.ml
```diff
@@ -29,23 +31,30 @@ let test_file f =
...
   | Ok l ->
     List.iter
       (fun file ->
         incr count_total;
-        match test_file file with
-        | Ok () -> ()
+        match test_file ~optimize:false file with
+        | Ok () -> begin
+          incr count_total;
+          match test_file ~optimize:true file with
```
zapashcanon, line 43: You're running this test only if the previous one was successful. It would be better to run it in all cases.

test/main.ml
```diff
@@ -29,23 +31,33 @@ let test_file f =
...
+            incr count_total_failed
+        end )
       (List.sort compare l);
     if !count_error > 0 then
-      Error (Format.sprintf "%d test failed !" !count_error)
+      Error (Format.sprintf "%d test failed" !count_error)
     else Ok ()
   | Error (`Msg e) ->
-    Format.eprintf "error      : %s@." e;
+    pp_error Format.std_formatter e;
```
zapashcanon, line 60: This should be `Format.err_formatter`. (Same on the other lines)

zapashcanon:
Thanks ! :)

PR #28, opened 2023-04-25
Fuzzing

epatrizio:


test/fuzz/env.ml
```diff
@@ -0,0 +1,58 @@
...
+  let name = Format.sprintf "g%d" n in
+  let globals = (name, typ) :: !v.globals in
+  let next_global = succ n in
+  v := { !v with next_global; globals };
+  name
+
+let add_local typ =
+  let n = !v.next_local in
+  let name = Format.sprintf "l%d" n in
+  let locals = (Some name, typ) :: !v.locals in
```
zapashcanon, line 41: You can replace `Some name` by `name`. We'll always use a name to make things easier.

test/fuzz/gen.ml
```diff
@@ -192,6 +234,7 @@ let func env =
   MFunc { type_f; locals; body; id }
 
 let fields env =
+  let memories = map ([memory env]) (fun m -> [m]) in (* for now, just one memory *)
```
zapashcanon, line 237: We should use an option for now. It'll be easier. :)

test/fuzz/gen.ml
```diff
@@ -173,6 +211,10 @@ let rec expr ~block_type ~stack ~locals env =
     let i = const i in
     list_cons i next
 
+let memory env =
+  let+ id = const @@ Env.add_memory env in
+  MMem (Some id, { min = 0; max = None})
```
zapashcanon, line 216: We should have a random min (but not too big !) and generate an (int option) for the max.

test/fuzz/basic.ml
```diff
@@ -336,3 +336,95 @@ let block_type env =
...
+let memory_fill = const Memory_fill
+
+let memory_init = const (Memory_init (Raw 0))
+
+let memarg =
+  let* offset = range 10 in
+  let+ align = range 10 in
+  { offset; align}
+
+let i32_load_i32 =
```
zapashcanon, line 355: Could you rename the function `i32_load` and do the same for all the following functions ? They're all going to use an i32, which is the memory offset. I'm not sure it makes thing clearer to have it in the name.

test/fuzz/basic.ml
```diff
@@ -336,3 +336,95 @@ let block_type env =
...
+
+let memory_copy = const Memory_copy
+
+let memory_fill = const Memory_fill
+
+let memory_init = const (Memory_init (Raw 0))
+
+let memarg =
+  let* offset = range 10 in
+  let+ align = range 10 in
```
zapashcanon, line 352: Why 10 ?

test/fuzz/basic.ml
```diff
@@ -0,0 +1,430 @@
...
+let global_set_f32 = global_set F32
+
+let global_set_f64 = global_set F64
+
+let local ntyp env =
+  let locals = Env.get_locals ntyp env in
+  List.map
+    ( fun (name, _) -> match name with
+      | Some n -> pair (const (Local_get (Symbolic n))) (const [ S.Push (Num_type ntyp) ])
+      | None -> assert false )
```
zapashcanon, line 273: If this can not happen, then maybe we should change the type of locals in the environment ?

test/fuzz/basic.ml
```diff
@@ -0,0 +1,430 @@
...
+let local_set_f32 = local_set F32
+
+let local_set_f64 = local_set F64
+
+let local_tee ntyp env =
+  let locals = Env.get_locals ntyp env in
+  List.map
+    ( fun (name, _) -> match name with
+      | Some n -> pair (const (Local_tee (Symbolic n))) (const [ S.Nothing ])
+      | None -> assert false )
```
zapashcanon, line 305: Here too.

test/fuzz/env.ml
```diff
@@ -0,0 +1,80 @@
...
+  let name = Format.sprintf "f%d" n in
+  env.next_fun <- succ n;
+  env.funcs <- (name, typ) :: env.funcs;
+  name
+
+let get_globals ntyp env ~only_mut =
+  let is_typ global =
+    let _, (m, v) = global in
+    match v with
+    | Num_type nt -> nt = ntyp && (not only_mut || (only_mut && m = Owi.Types.Var))
```
zapashcanon, line 62: Equivalent to the shorter:

```ocaml
| Num_type nt-> nt = ntyp &&  (not only_mut || m = Owi.Types.Var)
```

:)

test/fuzz/fuzzer.ml
```diff
@@ -0,0 +1,70 @@
...
+  let result_unoptimized =
+    let open Owi_unoptimized in
+    of_symbolic m |> run
+  in
+  let result_optimized =
+    let open Owi_optimized in
+    of_symbolic m |> run
+  in
+  (*
+  let _result_reference =
```
zapashcanon, line 14: I believe when can enable this now ?

test/fuzz/interprets.ml
```diff
@@ -0,0 +1,64 @@
...
+    let suffix = ".wast" in
+    let tmp_file = Filename.temp_file prefix suffix in
+    let chan = open_out tmp_file in
+    let fmt = Format.formatter_of_out_channel chan in
+    Format.pp_print_string fmt modul;
+    close_out chan;
+    let n = Sys.command @@ Format.sprintf "officialwasm %s" tmp_file in
+    match n with
+    | 0 -> Ok ()
+    | 42 -> Error "trap"
```
zapashcanon, line 62: Could you open an issue on the reference interpreter saying that we would like a different exit code for a validation error and a runtime trap ? Then add a link to the issue in a comment in the code, so that we'll use the correct number here in the future.

test/fuzz/basic.ml
epatrizio: Same with `_store_` functions

test/fuzz/basic.ml
epatrizio: I don't know if it's ok to generate unconstrained integers.
I chose small integers arbitrarily (10). I try to remove the constraint ...

test/fuzz/basic.ml
epatrizio: It's about `None -> assert false` ?
The reason is that locales are always generated with a symbolic name

test/fuzz/env.ml
epatrizio: of course ;)

test/fuzz/basic.ml
zapashcanon: Yes. I believe we should change the type of locals to be `(string * typ) list` or something similar instead of `param list` to avoid having to handle the `None` case everywhere.

zapashcanon: Yes. I believe we should change the type of locals in the environment to be `(string * typ) list` or something similar instead of `param list` to avoid having to handle the `None` case everywhere.

test/fuzz/basic.ml
```diff
@@ -13,6 +13,17 @@ let val_type =
...
 
+let symbolic =
+  let+ str = bytes in
+  (Symbolic str : indice)
+
+let raw =
+  let+ i = int in
+  (Raw i : indice)
+
+let indice =
```
zapashcanon, line 24: i don't think we're ever going to really need any of these three functions. You should remove them. (I explain a little bit more below).

test/fuzz/basic.ml
```diff
@@ -348,83 +359,101 @@ let memory_fill = const Memory_fill
...
 
-let i64_store32_i32 =
+let i64_store32 =
   let+ memarg in
   I64_store32 memarg
+
+let data_active =
+  let* ind = option indice in   (* is full indice gen OK ? bug offset label ? *)
+  let* vt = val_type in
+  let+ inst = const_of_val_type vt in   (* only i32 ? *)
```
zapashcanon, line 446: This is indeed always an `int32`. It's the offset at which it'll be in the memory. You should make sure that the length of the data + the offset is lower than the memory initial memory size (you'll have to store this in the environment).

test/fuzz/gen.ml
epatrizio: Mémo : warning, time explosion during fuzzing if memory size is fully random.
(constrained size for now)

test/fuzz/basic.ml
epatrizio: of course, I didn't understand ;)

test/fuzz/basic.ml
epatrizio: Memo: For now, adding this data/memory size constraint does not seem to be a "must have"

test/fuzz/basic.ml
```diff
@@ -459,20 +437,20 @@ let i64_store32 =
...
-  let+ inst = const_of_val_type vt in   (* only i32 ? *)
-  let exp = [ inst ] in
-  Data_active (ind, exp)
+let data_active (env : Env.t) =
+  match env.memory with
+  | Some name ->
+    let+ inst = const_i32 in
+    let exp = [ inst ] in
+    Data_active (Some (Symbolic name), exp)
+  | None -> assert false
```
zapashcanon, line 446: You should rewrite this to avoid the `assert false` case.

test/fuzz/env.ml
```diff
@@ -39,10 +37,8 @@ let add_data env =
   name
 
 let add_memory env =
-  let n = env.next_memory in
-  let name = Format.sprintf "m%d" n in
-  env.memories <- name :: env.memories;
-  env.next_memory <- succ n;
+  let name = "m0" in
```
zapashcanon, line 40: You should keep the `next_memory` stuff. :)

test/fuzz/env.ml
```diff
@@ -39,10 +37,8 @@ let add_data env =
...
+  env.memory <- Some name;
```
zapashcanon, line 41: You should check that it's `None` and fail otherwise.

test/fuzz/fuzzer.ml
```diff
@@ -50,21 +46,32 @@ let check_optimized m =
...
+let debug = ref false
+
+let options = [
+  "--debug", Arg.Set debug, " Debug mode (modul printing)"
+]
+
+(* let in_file_name = ref "" *)
+(* let set_file _s = print_endline "FFFFFF" *)
+
+let usage = "usage: dune exec -- ./test/fuzz/fuzzer.exe [--option]"
```
zapashcanon, line 64:
```ocaml
let usage = Format.sprintf "usage: dune exec -- %s [--option]" Sys.argv.(0)
```

test/fuzz/gen.ml
```diff
@@ -139,28 +139,37 @@ let expr_available_2_f64 =
...
+    
     let instr = If_else (None, Some typ1, expr_then, expr_else) in
     let pt_descr = S.Pop :: List.map (fun _ -> S.Pop) pt in
     let rt_descr = List.map (fun t -> S.Push t) rt in
     pair (const instr) (const (pt_descr @ rt_descr))
   end
   | _ -> assert false
 
-let rec expr ~block_type ~stack ~locals env =
+let rec expr ~block_type ~stack ~locals ~start env =
```
zapashcanon, line 172: I'm not convinced by this `start` thing. This is to avoid infinite loops, right ?

src/types.ml
```diff
@@ -528,8 +528,14 @@ struct
 
     (* TODO: when offset is 0 then do not print anything, if offset is N (memargN) then print nothing ? *)
     let memarg fmt { offset; align } =
+      let rec pow_2 n =
```
zapashcanon, line 531: As we're only computing `pow_2 x` and not `pow x n` we can do something better than the recursive algorithm. Hint: think about how bits (base 2) and power of 2 are related. You can do it in `O(1)`.

test/fuzz/gen.ml
epatrizio: `start` is a boolean parameter just to say: consider function calls only in the $start function
I don't know if it's the right way to do it ?

src/types.ml
epatrizio: `lsl` operator ;)

test/fuzz/gen.ml
zapashcanon: Why would we only call functions in the `$start` function ?

epatrizio: An initial restriction, to avoid recurcive calls

test/fuzz/gen.ml
```diff
@@ -283,6 +283,9 @@ let fields env =
   let* globals = list (global env) in
   let* funcs = list (func env) in
   let+ start_code =
+    let* () = const () in
```
zapashcanon, line 286: I don't understand this.

src/types.ml
```diff
@@ -526,16 +526,18 @@ struct
...
+        1 lsl n
       in
-      if offset = 0 && align = 0 then ()
-      else Format.fprintf fmt "offset=%d align=%d" offset (pow_2 align)
+      let off =
+        if offset = 0 then "" else Printf.sprintf "offset=%d" offset
+      in
+      let al =
+        Printf.sprintf "align=%d" (pow_2 align)
+      in
```
zapashcanon, line 539: can we avoid allocating the intermediate strings ?

test/fuzz/crowbarplus.ml
```diff
@@ -0,0 +1,6 @@
+include Crowbar
+
+(* TODO: PR to crowbar with this function ? *)
+let list_append l1 l2 = map [ l1; l2 ] List.append
+
+let list_cons hd tl = map [ hd; tl ] List.cons
```
zapashcanon, line 6: I think we can remove this file now ? :)

test/fuzz/syntax.ml
```diff
@@ -0,0 +1,5 @@
+let ( let* ) = Crowbarplus.dynamic_bind
+
+let ( let+ ) gen map_fn = Crowbarplus.map [ gen ] map_fn
+
+let ( and+ ) = Crowbarplus.pair
```
zapashcanon, line 5: We can also remove this one (you'll have to clone crowbar and run `dune install` until it's available on opam).

test/fuzz/fuzzer.ml
epatrizio: For now, OCaml Arg module could not be used (conflict with crowbar)
So, new hard coded parameters have been added (param.ml)

test/fuzz/fuzzer.ml
```diff
@@ -30,13 +30,10 @@ let check_optimized m =
...
   in
-  let result2 = true in
-  (* todo : intégrer le code de l'interprete de ref.
-     Nb. il va y avoir des bug sur le print, et donc cela va falloir voir/corriger
-      exemple : global ('mut' i32) -> enlever les ()
-  *)
-  (* let result2 =
-    match (result_unoptimized, result_reference) with
+  let result2 =
+    if not Param.reference_fuzzing then true
```
zapashcanon, line 34:
```ocaml
not Param.reference_fuzzing
||
match ...
```

test/fuzz/basic.ml
```diff
@@ -477,6 +477,19 @@ let expr_call (env : Env.t) (stack : val_type list) =
       | _ -> None )
     env.funcs
 
+let expr_br_if (env : Env.t) (stack : val_type list) =
+  List.filter_map
+    (fun (name, bt) ->
+      match bt with
+      | Arg.Bt_raw (_, (_pt, rt))
+        when S.is_stack_compatible (List.tl stack) (List.rev rt) ->
```
zapashcanon, line 485: This is going to fail when the stack is empty (of course this should not happen).

You could rewrite this like:

```ocaml
let expr_br_if (env : Env.t) (stack : val_type list) =
  match stack with
  | [] -> []
  | _hd::tl ->
  List.filter_map
    (fun (name, bt) ->
      match bt with
      | Arg.Bt_raw (_, (_pt, rt))
        when S.is_stack_compatible tl (List.rev rt) -> ...
```

This also avoid doing `List.tl` many times.

test/fuzz/gen.ml
epatrizio: `start` parameter has been removed.
Now, function calls (but not recursive) can be made anywhere.

test/fuzz/basic.ml
```diff
@@ -477,21 +477,22 @@ let expr_call (env : Env.t) (stack : val_type list) =
       | _ -> None )
     env.funcs
 
-let expr_br_if (env : Env.t) (stack : val_type list) =
+let expr_br_if (env : Env.t) (stack : val_type list) (is_block : bool) =
```
zapashcanon, line 480: Could you define a something like:

```ocaml
type label_kind =
| Loop
| Block
| Func (* I'm not sure this will be needed *)
```

and use this later in the code rather than the boolean ?

test/fuzz/basic.ml
```diff
@@ -500,18 +501,20 @@ let random_stack =
 let unreachable =
   pair (const Unreachable) random_stack
 
-let expr_br (env : Env.t) (stack : val_type list) =
+let expr_br (env : Env.t) (stack : val_type list) (is_block : bool) =
```
zapashcanon, line 504: Same here.

test/fuzz/env.ml
```diff
@@ -8,11 +8,13 @@ type t =
   ; mutable next_fun : int
   ; mutable next_local : int
   ; mutable next_block : int
+  ; mutable next_loop : int
   ; mutable datas : string list
   ; mutable memory : string option
   ; mutable globals : (string * global_type) list
   ; mutable locals : (string * val_type) list
   ; mutable blocks : (string * block_type) list
+  ; mutable loops : (string * block_type) list
```
zapashcanon, line 17: Why do you separate `blocks` and `loops` ?

test/fuzz/fuzzer.ml
```diff
@@ -1,49 +1,73 @@
+exception Timeout
+
 let () = Random.self_init ()
 
+let timeout_call_run (run : 'a -> unit Owi.Result.t) (m : 'a) =
```
zapashcanon, line 5: the `run` parameter should have type `unit -> 'a` and `timeout_call_run` should not need the `m` parameter. Then you would do:

```ocaml
  let result_unoptimized =
    let open Owi_unoptimized in
    let m = of_symbolic m in
    timeout_call_run (fun () -> run m)
  in
```

test/fuzz/gen.ml
```diff
@@ -44,7 +45,8 @@ let expr_available_1_i32 if_else expr ~locals ~stack env =
   ]
   @ B.local_set_i32 env @ B.local_tee_i32 env @ B.global_set_i32 env
   @ if B.memory_exists env then B.memory_grow :: load_instr else []
-  @ B.expr_br_if env stack
+  @ B.expr_br_if env stack true
+  @ B.expr_br_if env stack false
```
zapashcanon, line 49: You should be able to call this only once and make it generate everything that is needed.

test/fuzz/gen.ml
```diff
@@ -221,10 +237,11 @@ let rec expr ~block_type ~stack ~locals env =
       | _ -> []
     in
     let expr_available env =
-      expr_always_available block expr ~locals ~stack env
+      expr_always_available block loop expr ~locals ~stack env
       @ expr_available_with_current_stack
       @ B.expr_call env stack
-      @ B.expr_br env stack
+      @ B.expr_br env stack true
+      @ B.expr_br env stack false
```
zapashcanon, line 244: Same here, you should be able to call it only once.

test/fuzz/param.ml
```diff
@@ -1,5 +1,7 @@
 let debug = true
 
-let reference_fuzzing = true
+let reference_fuzzing = false
```
zapashcanon, line 3: Why did you disable this ?

test/fuzz/param.ml
```diff
@@ -1,5 +1,7 @@
...
 
 let initial_fuel = 20
+
+let max_time_execution = 1. (* seconds *)
```
zapashcanon, line 7: That's probably way to much. We should stop around `0.001` by default, otherwise running the 5000 modules is going to take forever.

test/fuzz/fuzzer.ml
```diff
@@ -1,49 +1,73 @@
...
-    | Ok (), Error msg ->
+    | Some Ok (), Some Error msg ->
       Format.ksprintf failwith
         "only reference module interpretation throws an error: %s" msg
+    | None, Some _ -> 
+      Format.ksprintf failwith
+        "only unoptimized module interpretation throws a timeout error"
+    | Some _, None ->
+      Format.ksprintf failwith
+        "only optimized module interpretation throws a timeout error"
```
zapashcanon, line 70: One nice thing would be to have a global counter so we have an idea of of many executions led to a timeout.

test/fuzz/env.ml
zapashcanon: We should have something like: `mutable blocks : (block_kind, string, block_type) list` and only this (not `loops` and `blocks`).

test/fuzz/param.ml
epatrizio: Only for my local tests, to speed up execution ;)
But the timeout process seems to work not so good with the official interpreter :(

test/fuzz/gen.ml
epatrizio: For the moment, I still have the 2 separate calls.
After `expr_br_if refactoring`, it seems ok to keep 2 separate calls ... ?

test/fuzz/env.ml
```diff
@@ -1,20 +1,22 @@
 open Owi.Symbolic
 open Crowbar
 
+type block_kind =
+  | Block
+  | Loop
```
zapashcanon, line 6: You should add `Func` and put only one of them when you start generating a function.

test/fuzz/basic.ml
```diff
@@ -501,20 +501,19 @@ let random_stack =
 let unreachable =
   pair (const Unreachable) random_stack
 
-let expr_br (env : Env.t) (stack : val_type list) (is_block : bool) =
-  let blocks_type = if is_block then env.blocks else env.loops in
+let expr_br (env : Env.t) (stack : val_type list) (bkind : Env.block_kind) =
```
zapashcanon, line 504: `bkind` should not be a parameter, you should do something link:

```ocaml
let* bkind = choose [ const Block; const Loop ] in
```

test/fuzz/basic.ml
```diff
@@ -477,22 +477,22 @@ let expr_call (env : Env.t) (stack : val_type list) =
       | _ -> None )
     env.funcs
 
-let expr_br_if (env : Env.t) (stack : val_type list) (is_block : bool) =
+let expr_br_if (env : Env.t) (stack : val_type list) (bkind : Env.block_kind) =
```
zapashcanon, line 480: Same comment as below.

test/fuzz/basic.ml
```diff
@@ -477,22 +477,22 @@ let expr_call (env : Env.t) (stack : val_type list) =
...
-      (fun (name, bt) ->
+      (fun (_, name, bt) ->
         match bt with
         | Arg.Bt_raw (_, (pt, rt)) when
-          if is_block then S.is_stack_compatible tl (List.rev rt)
-          else S.is_stack_compatible_param tl (List.rev pt) ->
-            Some ( pair
+          match bkind with
+          | Env.Block -> S.is_stack_compatible tl (List.rev rt)
+          | Env.Loop -> S.is_stack_compatible_param tl (List.rev pt)
```
zapashcanon, line 490: I don't understand how this compile. I guess you're missing a `when` instead of the `->`.

Anyway, you should add a `| Env.Func` case, that would branch on a `Raw` indice (a `Symbolic` won't work). In order to get the indice you must do something like `List.length @@ Env.get_blocks env`.

test/fuzz/basic.ml
```diff
@@ -501,20 +501,19 @@ let random_stack =
...
       match bt with
       | Arg.Bt_raw (_, (pt, rt)) when
-        if is_block then S.is_stack_compatible stack (List.rev rt)
-        else S.is_stack_compatible_param stack (List.rev pt) ->
-        Some
-          (pair
+        match bkind with
+        | Env.Block -> S.is_stack_compatible stack (List.rev rt)
+        | Env.Loop -> S.is_stack_compatible_param stack (List.rev pt)
+        -> Some ( pair
```
zapashcanon, line 512: Same comment as above.

test/fuzz/gen.ml
```diff
@@ -4,14 +4,14 @@ open Owi.Symbolic
...
+let expr_always_available block expr ~locals ~stack env =
   [ pair B.const_i32 (const [ S.Push (Num_type I32) ])
   ; pair B.const_i64 (const [ S.Push (Num_type I64) ])
   ; pair B.const_f32 (const [ S.Push (Num_type F32) ])
   ; pair B.const_f64 (const [ S.Push (Num_type F64) ])
   ; pair (const Nop) (const [ S.Nothing ])
-  ; block expr ~locals ~stack env
-  ; loop expr ~locals ~stack env
+  ; block Env.Block expr ~locals ~stack env
+  ; block Env.Loop expr ~locals ~stack env
```
zapashcanon, line 14: You should have only one call to `block` and choose the kind inside the `block` function.

test/fuzz/gen.ml
```diff
@@ -45,8 +45,8 @@ let expr_available_1_i32 if_else expr ~locals ~stack env =
   ]
   @ B.local_set_i32 env @ B.local_tee_i32 env @ B.global_set_i32 env
   @ if B.memory_exists env then B.memory_grow :: load_instr else []
-  @ B.expr_br_if env stack true
-  @ B.expr_br_if env stack false
+  @ B.expr_br_if env stack Env.Block
+  @ B.expr_br_if env stack Env.Loop
```
zapashcanon, line 49: Same comment as above.

test/fuzz/gen.ml
```diff
@@ -237,11 +227,11 @@ let rec expr ~block_type ~stack ~locals env =
...
     in
     let expr_available env =
-      expr_always_available block loop expr ~locals ~stack env
+      expr_always_available block expr ~locals ~stack env
       @ expr_available_with_current_stack
       @ B.expr_call env stack
-      @ B.expr_br env stack true
-      @ B.expr_br env stack false
+      @ B.expr_br env stack Env.Block
+      @ B.expr_br env stack Env.Loop
```
zapashcanon, line 234: Same here.

test/fuzz/gen.ml
```diff
@@ -156,30 +157,19 @@ let if_else expr ~locals ~stack env =
   end
   | _ -> assert false
 
-let block expr ~locals ~stack env =
+let block block_kind expr ~locals ~stack env =
```
zapashcanon, line 160: You should split this into a `block` and a `loop` functions.

test/fuzz/gen.ml
```diff
@@ -136,9 +136,10 @@ let expr_available_2_f64 =
 let if_else expr ~locals ~stack env =
```
zapashcanon, line 136: You should add this to the list of blocks. It's valid to branch to an `if`, see:

```wast
(module

  (func $f
    i32.const 1
    (if $a (result i32)
      (then
        i32.const 42
        br $a
      )
      (else
        i32.const 999
      )
    )
    drop
  ))
  ```

File: test/fuzz/basic.ml (hunk 37 above)
Comment (zapashcanon) [reply test/fuzz/basic.ml:490]: Oh indeed, sorry. This is a little bit confusing:

```ocaml
| Env.Loop -> S.is_stack_compatible_param stack (List.rev pt)
-> ...
```


Maybe you should remove the `when` and do something like:

```ocaml
match bt with
  | Arg.Bt_raw (_, (pt, rt)) ->
    let is_stack_compatible = match bkind with
    | Env.Block -> S.is_stack_compatible stack (List.rev rt)
    | Env.Loop -> S.is_stack_compatible_param stack (List.rev pt)
    in
    if not is_stack_compatible then None
    else ...
```

test/fuzz/gen.ml
```diff
@@ -267,7 +267,9 @@ let func env =
   Env.refill_fuel env;
   let* locals = list (local env) in
   let* type_f = B.block_type env in
+  let _ = Env.add_block env type_f Env.Func in
```
zapashcanon, line 270: Please don't use `let _`. :-)

Either `let () = ...` or `let (_meaningful_name : with_the_type) = ...` depending on if it's unit or not.

It will avoid some trouble if you ever change the arity of `Env.add_block` and then only partially apply the function.

test/fuzz/basic.ml
```diff
@@ -12,14 +12,57 @@ type num_size =
...
+let param = pair (const (None : string option)) val_type
+
+let func_type = pair (list param) (list val_type)
+
+let str_type =
+  let+ func_type in
+  Def_func_t func_type
+(* TODO: complete Def_struct_t / Def_array_t *)
+
+let sub_type = map [ final; str_type ] (fun f st -> (f, ([] : indice list), st))
```
zapashcanon, line 52: You could do:

```ocaml
let* final in
let+ str_type in
...
```

no ?

test/fuzz/basic.ml
```diff
@@ -12,14 +12,57 @@ type num_size =
...
 
+let val_storage_t =
+  let+ val_type in
+  Val_storage_t val_type
+
+let val_packed_t =
+  let+ packed_type in
+  Val_packed_t packed_type
+
+let storage_type = choose [ val_storage_t; val_packed_t ]
```
zapashcanon, line 64: I don't think we're going to need this. (We can keep it for later when we'll fuzz the GC proposal if you want :-))

test/fuzz/basic.ml
epatrizio: Yes, of course, sorry, I was focused on the equivalent of the `pair` function (~ `triplet`) which doesn't exist in the standard version in Crowbar.

test/fuzz/basic.ml
epatrizio: for the moment I don't use these generators ;)

test/fuzz/basic.ml
```diff
@@ -514,6 +514,56 @@ let data_drop (env : Env.t) =
...
+    env.elems
+
+let table_init (env : Env.t) =
+  List.flatten @@
+  List.map
+    (fun (name_x, _) -> 
+      List.map (fun (name_y, _) -> pair (const (Table_init (Symbolic name_x, Symbolic name_y) )) (const [ S.Pop; S.Pop; S.Pop ]))
+      env.tables
+    )
+    env.tables
```
zapashcanon, line 529: I'm not sure I understand what you're doing here. :-)

The `table.init` instruction expects a table index and an element index.

For instance, in `table.init $t_i $e_i`, `$t_i` must be a table and `$e_i` must be an element.

Here your are using two table indices !

test/fuzz/basic.ml
```diff
@@ -514,6 +514,56 @@ let data_drop (env : Env.t) =
...
+(* TODO: to be continued > add constraint between table $name_x and $name_y *)
+
+let table_copy (env : Env.t) =
+  List.flatten @@
+  List.map
+    (fun (name_x, _) -> 
+      List.map (fun (name_y, _) -> pair (const (Table_copy (Symbolic name_x, Symbolic name_y) )) (const [ S.Pop; S.Pop; S.Pop ]))
+      env.tables
+    )
+    env.tables
```
zapashcanon, line 539: Let's rewrite this in another way.

What about something like:

```ocaml
let tables = List.map const env.tables in
let* t1 = choose tables in
let* t2 = choose tables in
pair (
  const (
    Table_copy (
      Symbolic t1,
      Symbolic t2)))
(const [S.Pop; S.Pop; S.Pop])
```

(untested :-))

test/fuzz/basic.ml
```diff
@@ -29,10 +29,16 @@ let limits =
...
+(* TODO: complete No_final *)
 
-let elem_mode = const Elem_declarative (* for functions *)
-(* TODO: complete Elem_active / Elem_passive *)
+let elem_active =
+  let* ind = const (None : indice option) in
+  let+ expr = const [] in
+  Elem_active (ind,expr)
+
+let elem_mode = const Elem_passive
```
zapashcanon, line 40: I'm not sure we want to do it this way.

You should have an `elem` generator that would simply be:

```ocaml
let elem =
  choose [ elem_active; elem_passive; elem_declarative ]
```

epatrizio: Yes, of course. But for the moment, i consider only passive elements.

test/fuzz/basic.ml
epatrizio: Is it to avoid `List.flatten` (due to nested `List.map`) ?

epatrizio: Type of this instruction is `(instr * S.stack_op list) gen` but it's expected `(instr * S.stack_op list) gen list`

test/fuzz/basic.ml
epatrizio: in the commit below [c22f23b](https://github.com/OCamlPro/owi/pull/28/commits/c22f23b88e268a2bcf33f6db09607814a2ed51ac), I've added `elem_active`

test/fuzz/basic.ml
epatrizio: Done : https://github.com/OCamlPro/owi/pull/28/commits/4b296e6fea835931179ca46856dbd243f15c5534
In `table_copy` function, is it possible to avoid `if rt_x = rt_y then ... else pair (const (Nop)) (const [ S.Nothing ])` ?
I try `option ...` without success

zapashcanon: I believe this is fine for now. Could you simply add a `TODO` explaining that this should be removed at some point ? :)

test/fuzz/gen.ml
zapashcanon: Could you add a TODO with a link to https://github.com/OCamlPro/owi/pull/28#discussion_r1212835761 ?

test/fuzz/basic.ml
zapashcanon: Could you add a TODO with a link to https://github.com/OCamlPro/owi/pull/28#discussion_r1212838601 ?

test/fuzz/interprets.ml
zapashcanon: Could you add a TODO with a link to https://github.com/OCamlPro/owi/pull/28#discussion_r1212866678 ?

test/fuzz/basic.ml
zapashcanon: Could you add a TODO with a link to https://github.com/OCamlPro/owi/pull/28#discussion_r1213333308 ?

test/fuzz/fuzzer.ml
zapashcanon: Could you add a TODO to param.ml with a link to https://github.com/OCamlPro/owi/pull/28#discussion_r1231983824 ?

test/fuzz/basic.ml
epatrizio: This has been fixed but not referenced here, sorry.

test/fuzz/basic.ml
epatrizio: This has been fixed but not referenced here, sorry.

zapashcanon:
Thanks ! (And congratulations :D)

PR #31, opened 2023-05-23
Add `call_ref` and `return_call_ref`

chambart:
Note that the added tests don't pass right now due to an unimplemented part while typechecking `local_set`
See https://github.com/OCamlPro/owi/blob/090287833ae6ab1b47e902eca9ed9a616bcaee17/src/simplify.ml#L123-L126C1


zapashcanon:
Fix #30. Thanks !

PR #49, opened 2023-06-21
Functorized and more

chambart:
Draft


zapashcanon:
TODO for @chambart and @zapashcanon :
- [ ] add path conditions to extern
- [ ] fix globals and memory (symbolic)
- [ ] complete the functor interpret
- [ ] avoid unneeded call to the Choice monad
- [ ] remove now useless lazy

PR #53, opened 2023-09-17
Adds unsafe flag to skip typechecker

filipeom:
Attemps to introduce unsafe flag for optimized owi_sym.

This flag facilitates benchmarking on WebAssembly programs derived from C code, where prior typechecking has already occurred during compilation.

### DIFF
--- a/src/bin/owi_sym.ml
+++ b/src/bin/owi_sym.ml
@@ -114,7 +114,8 @@ let ( let*/ ) (t : 'a Result.t) (f : 'a -> 'b Result.t Choice.t) :
   'b Result.t Choice.t =
   match t with Error e -> Choice.return (Error e) | Ok x -> f x
 
-let simplify_then_link_then_run ~optimize (pc : unit Result.t Choice.t) file =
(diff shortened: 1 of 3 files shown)


zapashcanon:
Thanks !

Could you also add the flag to the `Compile.until_check` function ? (you'll have to add it to all intermediate functions)

This check should also be useless in case of generated module. It's not that expensive but it may save a little bit of time.

For some reason the `Check.modul` function returns a module but it's the same as the original one. I'm not sure why I didn't make it return a `(unit, string) Result.t` instead...

filipeom:
Done, if you want I can also change the signature of `Check.modul` and `Compile.until_check` to be of `(unit, string) Result.t` :smiley:

zapashcanon:
Thanks.

Actually I remember now why I did this. It was a reminder to add some kind of phantom type to make sure we always go through `Check.modul` (it's easy to forget...). Then we would also provide a `Check.modul_fake` or something to typecheck when we want to be unsafe but we know what we're doing.

I'll handle this later at some point, no need for you to do it. :)

PR #57, opened 2023-09-19
fix exec_extern_func

zapashcanon:
cc @chambart

### DIFF
--- a/src/interpret.ml
+++ b/src/interpret.ml
@@ -44,8 +44,6 @@ module Make (P : Interpret_functor_intf.P) :
     let consti i = const_i32 (Int32.of_int i)
   end
 
-  let ( let/ ) = Choice.bind
-
   let pop_choice stack =
     let b, stack = Stack.pop_bool stack in
     Choice.bind (Choice.select b) (fun b -> Choice.return (b, stack))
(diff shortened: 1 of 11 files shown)


src/interpret.ml
```diff
@@ -478,8 +478,11 @@ module Make (P : Interpret_functor_intf.P) :
       | I64 -> Stack.pop_i64 stack
       | F32 -> Stack.pop_f32 stack
       | F64 -> Stack.pop_f64 stack
-      | Externref _ety -> failwith "TODO"
-      (* Stack.pop_as_externref ety stack *)
+      | Externref ety -> (
+        let v, stack = Stack.pop_as_ref stack in
+        match Value.Ref.get_externref v ety with
+        | Ref_value v -> (v, stack)
+        | _ -> assert false )
```
chambart, line 485: This can be reached. This should be a proper error

chambart (changes requested):

zapashcanon:
@chambart PTAL, while doing this, I had to change the type of `pop_arg` (because it can fail now).

This led me to realize we could simplify a lot of the code in the interpreter: I removed all the `Result` wrapping in `Env.get_*` functions.

chambart:
This looks like a nice improvement.

PR #59, opened 2023-09-28
 Do not add concrete values to the path condition

filipeom:
In the `add_pc` function, avoid polluting the path condition with concrete values and promptly terminate the monad when `false` is introduced.


zapashcanon:
This seems obviously correct. :) Thanks !

Out of curiosity, is it helping the solvers that much ?

PR #67, opened 2023-10-06
Fixes assertion failure model

filipeom:
This PR addresses the issue where assertion violations were not properly reflected in the generated models due to the absence of assertion failure conditions in the path condition. In the absence of these conditions, the models only represent paths leading to assertions but not the actual violations themselves.

For instance, consider the following example program and its corresponding model. The model should capture the violation of the assertion, but it currently does not:

```wast
(module
  (import "symbolic" "i32_symbol" (func $i32_symbol (result i32)))
  (import "symbolic" "assume" (func $assume (param i32)))
  (import "symbolic" "assert" (func $assert (param i32)))
  (func $main (local i32) (local i32) (local i32)
    (call $i32_symbol)  ;; x
    (local.tee 0)
    (i32.const 0)
    (i32.eq)
    (call $assume)      ;; assume( x = 0 )
    (call $i32_symbol)
    (local.tee 1)
    (call $i32_symbol)
    (local.tee 2)
    (i32.add)           ;; y+z
    (i32.const 0)
    (i32.gt_s)          ;; (y+z)>0
    (if (result i32)
      (then
        (local.get 0)
        )
      (else
        (local.get 2)))
    (local.get 0)
    (i32.eq)
    (call $assert))
  (memory $0 1)
  (export "main" (func $main))
  (start $main))
```

In the expected scenario of an assertion violation in the path, specifically where `(x = 0 /\ not (y + z < 0))` is expected because the generated query is `(x = 0 /\ not (y + z < 0) => x = z)`, the provided model is as follows:

```
Model:
(model
  (symbol_1 i32 (i32 0))            ;; x
  (symbol_2 i32 (i32 -2147221503))  ;; y
  (symbol_3 i32 (i32 0))            ;; z
)
```

However, this model does not capture the failing condition where `x` can be different from `z` (i.e., `x != z`).

This PR rectifies this problem by introducing assertion conditions into the path condition. However, it's worth noting that this solution involves modifying every `assertion` function in each choice monad implementation, @chambart or @zapashcanon do you have any ideas that avoid so much changes?

### DIFF
--- a/src/choice_monad.ml
+++ b/src/choice_monad.ml
@@ -174,7 +174,10 @@ module List = struct
 
   let assertion c t =
     if check c t then [ ((), t) ]
-    else raise (Assertion (Encoding.Expression.to_string c, t))
+    else
+      let no = Symbolic_value.S.Bool.not c in
+      let t = { t with pc = no :: t.pc } in
+      raise (Assertion (Encoding.Expression.to_string c, t))
 
   let run (v : 'a t) (thread : thread) = List.to_seq (v thread)
 end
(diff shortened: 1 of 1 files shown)


zapashcanon:
LGTM, thanks !

> However, it's worth noting that this solution involves modifying every assertion function in each choice monad implementation, @chambart or @zapashcanon do you have any ideas that avoid so much changes?

I'll try to refactor the various modules a little bit.

Issue #12, opened 2023-01-16
add a --with-no-exhaustion flag


zapashcanon:
it would make sense only when `--script` is used, and would ignore all exhaustion tests from Wasm scripts, it's useful when debugging

PR #69, opened 2023-10-14
use subcommands instead of various flags/binaries

zapashcanon:
Hi,

Now we only have one entry point (`bin/owi.ml`) which is the `owi` command.

For now, three subcommands are available:
- `owi run` (concrete interpreter)
- `owi sym` (symbolic interpreter)
- `owi script` (reference test script interpreter)


zapashcanon:
Fix #12.

PR #71, opened 2023-10-16
One more concat simplification

filipeom:
Adds one more concatenation simplification to the symbolic memory. To avoid polluting the symbolic memory with these simplifications, once https://github.com/wasp-platform/encoding/issues/32 is completed, I'll move this code to `Expression.simplify`.

### DIFF
--- a/src/symbolic_memory.ml
+++ b/src/symbolic_memory.ml
@@ -71,6 +71,14 @@ module M = struct
 
   let concat ~msb ~lsb offset =
     assert (offset > 0 && offset <= 8);
+    let merge_extracts (e1, h, m1) (e2, m2, l) =
+      if not (m1 = m2 && Expr.equal e1 e2) then
+        Expr.(Extract (e1, h, m1) ++ Extract (e2, m2, l))
+      else
+        match Expr.type_of e1 with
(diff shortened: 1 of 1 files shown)


src/symbolic_memory.ml
```diff
@@ -71,6 +71,14 @@ module M = struct
...
   let concat ~msb ~lsb offset =
     assert (offset > 0 && offset <= 8);
+    let merge_extracts (e1, h, m1) (e2, m2, l) =
+      if not (m1 = m2 && Expr.equal e1 e2) then
+        Expr.(Extract (e1, h, m1) ++ Extract (e2, m2, l))
+      else
+        match Expr.type_of e1 with
+        | Some t -> if h - l = Types.size t then e1 else Extract (e1, h, l)
+        | None -> Extract (e1, h, l)
+    in
```
zapashcanon, line 81: This can be written a little bit shorter:

```ocaml
        match Expr.type_of e1 with
        | Some t when h - l = Types.size t -> e1
        | None | Some _ -> Extract (e1, h, l)
```

zapashcanon:
Thanks!

PR #73, opened 2023-10-31
Test-comp 24

filipeom:
Required test-comp changes. Depends on https://github.com/formalsec/encoding/commit/ce2225b42ee212def2a1ec07aafc0d66ad8581cf

### DIFF
--- a/src/cmd_sym.ml
+++ b/src/cmd_sym.ml
@@ -58,24 +58,9 @@ let symbolic_extern_module : Symbolic.P.extern_func Link.extern_module =
     let r = Expr.mk_symbol_s `I32Type (sprintf "%s_%i" name id) in
     Choice.return r
   in
-  let symbol_i32 () : Value.int32 Choice.t =
+  let symbol ty () : Value.int32 Choice.t =
     let id = Atomic.fetch_and_add sym_cnt 1 in
-    let r = Expr.mk_symbol_s `I32Type (sprintf "symbol_%i" id) in
-    Choice.return r
-  in
-  let symbol_i64 () : Value.int64 Choice.t =
-    let id = Atomic.fetch_and_add sym_cnt 1 in
-    let r = Expr.mk_symbol_s `I64Type (sprintf "symbol_%i" id) in
-    Choice.return r
-  in
-  let symbol_f32 () : Value.float32 Choice.t =
-    let id = Atomic.fetch_and_add sym_cnt 1 in
-    let r = Expr.mk_symbol_s `F32Type (sprintf "symbol_%i" id) in
-    Choice.return r
-  in
-  let symbol_f64 () : Value.float64 Choice.t =
-    let id = Atomic.fetch_and_add sym_cnt 1 in
-    let r = Expr.mk_symbol_s `F64Type (sprintf "symbol_%i" id) in
+    let r = Expr.mk_symbol_s ty (sprintf "symbol_%i" id) in
     Choice.return r
   in
   let assume_i32 (i : Value.int32) : unit Choice.t =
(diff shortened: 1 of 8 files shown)


src/cmd_sym.ml
```diff
@@ -219,55 +213,96 @@ let run_file ~unsafe ~optimize (pc : unit Result.t Choice.t) filename =
...
+let out_testcase ~dst ~err testcase =
+  let o = Xmlm.make_output ~nl:true ~indent:(Some 2) dst in
+  let tag ?(atts = []) name = (("", name), atts) in
+  let atts = if err then Some [ (("", "coversError"), "true") ] else None in
+  let to_string v = Format.asprintf "%a" Encoding.Value.pp_num v in
+  let input v = `El (tag "input", [ `Data (to_string v) ]) in
+  let testcase = `El (tag ?atts "testcase", testcase >>| input) in
+  let dtd =
+    "<!DOCTYPE testcase PUBLIC \"+//IDN sosy-lab.org//DTD test-format testcase \
+     1.1//EN\" \"https://sosy-lab.org/test-format/testcase-1.1.dtd\">"
```
zapashcanon, line 239: You can use the `{|` and `|}` delimiters to avoid the escaping:

```ocaml
let dtd =
  {|<!DOCTYPE testcase PUBLIC "+//IDN sosy-lab.org//DTD test-format testcase 1.1//EN" "https://sosy-lab.org/test-format/testcase-1.1.dtd">|}
```

src/cmd_sym.ml
```diff
@@ -219,55 +213,96 @@ let run_file ~unsafe ~optimize (pc : unit Result.t Choice.t) filename =
...
+  assert (Solver.check solver pc);
+  match Solver.model solver with None -> assert false | Some model -> model
 
-let stop_at_first_failure = true
+let mkdir_p_exn dir =
+  let rec get_intermediate_dirs d acc =
+    if Sys.file_exists d then acc
+    else get_intermediate_dirs (Filename.dirname d) (d :: acc)
+  in
+  if Sys.file_exists dir then ()
```
zapashcanon, line 225: I believe we can simply remove this check ?

src/cmd_sym.ml
```diff
@@ -41,54 +46,43 @@ let assert_extern_module : Symbolic.P.extern_func Link.extern_module =
...
       end
-      | _ ->
-        failwith
-          (Printf.sprintf "Text name %s" (Encoding.Expression.to_string i))
-    in
-    incr counter;
-    let r =
-      Encoding.Expression.mk_symbol_s `I32Type
-        (Printf.sprintf "%s_%i" name !counter)
+      | _ -> failwith (sprintf "Text name %s" (Expr.to_string i))
```
zapashcanon, line 62: Maybe: `Format.kasprintf failwith "Text name %a" Expr.pp i` ?

src/cmd_sym.ml
filipeom: Good idea! But without escaping we're going to have a gigantic line, are you okay with that?

zapashcanon: I don't mind. :)

src/cmd_sym.ml
```diff
@@ -41,54 +46,43 @@ let assert_extern_module : Symbolic.P.extern_func Link.extern_module =
 let names = [| "plop"; "foo"; "bar" |]
 
 let symbolic_extern_module : Symbolic.P.extern_func Link.extern_module =
-  let counter = ref 0 in
+  let sprintf = Printf.sprintf in
+  let next_sym_id () : int Choice.t =
+    let/ cnt = Choice.with_thread Thread.sym_cnt in
```
zapashcanon, line 51: Actually, I believe there's no need to put `sym_cnt` in `Thread`. Using a global `int Atomic.t` should be enough, right ?

src/cmd_sym.ml
```diff
@@ -219,55 +206,93 @@ let run_file ~unsafe ~optimize (pc : unit Result.t Choice.t) filename =
...
-          Some thread
-        | ETrap tr ->
-          Format.printf "TRAP: %s@." (Trap.to_string tr);
-          let model = get_model solver thread in
-          Format.printf "Model:@.%s@." model;
-          Some thread
-        | EVal (Error e) ->
-          Format.eprintf "%s@." e;
-          exit 1 )
+        Format.printf "%a@." Expr.pp_list pc;
```
zapashcanon, line 260: We could merge these two calls to `Format.printf` ?

src/cmd_sym.ml
```diff
@@ -219,55 +206,93 @@ let run_file ~unsafe ~optimize (pc : unit Result.t Choice.t) filename =
...
+            Some pc
+          | EVal (Error e) ->
+            Format.eprintf "%s@." e;
+            exit 1
+        in
+        let testcase =
+          List.sort
+            (fun (x1, _) (x2, _) -> compare x1 x2)
+            (Encoding.Model.get_bindings model)
+          >>| snd
```
zapashcanon, line 281: Could we replace this by `|> List.map snd` to avoid defining a custom operator that is used only once ?

zapashcanon:
> I'd just probably just like to complete the description

I don't know what you're talking about :sweat_smile:

filipeom:
> I don't know what you're talking about 😅 but fine, let me know when you believe it's ready

I meant the PR description, but it's pointless, feel free to merge :smiley:

PR #76, opened 2023-11-21
Updates encoding pin

filipeom:
Sets the encoding pin to v0.0.1, guaranteeing smooth future installations and providing a safeguard against conflicts with additions to the encoding's main branch. Should the pull request at https://github.com/ocaml/opam-repository/pull/24809 be accepted in the future, the pin can be removed.

### DIFF
--- a/src/cmd_sym.ml
+++ b/src/cmd_sym.ml
@@ -1,6 +1,5 @@
 open Syntax
-module E = Encoding
-module Expr = E.Expr
+module Expr = Encoding.Expr
 module Value = Symbolic_value.S
 module Choice = Symbolic.P.Choice
 module Solver = Thread.Solver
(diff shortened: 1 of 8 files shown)


src/cmd_sym.ml
```diff
@@ -1,5 +1,6 @@
 open Syntax
-module Expr = Encoding.Expression
+module E = Encoding
```
zapashcanon, line 2: Looks like this is used only once, we can remove it

zapashcanon:
> Should the pull request at https://github.com/ocaml/opam-repository/pull/24809 be accepted in the future, the pin can be removed.

It should be accepted quite quickly, you can already remove the pin. I'll merge once it's available. :)

zapashcanon:
Thanks!

PR #83, opened 2023-11-26
add fmt command to format .wast file

swrup:


src/cmd_fmt.ml
```diff
@@ -0,0 +1,26 @@
+(* SPDX-License-Identifier: AGPL-3.0-or-later *)
+(* Copyright © 2021 Léo Andrès *)
+(* Copyright © 2021 Pierre Chambart *)
```
zapashcanon, line 3: It should rather be `2023 Olivier Pierre`.

src/bin/owi.ml
```diff
@@ -52,6 +52,19 @@ let shared_man =
...
+  let info =
+    let doc = "Format a module" in
+    let man = [] @ shared_man in
+    Cmd.info "fmt" ~version ~doc ~sdocs ~man
+  in
+  let inplace =
+    let doc = "Format in-place, overwriting input file" in
+    Cmdliner.Arg.(value & flag & info [ "inplace"; "i" ] ~doc)
+  in
+  Cmd.v info Term.(const Cmd_fmt.cmd $ debug $ unsafe $ inplace $ file)
```
zapashcanon, line 66: I believe `debug` and `unsafe` should be removed here.

src/text.ml
```diff
@@ -108,7 +108,7 @@ type action =
 
 let pp_action fmt = function
   | Invoke (mod_name, name, c) ->
-    pp fmt "(invoke %a %s %a)" pp_id_opt mod_name name pp_consts c
+    pp fmt {|(invoke%a "%s" %a)|} pp_id_opt mod_name name pp_consts c
```
zapashcanon, line 111: You added some `"` so the code was broken before, right ?

src/types.ml
```diff
@@ -34,7 +34,11 @@ type _ indice =
 
 let pp_id fmt id = pp fmt "$%s" id
 
-let pp_id_opt fmt = function None -> () | Some i -> pp_id fmt i
+let pp_id_opt fmt = function
+  | None -> ()
+  | Some i ->
+    pp fmt " ";
+    pp_id fmt i
```
zapashcanon, line 41: rather `pp fmt " %a" pp_id i`

src/types.ml
```diff
@@ -306,6 +310,12 @@ let pp_result_ fmt vt = pp fmt "(result %a)" pp_val_type vt
 
 let pp_result_type fmt results = pp_list ~pp_sep:pp_space pp_result_ fmt results
 
+(* wrap printer to print a space before a non empty list *)
+(* TODO or make it an optional arg of pp_list? *)
+let with_space_list printer fmt l =
+  (match l with [] -> () | _l -> pp fmt " ");
+  printer fmt l
```
zapashcanon, line 317:
```ocaml
let with_space_list printer fmt = function
  | [] -> ()
   | l -> pp fmt " %a" printer l
```

src/types.ml
```diff
@@ -675,7 +692,7 @@ type 'a export =
   }
 
 let pp_export fmt (e : text export) =
-  pp fmt "(export %a %a)" pp_string e.name pp_export_desc e.desc
+  pp fmt {|(export "%s" %a)|} e.name pp_export_desc e.desc
```
zapashcanon, line 695: this was broken too ? :sweat_smile:

src/types.ml
```diff
@@ -750,17 +767,17 @@ type 'a const =
...
+  | Const_I64 i -> pp fmt "(i64.const %Ld)" i
+  | Const_F32 f -> pp fmt "(f32.const %a)" Float32.pp f
+  | Const_F64 f -> pp fmt "(f64.const %a)" Float64.pp f
+  | Const_null rt -> pp fmt "(ref.null %a)" pp_heap_type rt
+  | Const_host i -> pp fmt "(ref.host %d)" i
+  | Const_extern i -> pp fmt "(ref.extern %d)" i
   | Const_array -> pp fmt "ref.array"
   | Const_eq -> pp fmt "ref.eq"
   | Const_i31 -> pp fmt "ref.i31"
   | Const_struct -> pp fmt "ref.struct"
```
zapashcanon, line 780: The parens are not needed for `ref.{array,eq,i31,struct}` ?

src/types.ml
```diff
@@ -750,17 +767,17 @@ type 'a const =
...
 
 let pp_consts fmt c =
-  pp_list ~pp_sep:pp_space (fun fmt c -> pp fmt "(%a)" pp_const c) fmt c
+  pp_list ~pp_sep:pp_space (fun fmt c -> pp fmt "%a" pp_const c) fmt c
```
zapashcanon, line 783: `pp_list ~pp_sep:pp_space pp_const fmt c`

zapashcanon:
Thanks !

Could you remove `test/print/print_text.ml` and update `test/print/print.t` to replace the line:

`dune exec -- ./print_text.exe m.wast`

by

`dune exec owi -- fmt m.wast`

?

zapashcanon:
Also fix #83.

zapashcanon:
One more little thing: could you try to check the file extension and if it's `.wast` then do what you already did (i.e. `Parse.Script.from_file` and `Text.pp_script`.

But if it's `.wat` then, call `Parse.Module.from_file` and `Text.pp_modul` instead ?

src/text.ml
Swrup: yes

src/types.ml
Swrup: yes same thing as invoke

src/types.ml
Swrup: I don't know,
since `ref.{array,eq,i31,struct}` are just one "word" it is probably not needed, I dont' know how to check this

zapashcanon: These are used to test the return value of an invoke through `assert_return` but they're from the GC proposal so you'd need to clone [WebAssembly/gc](https://github.com/WebAssembly/gc) to try them.

You can find some files in `test/gc` (they'll be moved to `test/reference/gc` once we support the GC proposal completely).

If you look at the `i31.wast` file, you'll see one instance of this (line 28).

The reason they have a single work is because we can not know what will be the real value hold by the reference in case of struct/eq/array etc. (we could for i31 but that's another story).

Anyway, I tried removing the parenthesis and got an error from the reference interpreter so I guess they should stay here.

Swrup:
> Could you add a third mode where we parse, then we format and check that everything goes well ? It shouldn't be too hard by simply looking at the `test_directory` and `test_file` functions.

It's not going well :weary:

src/cmd_fmt.ml
```diff
@@ -2,22 +2,31 @@
...
-let read_script filename =
+let get_printer filename =
   if not @@ Sys.file_exists filename then
     error_s "file `%s` doesn't exist" filename
-  else Parse.Script.from_file ~filename
+  else
+    let ext = Filename.extension filename in
+    match ext with
+    | ".wat" ->
+      Result.bind (Parse.Module.from_file ~filename) (fun v ->
```
zapashcanon, line 12:
```ocaml
let+ v = Parse.Module.from_file ~filename in
fun fmt () -> Text.pp_modul fmt v
```

src/cmd_fmt.ml
```diff
@@ -2,22 +2,31 @@
...
+          Ok (fun fmt () -> Text.pp_modul fmt v) )
+    | ".wast" ->
+      Result.bind (Parse.Script.from_file ~filename) (fun v ->
+          Ok (fun fmt () -> Text.pp_script fmt v) )
```
zapashcanon, line 16: same here

test/main.ml
```diff
@@ -12,14 +12,30 @@ let pp_ok () = Format.fprintf fmt "%a !@." pp_green "OK"
...
+    | `Normal -> "file          " )
+    Fpath.pp filename;
   try
-    match Owi.Parse.Script.from_file ~filename:(Fpath.to_string f) with
+    let res =
+      match mode with
+      | `Optimize | `Normal ->
+        Owi.Parse.Script.from_file ~filename:(Fpath.to_string filename)
+      | `Formatted ->
+        Result.bind
```
zapashcanon, line 28:
```ocaml
let+ s = ... in
Parse.Script.from_string s
```

test/main.ml
```diff
@@ -42,21 +58,17 @@ let test_directory d =
     List.iter
       (fun file ->
         incr count_total;
-        begin
-          match test_file ~optimize:false file with
+        let run mode =
```
zapashcanon, line 61: can you lift this out of the `List.iter` ?

test/main.ml
```diff
@@ -12,14 +12,30 @@ let pp_ok () = Format.fprintf fmt "%a !@." pp_green "OK"
...
 
-let test_file ~optimize f =
+let test_file mode filename =
   Format.fprintf fmt "testing %s: `%a`... "
-    (if optimize then "optimized file" else "file          ")
-    Fpath.pp f;
+    ( match mode with
+    | `Optimize -> "optimized file"
+    | `Formatted -> "formatted file"
+    | `Normal -> "file          " )
```
zapashcanon, line 20: Could you define the following instead:

```ocaml
type test_mode =
  | Normal
  | Fmt
  | Opt
```

and avoid the polymorphic variant ?

PR #95, opened 2023-12-10
Heap tracking

filipeom:
Hi @zapashcanon!

Here's a quick summary of the introduced changes:

1. Adds heap pointer tracking to the symbolic memory, enabling the detection of:
    - Heap buffer overflows
    - Use-after-free errors
    - Double frees
    - Memory leaks (Currently not reporting, but we only need to check if there are still bindings in the `chunks` table at the end of the program)

2. Enables the creation of `I8` symbols with the external function `i8_symbol`.

3. Stores concrete bytes in the symbolic memory using `I8` instead of `I32`.

Ideally, in the external functions, I would prefer a more direct way of interacting with the memory. This would simplify the code for `alloc` and `free`, and it would also allow for a more accurate characterization of memory leaks.

Also, I added a couple C examples [here](https://github.com/formalsec/waspc/tree/main/examples/memory-safety), but the compiled version is also there if you want to test it directly :smiley:

### DIFF
--- a/src/cmd_sym.ml
+++ b/src/cmd_sym.ml
@@ -104,26 +104,45 @@ let symbolic_extern_module : Symbolic.P.extern_func Link.extern_module =
   { functions }
 
 let summaries_extern_module : Symbolic.P.extern_func Link.extern_module =
+  let open Expr in
+  let i32 v = match v.e with Val (Num (I32 v)) -> v | _ -> assert false in
+  let ptr v = match v.e with Ptr (b, _) -> b | _ -> assert false in
   let abort () : unit Choice.t = Choice.add_pc @@ Value.Bool.const false in
-  let alloc (base : Value.int32) (_size : Value.int32) : Value.int32 Choice.t =
-    Choice.return base
+  let alloc (base : Value.int32) (size : Value.int32) : Value.int32 Choice.t =
+    let base : int32 = i32 base in
+    Choice.with_thread (fun t ->
+        let memories = Thread.memories t in
+        Env_id.Tbl.iter
+          (fun _ tbl ->
+            Symbolic_memory.ITbl.iter
+              (fun _ (m : Symbolic_memory.M.t) ->
+                Hashtbl.replace m.chunks base size )
+              tbl )
+          memories;
+        Ptr (base, Value.const_i32 0l) @: Ty_bitv S32 )
   in
-  let dealloc (_base : Value.int32) : unit Choice.t = Choice.return () in
-  let is_symbolic (_a : Value.int32) (_n : Value.int32) : Value.int32 Choice.t =
-    (* TODO: load n bytes from address a and check if it's a Val *)
-    Choice.return @@ Value.const_i32 0l
+  let free (p : Value.int32) : unit Choice.t =
+    let base = ptr p in
+    Choice.with_thread (fun t ->
+        let memories = Thread.memories t in
+        Env_id.Tbl.iter
+          (fun _ tbl ->
+            Symbolic_memory.ITbl.iter
+              (fun _ (m : Symbolic_memory.M.t) ->
+                if not (Hashtbl.mem m.chunks base) then
+                  (* Raise trap here instead *)
+                  failwith "Memory leak use after free";
+                Hashtbl.remove m.chunks base )
+              tbl )
+          memories )
   in
   let functions =
     [ ( "alloc"
       , Symbolic.P.Extern_func.Extern_func
           (Func (Arg (I32, Arg (I32, Res)), R1 I32), alloc) )
     ; ( "dealloc"
-      , Symbolic.P.Extern_func.Extern_func (Func (Arg (I32, Res), R0), dealloc)
-      )
+      , Symbolic.P.Extern_func.Extern_func (Func (Arg (I32, Res), R0), free) )
     ; ("abort", Symbolic.P.Extern_func.Extern_func (Func (UArg Res, R0), abort))
-    ; ( "is_symbolic"
-      , Symbolic.P.Extern_func.Extern_func
-          (Func (Arg (I32, Arg (I32, Res)), R1 I32), is_symbolic) )
     ]
   in
   { functions }
--- a/src/cmd_sym.ml
+++ b/src/cmd_sym.ml
@@ -66,7 +66,9 @@ let symbolic_extern_module : Symbolic.P.extern_func Link.extern_module =
   let symbol ty () : Value.int32 Choice.t =
     let id = Atomic.fetch_and_add sym_cnt 1 in
     let r = Expr.mk_symbol @@ mk_symbol ty (Format.sprintf "symbol_%i" id) in
-    Choice.return r
+    match ty with
+    | Ty_bitv S8 -> Choice.return @@ Expr.(Cvtop (ExtU 24, r) @: Ty_bitv S32)
+    | _ -> Choice.return r
   in
   let assume_i32 (i : Value.int32) : unit Choice.t =
     let c = Value.I32.to_bool i in
@@ -81,6 +83,9 @@ let symbolic_extern_module : Symbolic.P.extern_func Link.extern_module =
     [ ( "i32"
       , Symbolic.P.Extern_func.Extern_func
           (Func (Arg (I32, Res), R1 I32), symbolic_i32) )
+    ; ( "i8_symbol"
+      , Symbolic.P.Extern_func.Extern_func
+          (Func (UArg Res, R1 I32), symbol (Ty_bitv S8)) )
     ; ( "i32_symbol"
       , Symbolic.P.Extern_func.Extern_func
           (Func (UArg Res, R1 I32), symbol (Ty_bitv S32)) )
--- a/src/cmd_sym.ml
+++ b/src/cmd_sym.ml
@@ -136,7 +136,7 @@ let summaries_extern_module : Symbolic.P.extern_func Link.extern_module =
               (fun _ (m : Symbolic_memory.M.t) ->
                 if not (Hashtbl.mem m.chunks base) then
                   (* TODO: trap instead? *)
-                  failwith "Memory leak use after free";
+                  failwith "Memory leak double free";
                 Hashtbl.remove m.chunks base )
               tbl )
           memories )
--- a/src/cmd_sym.ml
+++ b/src/cmd_sym.ml
@@ -135,7 +135,7 @@ let summaries_extern_module : Symbolic.P.extern_func Link.extern_module =
             Symbolic_memory.ITbl.iter
               (fun _ (m : Symbolic_memory.M.t) ->
                 if not (Hashtbl.mem m.chunks base) then
-                  (* Raise trap here instead *)
+                  (* TODO: trap instead? *)
                   failwith "Memory leak use after free";
                 Hashtbl.remove m.chunks base )
               tbl )
--- a/src/cmd_sym.ml
+++ b/src/cmd_sym.ml
@@ -110,8 +110,16 @@ let symbolic_extern_module : Symbolic.P.extern_func Link.extern_module =
 
 let summaries_extern_module : Symbolic.P.extern_func Link.extern_module =
   let open Expr in
-  let i32 v = match v.e with Val (Num (I32 v)) -> v | _ -> assert false in
-  let ptr v = match v.e with Ptr (b, _) -> b | _ -> assert false in
+  let i32 v =
+    match v.e with
+    | Val (Num (I32 v)) -> v
+    | _ -> Log.err {|Unable to cast int32 of "%a"|} Expr.pp v
+  in
+  let ptr v =
+    match v.e with
+    | Ptr (b, _) -> b
+    | _ -> Log.err {|Unable to fetch pointer base of "%a"|} Expr.pp v
+  in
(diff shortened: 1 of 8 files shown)


src/symbolic.ml
```diff
@@ -106,7 +106,47 @@ module P = struct
...
+
+    let blit = M.blit
+
+    let blit_string = M.blit_string
+
+    let size = M.size
+
+    let size_in_pages = M.size_in_pages
+
+    let get_limit_max = M.get_limit_max
```
zapashcanon, line 149: You can `include Symbolic_memory.M` in the beginning to avoid having to re-export all these functions

src/cmd_sym.ml
```diff
@@ -104,26 +109,45 @@ let symbolic_extern_module : Symbolic.P.extern_func Link.extern_module =
   { functions }
 
 let summaries_extern_module : Symbolic.P.extern_func Link.extern_module =
+  let open Expr in
+  let i32 v = match v.e with Val (Num (I32 v)) -> v | _ -> assert false in
+  let ptr v = match v.e with Ptr (b, _) -> b | _ -> assert false in
```
zapashcanon, line 114: In both cases, could you use `Log.err` and add a proper error message instead of the `assert false` ?

src/symbolic_memory.ml
```diff
@@ -5,6 +5,8 @@
 module Intf = Interpret_functor_intf
 module Value = Symbolic_value.S
 
+let ( let* ) = Stdlib.Result.bind
```
zapashcanon, line 8: You can use `open Syntax` instead

src/symbolic_memory.ml
```diff
@@ -20,16 +22,28 @@ module M = struct
...
     ; mutable size : int
+    ; chunks : (Int32.t, Expr.t) Hashtbl.t
     }
 
   let create size =
-    { data = Hashtbl.create 128; parent = None; size = Int32.to_int size }
+    { data = Hashtbl.create 128
+    ; parent = None
+    ; size = Int32.to_int size
+    ; chunks = Hashtbl.create 0
```
zapashcanon, line 32: You can put `Hashtbl.create 16` here, it's what the stdlib is going to do anyway.

src/symbolic_memory.ml
```diff
@@ -20,16 +22,28 @@ module M = struct
...
+    }
 
-  let concretize_i32 a =
-    match a with Val (Num (I32 i)) -> i | _ -> assert false
+  let i32 v =
+    match v.e with
+    | Val (Num (I32 i)) -> i
+    | _ ->
+      Format.kasprintf failwith "Unsupported symbolic value reasoning over '%a'"
```
zapashcanon, line 39: We have `Log.err` which is doing exactly this. :)

src/symbolic_memory.ml
```diff
@@ -20,16 +22,28 @@ module M = struct
...
+        Expr.pp v
 
   let grow m delta =
-    let delta = Int32.to_int @@ concretize_i32 delta.e in
+    let delta =
+      match delta.e with Val (Num (I32 i)) -> i | _ -> assert false
```
zapashcanon, line 44: Isn't this the `concretize_i32` (now `i32`) function defined just above ?

src/symbolic_memory.ml
```diff
@@ -43,54 +57,62 @@ module M = struct
...
         let dst = Int32.of_int (dst + i) in
-        Hashtbl.replace m.data dst (Value.const_i32 byte)
+        Hashtbl.replace m.data dst (Val (Num (I8 byte)) @: Ty_bitv S8)
       done;
       Value.Bool.const false
     end
 
-  let clone m = { data = Hashtbl.create 0; parent = Some m; size = m.size }
+  let clone m =
+    { data = Hashtbl.create 0
```
zapashcanon, line 78: `16` here too

src/symbolic_memory.ml
```diff
@@ -111,51 +133,81 @@ module M = struct
...
+    | Val (Num (I32 i)) -> Ok i
+    | Ptr (base, offset) -> (
+      match Hashtbl.find_opt m.chunks base with
+      | None -> Error Trap.Memory_leak_use_after_free
+      | Some size ->
+        let ptr = Int32.add base (i32 offset) in
+        if ptr < base || ptr > Int32.add base (i32 size) then
+          Error Trap.Memory_heap_buffer_overflow
+        else Ok ptr )
+    | _ -> Format.kasprintf failwith "Cannot calculate address of: %a" Expr.pp a
```
zapashcanon, line 150: `Log.err`

src/symbolic_memory.ml
```diff
@@ -111,51 +133,81 @@ module M = struct
...
     loop a n 1 v0
 
+  (* TODO: *)
+  (* 1. Let pointers have symbolic offsets *)
+  (* 2. Let addresses have symbolic values *)
+  let calculate_address m (a : int32) : (Int32.t, Trap.t) Stdlib.Result.t =
+    match a.e with
+    | Val (Num (I32 i)) -> Ok i
+    | Ptr (base, offset) -> (
+      match Hashtbl.find_opt m.chunks base with
```
zapashcanon, line 143: You can use `Hashtbl.find` and match on `exception Not_found -> ... | size -> ...` to avoid an allocation in the most frequent case.

src/symbolic_memory.ml
```diff
@@ -111,51 +133,81 @@ module M = struct
...
   let load_8_s m a =
-    let v = loadn m (concretize_i32 a.e) 1 in
+    let* a = calculate_address m a in
+    let v = loadn m a 1 in
     match v.e with
-    | Val (Num (I32 i8)) -> Value.const_i32 (Int32.extend_s 8 i8)
-    | _ -> Cvtop (ExtS 24, v) @: Ty_bitv S32
+    | Val (Num (I8 i8)) ->
+      Ok (Value.const_i32 (Int32.extend_s 8 (Int32.of_int i8)))
+    | _ -> Ok (Cvtop (ExtS 24, v) @: Ty_bitv S32)
```
zapashcanon, line 158: If you replace the `let*` by `let+` in the beginning, then you can remove the `Ok` in each case of the `match`.

src/symbolic_memory.ml
```diff
@@ -111,51 +133,81 @@ module M = struct
...
 
   let load_8_u m a =
-    let v = loadn m (concretize_i32 a.e) 1 in
+    let* a = calculate_address m a in
```
zapashcanon, line 161: `let+`

src/symbolic_memory.ml
```diff
@@ -111,51 +133,81 @@ module M = struct
...
+    let v = loadn m a 1 in
     match v.e with
-    | Val (Num (I32 _)) -> v
-    | _ -> Cvtop (ExtU 24, v) @: Ty_bitv S32
+    | Val (Num (I8 i)) -> Ok (Value.const_i32 (Int32.of_int i))
+    | _ -> Ok (Cvtop (ExtU 24, v) @: Ty_bitv S32)
 
   let load_16_s m a =
-    let v = loadn m (concretize_i32 a.e) 2 in
+    let* a = calculate_address m a in
```
zapashcanon, line 168: `let+`

src/symbolic_memory.ml
```diff
@@ -111,51 +133,81 @@ module M = struct
...
+    let v = loadn m a 2 in
     match v.e with
-    | Val (Num (I32 i16)) -> Value.const_i32 (Int32.extend_s 16 i16)
-    | _ -> Cvtop (ExtS 16, v) @: Ty_bitv S32
+    | Val (Num (I32 i16)) -> Ok (Value.const_i32 (Int32.extend_s 16 i16))
+    | _ -> Ok (Cvtop (ExtS 16, v) @: Ty_bitv S32)
 
   let load_16_u m a =
-    let v = loadn m (concretize_i32 a.e) 2 in
+    let* a = calculate_address m a in
```
zapashcanon, line 175: `let+`

src/symbolic_memory.ml
```diff
@@ -111,51 +133,81 @@ module M = struct
...
+    let v = loadn m a 2 in
     match v.e with
-    | Val (Num (I32 _)) -> v
-    | _ -> Cvtop (ExtU 16, v) @: Ty_bitv S32
+    | Val (Num (I32 _)) -> Ok v
+    | _ -> Ok (Cvtop (ExtU 16, v) @: Ty_bitv S32)
 
-  let load_32 m a = loadn m (concretize_i32 a.e) 4
+  let load_32 m a =
+    let* a = calculate_address m a in
```
zapashcanon, line 182: `let+`

src/symbolic_memory.ml
```diff
@@ -111,51 +133,81 @@ module M = struct
...
+    Ok (loadn m a 4)
 
-  let load_64 m a = loadn m (concretize_i32 a.e) 8
+  let load_64 m a =
+    let* a = calculate_address m a in
```
zapashcanon, line 186: `let+`

src/symbolic_memory.ml
```diff
@@ -111,51 +133,81 @@ module M = struct
...
+      let i' = Int32.(to_int @@ logand 0xffl @@ shr_s i @@ of_int (pos * 8)) in
+      Val (Num (I8 i')) @: Ty_bitv S8
     | Val (Num (I64 i)) ->
-      let i' = Int64.(logand 0xffL @@ shr_s i @@ of_int (pos * 8)) in
-      Value.const_i32 (Int32.of_int64 i')
+      let i' = Int64.(to_int @@ logand 0xffL @@ shr_s i @@ of_int (pos * 8)) in
+      Val (Num (I8 i')) @: Ty_bitv S8
+    | Cvtop (ExtU 24, { e = Symbol s; ty })
+    | Cvtop (ExtS 24, { e = Symbol s; ty })
+      when ty = Ty_bitv S8 ->
```
zapashcanon, line 199: I believe a `when` is not needed here and you can directly put it in the pattern.

src/symbolic_memory.ml
```diff
@@ -111,51 +133,81 @@ module M = struct
...
+      Symbol s @: ty
     | _ -> Extract (v, pos + 1, pos) @: Ty_bitv S8
 
   let storen m ~addr v n =
-    let a0 = concretize_i32 addr.e in
+    let* a0 = calculate_address m addr in
```
zapashcanon, line 204: `let+`

src/symbolic_memory.ml
```diff
@@ -111,51 +133,81 @@ module M = struct
...
     for i = 0 to n - 1 do
       let addr' = Int32.add a0 (Int32.of_int i) in
       let v' = extract v i in
       Hashtbl.replace m.data addr' v'
-    done
+    done;
+    Ok ()
```
zapashcanon, line 210: with the `let+` you can leave this as before

src/symbolic_memory.ml
```diff
@@ -43,54 +57,62 @@ module M = struct
...
+    { data = Hashtbl.create 0
+    ; parent = Some m
+    ; size = m.size
+    ; chunks = Hashtbl.copy m.chunks (* TODO: we can make this lazy as well *)
+    }
 
   let rec load_byte_opt a m =
     match Hashtbl.find_opt m.data a with
     | Some b -> Some b
     | None -> Option.bind m.parent (load_byte_opt a)
```
zapashcanon, line 87: I know this is not code from the current PR, but could you also change it to use `find` and match on `exception Not_found` and `b` directly ?

src/symbolic_memory.ml
```diff
@@ -43,54 +57,62 @@ module M = struct
...
 
-  let load_byte m a = Option.value (load_byte_opt a m) ~default:Value.I32.zero
+  let load_byte m a =
+    Option.value (load_byte_opt a m) ~default:(Val (Num (I8 0)) @: Ty_bitv S8)
```
zapashcanon, line 90: Could you rewrite these two functions into a single one:

```ocaml
let rec load_byte { data; parent; _ } a =
  try Hashtbl.find data a
  with Not_found ->
     begin match parent with
     | None -> Val (Num (I8 0)) @: Ty_bitv S8)
     | Some parent -> load_byte parent a
     end
```

src/symbolic_memory.ml
```diff
@@ -43,54 +57,62 @@ module M = struct
...
+
+  let merge_extracts (e1, h, m1) (e2, m2, l) =
+    if not (m1 = m2 && Expr.equal e1 e2) then
```
zapashcanon, line 93: `if m1 <> m2 || not (Expr.equal e1 e2)`

src/symbolic_memory.ml
filipeom: Atm, I cannot use the `let*` defined in syntax as it uses `Owi.Result.t` which is `('a, string) Stdlib.Result.t`. I'm using `('a, Trap.t) Stdlib.Result.t` so that I'm able to raise `Error trap` in the symbolic memory and then `Choice.trap` in `symbolic.ml`

zapashcanon: Oh OK. I need to change our error type at some point. Could you add a `TODO: use Syntax module` comment so that I think to update it when I change the types ?

filipeom: Yes, can do :smiley:

zapashcanon:
Thanks!

PR #96, opened 2024-01-05
Test coverage improvement

epatrizio:


test/optimize/not_exists.wast
```diff
@@ -0,0 +1,4 @@
+(module
+  (func $start)
+  (start $start)
+)
```
zapashcanon, line 4: You should remove this file.

test/sym/not_exists.wast
```diff
@@ -0,0 +1,4 @@
+(module
+  (func $start)
+  (start $start)
+)
```
zapashcanon, line 4: You should remove this file.

src/cmd_opt.ml
```diff
@@ -6,7 +6,7 @@ open Syntax
 
 let optimize_file ~unsafe filename =
   if not @@ Sys.file_exists filename then
-    error_s "file `%s` doesn't exist" filename
+    error_s "Error: file `%s` doesn't exist" filename
```
zapashcanon, line 9: Could you remove the `Error:` ? I used to do this but then ended up in many nested `Error: Error:` when wrapping errors and it quickly becomes a mess. :sweat_smile:

test/optimize/dune
```diff
@@ -12,6 +12,7 @@
   ibinop.wast
   if.wast
   local.wast
+  not_exists.wast
```
zapashcanon, line 15: You should remove this.

test/sym/dune
```diff
@@ -2,4 +2,5 @@
  (deps
   %{bin:owi}
   mini_test.wast
+  not_exists.wast
```
zapashcanon, line 5: You should remove this.

src/cmd_sym.ml
```diff
@@ -9,37 +9,16 @@ let print_solver_time = false
...
-      , Symbolic.P.Extern_func.Extern_func (Func (Arg (I32, Res), R0), print_i32)
-      )
-    ]
-  in
-  { functions }
-
-let assert_extern_module : Symbolic.P.extern_func Link.extern_module =
-  let positive_i32 (i : Value.int32) : unit Choice.t =
+let assume_extern_module : Symbolic.P.extern_func Link.extern_module =
+  let assume_positive_i32 (i : Value.int32) : unit Choice.t =
```
zapashcanon, line 13: You can merge all of this inside a single `symbolic` module.

test/sym/binop_i32.wast
```diff
@@ -0,0 +1,31 @@
...
+
+  (func $start
+    (local $x i32)
+    (local $res i32)
+    
+    (local.set $x (call $i32_symbol))
+
+    (local.set $res (i32.add (local.get $x) (i32.const 42)))
+    (if (i32.eqz (local.get $res))
+      (then unreachable))
```
zapashcanon, line 12: actually, this could probably be rewritten as:

```wat
(i32.add (local.get $x) (i32.const 42))
i32.eqz
(if (then unreachable))
```

To avoid having a `res` variable.

test/sym/assume.t
```diff
@@ -1,3 +1,8 @@
-assume extern module:
+symbolic extern module (assume and assert test):
   $ dune exec owi -- sym assume.wast
+  CHECK:
+  (i32.not (i32.to_bool symbol_0))
+  (i32.ge symbol_0 (i32 0))
+  (i32.to_bool symbol_0)
+  /CHECK OK
```
zapashcanon, line 7: We don't want this `CHECK ...` stuff, it should be commented in the `.ml` file (I believe it's in `choice_monad.ml` but I did not check.

src/cmd_sym.ml
```diff
@@ -61,6 +51,9 @@ let symbolic_extern_module : Symbolic.P.extern_func Link.extern_module =
     ; ( "assume"
       , Symbolic.P.Extern_func.Extern_func
           (Func (Arg (I32, Res), R0), assume_i32) )
+    ; ( "assume_positive"
```
zapashcanon, line 54: This should still have the name `assume_positive_i32`.

test/optimize/not_exists_run.t
```diff
@@ -0,0 +1,4 @@
+file doesn't exist (run option):
+  $ dune exec -- owi run not_exists.wast --debug --profiling
+  file `not_exists.wast` doesn't exist
+  [1]
```
zapashcanon, line 4: This should probably be in `test/run/`.

test/optimize/not_exists_script.t
```diff
@@ -0,0 +1,4 @@
+file doesn't exist (script option):
+  $ dune exec -- owi script not_exists.wast --debug --profiling
+  file `not_exists.wast` doesn't exist
+  [1]
```
zapashcanon, line 4: This should probably be in `test/script/`.

test/sym/assume.wast
```diff
@@ -0,0 +1,17 @@
...
+  (import "symbolic" "assume" (func $assume_i32 (param i32)))
+  (import "symbolic" "assume_positive_i32" (func $positive_i32 (param i32)))
+  (import "symbolic" "assert" (func $assert_i32 (param i32)))
+
+  (func $start
+    (local $x i32)
+    (local.set $x (call $i32_symbol))
+    (call $assume_i32 (local.get $x))
+    (call $positive_i32 (local.get $x))
+    (call $assert_i32 (local.get $x))
```
zapashcanon, line 12: It would be nice to have different tests for each of them. Here using `positive_i32` and `assume_i32` on the same variable is redundant and doesn't guarantee proper testing of both.

epatrizio: It should be better now
https://github.com/OCamlPro/owi/pull/96/commits/d877c3859fb00b78142516c0cbb397c53baec90f

test/sym/unop.t
```diff
@@ -40,109 +56,172 @@ unop i64:
...
-      (symbol_1 f64 (f64 nan))
+      (symbol_1 f64 (f64 1.031494))
       (symbol_2 i32 (i32 1))
       (symbol_0 f32 (f32 nan)))
   Trap: unreachable
   Model:
     (model
-      (symbol_1 f64 (f64 606188394992561843776909558182978781184.000000))
-      (symbol_2 i32 (i32 33930))
+      (symbol_1 f64 (f64 -338460706455329128135018695364373065576360054171204343201037173805357995568625611429596369334345869072216864991017687935090688.000000))
```
zapashcanon, line 155: Hahaha. We should probably try to use another printing function for big floats like this. I guess OCaml has a way to print floats in scientific notation. cc @filipeom

filipeom: Hahah, yes this isn't an easy number to decode :sweat_smile: I can fix this directly in the encoding, it should be easy.

test/sym/assume.wast
zapashcanon: Well, I still believe it's not enough. Doing `positive_i32` and `assert_i32` on the same variable seems quite unlikely to happen in real code and it is still not clear what this is testing.

filipeom: May I offer a suggestion? Essentially `positive_i32` is just a more efficient way to write: 

```wast
(call $assume_i32 (i32.ge (i32.const 0) (local.get $x)))
``` 

We can easily encode such tests that prove integer overflow problems:
 
$x \geq 0  \wedge y > 1 \Rightarrow (x + y) > 0$

This inequality translates to the following WebAssembly program:

```wast
  (func $start
    (local $x i32)
    (local $y i32)
    (local.set $x (call $i32_symbol))
    (local.set $y (call $i32_symbol))
    (call $positive_i32 (local.get $x)) ;; x >= 0
    (call $assume_i32 (i32.gt_s (local.get $y) (i32.const 0)) ;; y > 0
    (if (i32.le_s (i32.add (local.get $x) (local.get $y)) (i32.const 0))
      (then unreachable))
    ;; Alternatively
    (call $assert_i32 (i32.gt_s (i32.add (local.get $x) (local.get $y)) (i32.const 0)))
)
```

zapashcanon: So your suggestion is to remove the function ? It thinks it's nice to have it as an illustration for some examples.

test/sym/unop.t
zapashcanon: Ooops, scientific notation is not going to help here, I meant hexadecimal.

test/sym/assume.wast
filipeom: Well, perhaps I sounded dismissive of `positive_i32`, but that was not my intention. I believe it is useful.

In my comment, I'm suggesting an alternative way to rewrite this test while maintaining its existing structure—specifically, using `positive_i32` once, `assume_i32` once, and `assert_i32` once. Instead of directly using `assert_i32` on a variable, I propose applying it to the expression $x + y > 0$. This allows the assert to quickly find the overflow bug, showing its correct behaviour.

test/sym/unop.t
filipeom: You don't like [this](https://github.com/formalsec/encoding/blob/main/test/test_qf_bvfp.expected) representation? The `Model.pp` is just for display purposes, the model we output in the `.xml` files come in hexadecimal so there won't be any loss in precision.

But I'm not opposed to print these numbers in hexadecimal. Do you think every IEEE float and bitvec integer should be printed in hexadecimal, or just the big ones?

zapashcanon: Yes I was worried about not being able to get the exact value. When is the `.xml` file generated (I don't remember sorry) ? I'm fine with the current display if there's a way to get the exact value somehow. :)

test/sym/assume.wast
zapashcanon: Oh okay my bad, I read to quickly. Indeed, I think the proposed program makes sense as a test. :-)

test/sym/unop.t
filipeom: It's generated at the end of every feasible branch [here](https://github.com/OCamlPro/owi/blob/71fce491657e5465b715debfab7f045a1b4c1e51/src/cmd_sym.ml#L307). 

In the encoding there's two functions to print values: `pp` that displays floats regularly and `pp_num` that displays fps and bitvectors in hexadecimal. But I should probably improve this and create a more explicit function for this purpose

test/sym/assume.wast
epatrizio: Thanks for feedbacks @filipeom and @zapashcanon 
It was a bit difficult for me to write the right test ;)

test/sym/binop_f64.wast
```diff
@@ -0,0 +1,47 @@
...
+
+    (f64.eq (f64.min (local.get $x) (f64.const 42)) (f64.const 0))
+    (if (then unreachable))
+
+    (f64.eq (f64.max (local.get $x) (f64.const 42)) (f64.const 0))
+    (if (then unreachable))
+
+    ;; File "src/symbolic_value.ml", line 508, characters 24-30: Assertion failed
+    ;; (f64.eq (f64.copysign (local.get $x) (f64.const 42)) (f64.const 0))
+    ;; (if (then unreachable))
```
zapashcanon, line 41: I believe you still have to open an issue so we can track this.

test/sym/unop_f32.wast
```diff
@@ -0,0 +1,62 @@
...
+    (if (then unreachable))
+
+    (f32.ge (f32.abs (local.get $x)) (f32.const 42))
+    (if (then unreachable))
+
+    (f32.le (f32.neg (local.get $x)) (f32.const 42))
+    (if (then unreachable))
+
+    ;; owi: internal error, uncaught exception:
+    ;;  File "src/symbolic_value.ml", line 420, characters 18-24: Assertion failed
```
zapashcanon, line 34: Same here.

test/sym/binop_f32.wast
```diff
@@ -0,0 +1,47 @@
...
+    (f32.eq (f32.div (local.get $x) (f32.const 42)) (f32.const 0))
+    (if (then unreachable))
+
+    (f32.eq (f32.min (local.get $x) (f32.const 42)) (f32.const 0))
+    (if (then unreachable))
+
+    (f32.eq (f32.max (local.get $x) (f32.const 42)) (f32.const 0))
+    (if (then unreachable))
+
+    ;; File "src/symbolic_value.ml", line 436, characters 24-30: Assertion failed
```
zapashcanon, line 39: Same.

test/sym/memory.wast
```diff
@@ -61,6 +62,12 @@
   )
 
   (func $start
+    i32.const 1
+    memory.grow
```
zapashcanon, line 66: I believe this is not needed because `(memory $m 1)` already gives a size 1 to the memory as it's the minimal size. :-)

zapashcanon: And there's no symbolic value in this test. :-)

test/sym/table.wast
```diff
@@ -0,0 +1,27 @@
...
+    (call_indirect
+      $t
+      (type $type)
+      (i32.const 0)
+    )
+    drop
+  )
+
+  (start $start)
+)
```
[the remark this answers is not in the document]
epatrizio: I've added a i32 symbol for the table function index in `call_indirect` instruction

test/sym/memory.wast
epatrizio: I've tried a `memory.grow` with a symbolic value :
```console
  Failure("Unsupported symbolic value reasoning over \"(i32.wrap_i64 (i64.mul (i64.extend_i32_s symbol_0) (i64 65536)))\"")
```
What type of error is it?

epatrizio: Can i perform a memory load or store at a symbolic position ?

filipeom: > What type of error is it?

It's related to your second question:

Currently no, but I just started working on it. I'll opt to concretize addresses for now. I'll to submit a PR today :smiley:

zapashcanon:
```quote
I've tried a memory.grow with a symbolic value :
```

Yes, I know it's not working with a symbolic value, but what I meant is that the test is useless without the symbolic value (because we are already testing this) so you can basically remove the test for now.

test/sym/table.wast
```diff
@@ -0,0 +1,27 @@
...
+  (type $type (func (param i32) (result i32)))
+
+  (func $f (param $x i32) (result i32)
+    (if (i32.eqz (local.get $x))
+      (then unreachable))
+    local.get $x
+  )
+
+  (func $start
+    call $i32_symbol
```
zapashcanon, line 17: What is this symbol used for ? Can't you simply remove it ?

test/sym/table.wast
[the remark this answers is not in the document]
zapashcanon: Hum OK. I read too quickly the first time.

I suggest you change the type from `(param i32) (result i32)` to `(param i32)`, remove the final `local.get $x` in `$f` and the `drop` in `$start`.

It's generally better to write tests as small as possible, that do something meaningful and easy to understand.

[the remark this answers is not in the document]
epatrizio: https://github.com/OCamlPro/owi/pull/96/commits/bb7beeef10445db9cd758e49470555def79bc1dd

test/sym/table.t
```diff
@@ -0,0 +1,15 @@
...
+  Trap: undefined element
+  Model:
+    (model
+      (symbol_1 i32 (i32 -2147483648)))
+  Trap: unreachable
+  Model:
+    (model
+      (symbol_0 i32 (i32 0))
+      (symbol_1 i32 (i32 0)))
+  Reached 2 problems!
```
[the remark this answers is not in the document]
epatrizio: Could the problem come from `Trap: undefined element` ? (`call_indirect` at a table symbolic index)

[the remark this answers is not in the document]
epatrizio: https://github.com/OCamlPro/owi/issues/107

test/sym/binop_i32-2.wast
```diff
@@ -0,0 +1,42 @@
...
+    (if (then unreachable))
+
+    (i32.eqz (i32.shr_s (local.get $z) (local.get $w)))
+    (if (then unreachable))
+
+    (i32.eqz (i32.shr_u (local.get $z) (local.get $w)))
+    (if (then unreachable))
+
+    ;; owi: internal error, uncaught exception:
+    ;;  Failure("z3_mappings: rotl|rotr not implemented!")
```
zapashcanon, line 34: could you open an issue on encoding for this ?

zapashcanon:
On the main branch there is a `test/sym/grow.wast` now, once you re-based don't forget to add it to the list of cram tests. :)

test/sym/binop_i32-2.wast
filipeom: This should be in https://github.com/formalsec/encoding/commit/87954e4a269551422293ad69f3df6d68ee265619 but I haven't tested it

filipeom:
> On the main branch there is a `test/sym/grow.wast` now, once you re-based don't forget to add it to the list of cram tests. :)

Or just create a test to check for out of bounds with a symbolic memory grow. I didn't pay much consideration for formatting when writing the test so it might be ugly

test/sym/binop_i32-2.wast
zapashcanon: Thanks!

zapashcanon:
Merged as part of #119.

PR #100, opened 2024-01-16
add `owi c` command and examples for all subcommands

zapashcanon:
cc @filipeom


src/libc/src/memcmp.c
```diff
@@ -0,0 +1,21 @@
...
+  ++count;
+  while (--count) {
+    // TODO: check this change !
+    r = (*d - *s);
+    if (r) {
+      return r;
+    }
+    ++d;
+    ++s;
+  }
```
zapashcanon, line 19: @filipeom @chambart 

The original code was:

```c
  while (--count) {
    if (r = (*d - *s))
      return r;
    ++d;
    ++s;
  }
```

But it was raising annoying warnings saying we should add parentheses. I guess it was suggesting using `if ( (r = (*d - *s)) )`, I tried to make the code even more clearer. Could one of you confirm it is correct ? It's been a very long time I had to write C. :-)

src/thread.ml
```diff
@@ -29,7 +29,7 @@ let globals t = t.globals
 let solver_mod : Solver.t solver_module = (module Solver)
 
 let create () =
-  let solver = S (solver_mod, Solver.create ~logic:QF_BVFP ()) in
+  let solver = S (solver_mod, Solver.create ~logic:QF_BV ()) in
```
filipeom, line 32: Is this intentional? I'm only worried that when mixing floats with bvs Z3 will return `Unknown` as there's no logic to solve FPA

filipeom (approved):

src/libc/src/memcmp.c
filipeom: Looks equivalent to me :smiley:

src/libc/include/setjmp.h
```diff
@@ -0,0 +1,8 @@
+#ifndef _SETJMP_H_
+#define _SETJMP_H_
+
+struct __jmp_buf_tag {};
```
zapashcanon, line 4: I tried to enable `-Wall -pedantic-errors` and I'm getting the following:

```clang
include/setjmp.h:4:1: error: empty struct is a GNU extension [-Werror,-Wgnu-empty-struct]
```
It is indeed undefined behavior. I guess we should get rid of that but I'm not sure about how we should proceed and I'd like to discuss this with you.

src/thread.ml
zapashcanon: Oops sorry. For some reason I was getting an error on latest encoding and changed this without knowing what I was doing. :sweat_smile:

zapashcanon:
As you can see, I got rid of the Makefile, this allows to build everyting with `dune` and a small script in OCaml (`src/libc/script.ml`).

I also moved all the dune-site stuff directly where the files live (for instance we have a `src/libc/include/dune` now).

I also moved the instrumentor stuff into `libc`, it is not really related to the libc but let's say this folder is for all C related stuff.

src/libc/include/setjmp.h
filipeom: Yeah, this is because I just couldn't be bothered with the complexity of https://github.com/ensc/dietlibc/blob/master/include/setjmp.h it's also highly tied to the underlying architecture so I think it's non trivial to fully fix. Although, in the short term I could complete the struct so as to not throw compilation warnings.

Also, in the future I think it would also be best for use to move to `wasi-libc`

owi.opam.template
```diff
@@ -1 +1,4 @@
 available: (arch = "x86_32" | arch = "x86_64" | arch = "arm64") & os != "win32"
+depexts: [
+  ["llvm" "clang" "lld" "wabt"]  {os-family = "debian"}
```
filipeom, line 3:
```suggestion
  ["llvm" "clang" "lld" "wabt"]  {os-family = "debian"}
  ["llvm" "wabt"]  {os-family = "osx"}
```
CI is failing, and it appears to be because we're not specifying the dependencies here. However, I'm not quite sure if these are the only packages required for macOS. Additionally, I couldn't find mentions of `lld` or `clang` in Homebrew packages, so I assume they must be included in `llvm`.

zapashcanon:
I removed the Quick-start section of the README and wrote a short example of all sub-commands we have. It is much better now IMO.

src/libc/include/setjmp.h
zapashcanon: cc @chambart who is a fan of setjmp and longjmp I'm sure :-)

dune-project
```diff
@@ -1,26 +1,28 @@
...
+
 (name owi)
 
 (license AGPL-3.0-or-later)
 
 (authors
  "Léo Andrès <contact@ndrs.fr>"
- "Pierre Chambart <pierre.chambart@ocamlpro.com>")
+ "Pierre Chambart <pierre.chambart@ocamlpro.com>"
+ "Filipe Marques")
```
zapashcanon, line 19: cc @filipeom, I believe you quite deserved to be listed as an author now :smile: 

If you want to add an e-mail there, you can either tell me which one, either add it after this get merged.

owi.opam.template
zapashcanon: Thanks. I added this line directly myself as running `dune b owi.opam` was needed to generate the new `owi.opam` file...

Let's see if this is enough :sweat_smile:

dune-project
filipeom, line 19: Well thank you so much for the consideration! :smiley: You can just commit this suggestion if you'd like:

```suggestion
 "Filipe Marques <filipe.s.marques@tecnico.ulisboa.pt>")
```

owi.opam.template
filipeom: This appears to have done nothing :sweat_smile: 

I was looking at the `os-family` available in [here](https://github.com/ocaml-opam/opam-depext#opam-depext-plugin) and perhaps wrongfully assumed `osx` would work. Inspecting the CI, it looks like it's looking for `os-family = "homebrew"`? I have no current way to test this tho, sorry.

```sh
...
# Detecting depexts using vars: arch=x86_64, os=macos, os-distribution=homebrew, os-family=homebrew
...
```

src/libc/include/setjmp.h
zapashcanon: will fix later, see #103

owi.opam.template
zapashcanon: Well, still not. Let's merge as the CI is green on Linux. We'll fix macOS later.

filipeom: At least it's installing `wabt` and `llvm` not. But it's still saying `opt: command not found`

zapashcanon: Yes I saw... @krtab is going to have a look.

PR #105, opened 2024-01-17
Fixes deploy workflow and removes unecessary open

filipeom:
Deploy workflow does not build `libc` that's why it fails

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -29,6 +29,7 @@ jobs:
           mv _build/default/_doc/_html doc/api
       - name: coverage
         run: |
+          opam exec -- dune build @install
           BISECT_FILE=$(pwd)/bisect opam exec -- dune runtest --force --instrument-with bisect_ppx
           opam exec -- bisect-ppx-report html -o doc/coverage
       - name: coverage-badge
--- a/src/symbolic_value.ml
+++ b/src/symbolic_value.ml
@@ -169,8 +169,6 @@ module S = struct
   end
 
   module I32 = struct
-    open Expr
-
     type num = Expr.t
 
     type vbool = Expr.t


zapashcanon:
Oh thanks!

Could you remove my failed tentative fix in b56044ee8ce9c90c4ea2d31f1204fa1e1e00a73e ?

Also, I believe it would be better to add the `dune build @install` part in the `coverage` section (before `BISECT_FILE=... opam exec -- dune runtest`).

But maybe it also means that our dune rules are wrong and we should say somehow that they depend on `libc` being built instead of adding this ? What do you think ?

filipeom:
> Could you remove my failed tentative fix in [b56044e](https://github.com/OCamlPro/owi/commit/b56044ee8ce9c90c4ea2d31f1204fa1e1e00a73e) ?

Yes of course :)
 
> Also, I believe it would be better to add the `dune build @install` part in the `coverage` section (before `BISECT_FILE=... opam exec -- dune runtest`).

I can do this as well

> But maybe it also means that our dune rules are wrong and we should say somehow that they depend on `libc` being built instead of adding this ? What do you think ?

I agree, I'd expect it when we do `dune exec` that the lib would be built as well. This would remove the need for this "fix". I can look into this quickly to see if I can solve it :)

filipeom:
This is my understanding of the problem atm:

0. To reproduce:`dune clean && dune build @doc && dune exec owi c -- example/c/poly.c`
1. `dune build @doc` does not build `owi` executable nor compiles libc
2. When tests do `dune exec` it will only build Owi's executable and not libc

How can we set an explicit dependency of owi's executable on the artifacts generated by the install stanzas inside `src/libc`?

zapashcanon:
Should be fixed by 73b45024a183032c47b86a689c5c2a75f991f26f

Before this commit, running `dune clean && dune runtest` would fail but it succeeds now.

Quoting the [Dependency specification](https://dune.readthedocs.io/en/stable/concepts/dependency-spec.html) doc:

> `(package <pkg>)` depends on all files installed by `<package>`, as well as on the transitive package dependencies of `<package>`. This can be used to test a command against the files that will be installed.

Now `runtest` will succeed without needing to `build @all` before. If we were to run individual tests (I don't even know how to do this) they would other would probably fail because they will lack a dependency on the install file. But this is probably good enough for now...

PR #108, opened 2024-01-19
fix macos build in CI

zapashcanon:
maybe, or not


zapashcanon:
looks like it was rater a not :-)

filipeom:
I think I already tried `echo "/usr/local/opt/llvm/bin" >> $GITHUB_PATH` and it also failed :disappointed:

Issue #99, opened 2024-01-15
Unsupported symbolic value reasoning \"(++ symbol_0 (++ symbol_0 (++ symbol_0 symbol_0)))\"")


krtab:
Hi @filipeom 
With the added patch of https://github.com/formalsec/encoding/pull/50 I get the following error

```
owi: internal error, uncaught exception:
     Failure("Unsupported symbolic value reasoning over \"(++ symbol_0 (++ symbol_0 (++ symbol_0 symbol_0)))\"")
```

Wat file:  
[rust-owi.txt](https://github.com/OCamlPro/owi/files/13941593/rust-owi.txt)

filipeom:
It's trying to `store`/`load` values with a symbolic address. I should have already ported wasp's [memory models](https://github.com/formalsec/wasp-private/tree/memory-refactor/wasp/lib/symbolic/static/memory) here but I haven't had the time :disappointed:. I'll make the symbolic interpreter concretise for now

krtab:
> It's trying to `store`/`load` values with a symbolic address. I should have already ported wasp's [memory models](https://github.com/formalsec/wasp-private/tree/memory-refactor/wasp/lib/symbolic/static/memory) here but I haven't had the time 😞. I'll make the symbolic interpreter concretise for now

Could you give me access to the repo ?

filipeom:
I should have linked the new repo here https://github.com/formalsec/memory-models. But I'll give you access to both

krtab:
Closing in favor of #182

PR #109, opened 2024-01-19
Concretise symbolic memory addresses and allow symbolic heap sizes

filipeom:
## Two things this PR addresses

1. **Allow Symbolic Heap Sizes:** `memory.grow` now accepts symbolic arguments (example in [grow.wast](https://github.com/formalsec/owi/blob/concretise/test/sym/grow.wast)).

2. **Concretisation of Symbolic Memory Accesses:** This results in an under-approximating analysis, meaning that we accept some unsoundness for increased completeness.

Example of a scenario we might fail to encounter:

   ```c
   #include <owi.h>
   int main() {
     int i = owi_i32();
     int n = 16;
     owi_assume(i > 0);

     int *chunk = (int *)malloc(sizeof(int) * n);
     chunk[i] = 0;

     return 0;
   }
   ```

Depending on the concretisation of `i`, we might not flag the heap buffer overflow vulnerability. However, previously, we would encounter a failure.

But, we can always introduce some form of branching here to try to mitigate this.

## What We Still Don't Support

Symbolic allocations, i.e., using `malloc(symbolic_size)`, are not yet supported. This is because it would make the `__heap_base` (i.e., `bump_pointer` in `malloc`) symbolic. However, I can also concretise the interactions with `malloc`. @krtab, please let me know if this is required for #99.

### DIFF
--- a/src/symbolic_memory.ml
+++ b/src/symbolic_memory.ml
@@ -138,57 +138,33 @@ module M = struct
     let v0 = load_byte m a in
     loop a n 1 v0
 
-  (* TODO: *)
-  (* 1. Let pointers have symbolic offsets *)
-  (* 2. Let addresses have symbolic values *)
-  let calculate_address m (a : int32) : (Int32.t, Trap.t) Stdlib.Result.t =
-    match a.e with
-    | Val (Num (I32 i)) -> Ok i
-    | Ptr (base, offset) -> (
-      match Hashtbl.find m.chunks base with
-      | exception Not_found -> Error Trap.Memory_leak_use_after_free
-      | size ->
-        let ptr = Int32.add base (i32 offset) in
-        if ptr < base || ptr > Int32.add base (i32 size) then
-          Error Trap.Memory_heap_buffer_overflow
-        else Ok ptr )
-    | _ -> Log.err {|Unable to calculate address of: "%a"|} Expr.pp a
-
-  let load_8_s m a =
-    let+ a = calculate_address m a in
-    let v = loadn m a 1 in
+  let load_8_s m (a : int32) =
+    let v = loadn m (i32 a) 1 in
     match v.e with
     | Val (Num (I8 i8)) -> Value.const_i32 (Int32.extend_s 8 (Int32.of_int i8))
     | _ -> Cvtop (ExtS 24, v) @: Ty_bitv S32
 
-  let load_8_u m a =
-    let+ a = calculate_address m a in
-    let v = loadn m a 1 in
+  let load_8_u m (a : int32) =
+    let v = loadn m (i32 a) 1 in
     match v.e with
     | Val (Num (I8 i)) -> Value.const_i32 (Int32.of_int i)
     | _ -> Cvtop (ExtU 24, v) @: Ty_bitv S32
 
-  let load_16_s m a =
-    let+ a = calculate_address m a in
-    let v = loadn m a 2 in
+  let load_16_s m (a : int32) =
+    let v = loadn m (i32 a) 2 in
     match v.e with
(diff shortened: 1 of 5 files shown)


src/symbolic_memory.ml
```diff
@@ -38,27 +38,30 @@ module M = struct
     | _ -> Log.err {|Unsupported symbolic value reasoning over "%a"|} Expr.pp v
 
   let grow m delta =
-    let delta = Int32.to_int @@ i32 delta in
-    let old_size = m.size * page_size in
-    m.size <- max m.size ((old_size + delta) / page_size)
+    let old_size = Value.I32.mul m.size page_size in
+    let new_size = Value.I32.(div (add old_size delta) page_size) in
```
zapashcanon, line 42: I see the `max` is not here anymore. Could you explain me why?

zapashcanon: Oh actually, it's encoded through the `Triop (Ite` right ?

src/symbolic_memory.ml
```diff
@@ -38,27 +38,30 @@ module M = struct
...
-  let size_in_pages { size; _ } = Value.const_i32 @@ Int32.of_int @@ size
+  let size_in_pages { size; _ } = size
 
   let fill _ = assert false
 
   let blit _ = assert false
 
   let blit_string m str ~src ~dst ~len =
-    (* Always concrete? *)
+    (* This function is only used in memory init so everything will be concrete *)
```
zapashcanon, line 57: OK thus maybe we should change the expected type in the interface to avoid re-using it one day in a symbolic case ?

src/symbolic_memory.ml
```diff
@@ -38,27 +38,30 @@ module M = struct
...
+    let str_len = String.length str in
+    let mem_len = Int32.(to_int (i32 m.size) * to_int (i32 page_size)) in
     let src = Int32.to_int @@ i32 src in
     let dst = Int32.to_int @@ i32 dst in
     let len = Int32.to_int @@ i32 len in
     if
-      src < 0 || dst < 0
-      || src + len > String.length str
-      || dst + len > m.size * page_size
+      src < 0 || dst < 0 || len < 0
```
zapashcanon, line 64: Why is this needed ?

src/symbolic_memory.ml
filipeom: Yes, because it's symbolic we might need to encode the greater size in a Ite expression. But I use `Value.Bool.select_expr` so as to not create concrete Ite expressions

src/symbolic_memory.ml
filipeom: Yes, this would be better to avoid accidents :)

src/symbolic_memory.ml
filipeom: It's checking for out of bounds. I tried being loyal to the [concrete version](https://github.com/OCamlPro/owi/blob/main/src/concrete_memory.ml#L58)

zapashcanon: Oh indeed. This should probably be moved into the interpreter anyway (I believe there's some `TODO` in the code).

filipeom: Yes there is. I can first make a PR moving the out of bounds checks of `Memory.blit_string`, `Memory.blit`, and `Memory.fill` to the interpreter if you'd like.

src/symbolic_memory.ml
zapashcanon: Yes, please do. :)

zapashcanon: OK actually, this is likely to be true in C compiled programs but in a Wasm program one can write a `memory.init` in the middle of a program with a symbolic value. Let's keep it like this.

PR #110, opened 2024-01-19
Add missing crunch dependency

krtab:


dune-project
```diff
@@ -55,6 +55,7 @@
     (>= "0.1")
     :dev))
   (odoc :with-doc)
+  crunch
```
zapashcanon, line 58: Thanks! Could you mark it as a `dev` dependency ? `(crunch :dev)` should be enough.

zapashcanon:
Thanks!

PR #112, opened 2024-01-19
Tentative CI fix on macOS

krtab:


krtab:
Well it is going further ! But now, we lack wasm-ld.

zapashcanon:
Thanks! I included your commit in #114

PR #104, opened 2024-01-17
[owic] fix malloc alignement

epatrizio:

PR #118, opened 2024-01-22
Add malloc align test

krtab:
Add a test for malloc well-alignedness.

The test is pretty long to run (8 sec on my machine).

Depends on #104

### DIFF
--- a/test/c/dune
+++ b/test/c/dune
@@ -1,3 +1,2 @@
 (cram
- (deps
-  %{bin:owi}))
+ (deps %{bin:owi}))
--- a/test/owic/stdlib/dune
+++ b/test/owic/stdlib/dune
@@ -1,0 +1,8 @@
+(rule
+ (alias runtest)
+ (deps
+  malloc_aligned.c
+  %{bin:owi}
+  (package owi))
+ (action
+  (run dune exec owi -- c ./malloc_aligned.c)))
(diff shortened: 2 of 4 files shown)


zapashcanon:
Thanks! Do you want to try adding a dune rule yourself so that it becomes part of `dune runtest` or should I do it ?

krtab:
I'm not sure we should add it given its large run time.

filipeom:
Thanks for the test @krtab!! There's a bug regarding the `Ptr` type returned by `owi_alloc`. I just ran this locally, and with `-w 8` owi is making 8200 calls to the solver, totaling  34s in the solver. This means that the expressions in `owi_assert`, i.e., stuff like this:

```
calling func : func owi_assert
stack        : [  ]
running instr: local.get 0
stack        : [ (i32.of_bool (i32.eq (i32.rem (Ptr (i32 8390672) (i32 0)) (i32 1)) (i32 0))) ]
```

Is being discharged to the solver instead of being solved concretely

krtab:
https://github.com/OCamlPro/owi/pull/118/commits/ee66c6ad020fe6bd79c81f939782db036beadaca is actually in #127  which should be merged first

krtab:
Performance issue fixed in https://github.com/formalsec/encoding/pull/65

zapashcanon:
Could you run `dune fmt` and commit the formatted `dune` file ? :-)

Otherwise, should I merge this ? It's all OK for me.

krtab:
> Otherwise, should I merge this ? It's all OK for me.

Yes, the slight performance issue will be fixed once we merge hashconsing in encoding so :green_circle:

PR #119, opened 2024-01-22
test stuff

zapashcanon:
this is a rebase of #96 along with various improvements to the testsuite which was a big mess as I never took the time to clean it a bit


filipeom:
I'm a bit worried about the constant fluctuation of some values in the outputted models between commits. Without hard constraints on the actual values two runs of the same test might produce different models because of the solver's RNG.

zapashcanon:
Yes, I saw this. We could either use non-deterministic cram tests or try to use the solver in a deterministic mode when testing ?

filipeom:
> Yes, I saw this. We could either use non-deterministic cram tests or try to use the solver in a deterministic mode when testing ?

I like the current setup for testing, but how would the non-deterministic cram tests look like?
 
I think Z3 has some parameters to set the solver seed, but it's not clear to me how it works. I can look into this and come back to you

zapashcanon:
I believe we could simply replace a line that changes too much by `...` and that would be enough in the current setting. :)

Another solution would be to rewrite the changing tests so that they don't vary to much (by adding many `assume` for instance).

Also, I opened a PR on encoding to start move towards a more stable output.

filipeom:
> Another solution would be to rewrite the changing tests so that they don't vary to much (by adding many `assume` for instance).

Yeah, I think probably adding assumes to the tests that have multiple solutions is sufficient

test/sym/assume.wat
```diff
@@ -4,18 +4,21 @@
...
-    ;; Alternatively (x + y > 0)
+    (call $assume_i32 (i32.gt_s (local.get $y) (i32.const 0))) ;; y > 0
+
+    ;; if x + y <= 0
+    (if (i32.le_s (i32.add (local.get $x) (local.get $y)) (i32.const 0))
+      ;; possible if they overflow
+      (then unreachable))
+
+    ;; otherwise (x + y > 0)
     (call $assert_i32 (i32.gt_s (i32.add (local.get $x) (local.get $y)) (i32.const 0)))
```
filipeom, line 21: This is essentially testing the same thing. assert (x + y > 0) is the same as checking the satisfiability of (x + y <= 0)

zapashcanon: Yes I believe this tests was written to ensure we are adding the correct path condition when testing the condition in an `if then else`.

filipeom: I believe this was the same test I proposed to modify in https://github.com/OCamlPro/owi/pull/96#discussion_r1451038234. What I'm saying it that having:

```wast
  (if (i32.le_s (i32.add (local.get $x) (local.get $y)) (i32.const 0))
      (then unreachable))
```

Is equivalent to just writing: 

```wast
(call $assert_i32 (i32.gt_s (i32.add (local.get $x) (local.get $y)) (i32.const 0)))
```

And because this program always overflows the assert will never be executed. However, I agree that this test can also test what you're saying. But for that we would need to change `assume.t` to include `--no-stop-at-failure`:

```
symbolic extern module (assume and assert test):
  $ dune exec owi -- sym assume.wat --no-stop-at-failure
```

zapashcanon: > However, I agree that this test can also test what you're saying. But for that we would need to change assume.t to include --no-stop-at-failure:

Indeed, this was missing... I understand the confusion now. Will fix.

filipeom:
Could this error be why it's not pinning the encoding correctly? 
```
 [WARNING] Failed checks on owi package definition from source at file:///home/runner/work/owi/owi:
      error  3: File format error in 'pin-depends' at line 64, column 4: while expecting versioned package: OpamPackage.of_string
  owi is now pinned to git+file:///home/runner/work/owi/owi#HEAD (version dev)
```

zapashcanon:
Haha, yes, I was fixing it, I can never remember the syntax of this stuff..

filipeom:
> Haha, yes, I was fixing it, I can never remember the syntax of this stuff..

I was reading the manual and I think you were just missing the `.VERSION` per your last commit. But I also like your idea of pinning it to the last commit known to not be broken :sweat_smile:

filipeom:
Btw there's no encoding target in `owi.opam`

filipeom:
> I was hoping it was not necessary to add a package to `depends` if it was in `pin-depends` looks like I was wrong...

Yeah It looks redundant having to specify both, but there's probably a reason for it? :sweat_smile:

zapashcanon:
Yes, I asked on opam https://github.com/ocaml/opam/issues/5795

zapashcanon:
Anyway, CI is green, let's merge this so that anyone can easily keep working on all of this. :sweat_smile:

PR #125, opened 2024-01-23
Refactor module p

epatrizio:
Preparing work for fuzzing (done with @chambart)

### DIFF
--- a/src/symbolic.ml
+++ b/src/symbolic.ml
@@ -6,7 +6,30 @@ open Types
 open Hc
 module Solver = Thread.Solver
 
-module P = struct
+module type Thread = sig
+  type t
+
+  type 'a solver_module = (module Encoding.Solver_intf.S with type t = 'a)
+
+  type solver = S : 'a solver_module * 'a -> solver
+
+  val memories : t -> Symbolic_memory.memories
+
+  val tables : t -> Symbolic_table.tables
+
+  val globals : t -> Symbolic_global.globals
+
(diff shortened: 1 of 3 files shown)


src/symbolic.ml
```diff
@@ -33,7 +55,7 @@ module P = struct
 
   type thread = Thread.t
 
-  module Choice = Symbolic_choice.MT
+  module Choice = Choice_monad
```
zapashcanon, line 58: This can be removed if you rename the functor parameter `Choice_monad` to `Choice`.

src/symbolic.ml
```diff
@@ -6,7 +6,29 @@ open Types
 module Def_value = Concrete_value
 module Solver = Thread.Solver
 
-module P = struct
+module type Thread = sig
+  type t
+
+  type 'a solver_module = (module Encoding.Solver_intf.S with type t = 'a)
+  type solver = S : 'a solver_module * 'a -> solver
```
krtab, line 13: This and

```ocaml
  val solver: t -> solver

  val pc: t -> Symbolic_value.S.vbool list
```

were added unsupervised by me and @epatrizio so you may want to use double the amount of caution when reviewing.

zapashcanon: I'm not sure this is what we want, but you'll find out while writing the fuzzer part. :-)

krtab:
Can you run `dune fmt` ?

krtab (approved):

Issue #129, opened 2024-01-26
Killing owi c doesn't kill owi sym


krtab:
Run `dune exec owi -- c test/owic/stdlib/malloc_aligned.c`, ctrl-c it before it finished, the `owi sym` process remains alive.

(You may need to test on a c file that does an infinite loop to have time to crtl c it)

zapashcanon:
Haha, this one should not be too hard to fix if you are interested.

In `cmd_c.ml` you have a function `let owi ..` that calls `owi sym` directly through a command instead of using the appropriate module. This is a leftover from when they were in a separate binary.

PR #133, opened 2024-01-27
call cmd_owi from cmd_c directly instead of running a command, add all

zapashcanon:
the cmd_owi cli parameters to cmd_c, use fpath
where possible, use Bos.OS.File instead of Stdlib.{open_in,open_out}, remove a weird Obj.magic


src/cmd_sym.ml
```diff
@@ -168,24 +168,24 @@ let simplify_then_link_then_run ~unsafe ~optimize (pc : unit Result.t Choice.t)
...
+  let rec get_intermediate_dirs (d : Fpath.t) acc =
+    if Sys.file_exists (Fpath.to_string d) then acc
+    else
+      get_intermediate_dirs
+        (Fpath.normalize d |> Fpath.split_base |> fst)
+        (d :: acc)
   in
   let intermediate_dirs = get_intermediate_dirs dir [] in
-  List.iter (fun d -> Sys.mkdir d 0o755) intermediate_dirs
+  List.iter (fun d -> Sys.mkdir (Fpath.to_string d) 0o755) intermediate_dirs
```
filipeom, line 188: I guess this function can be removed in favour of `OS.Dir.create ~path:true`

zapashcanon: Indeed. Done (with others along the way) :-)

zapashcanon:
needs #134 to pass

zapashcanon:
Merging as this is failing only because `Domain.recommended_domain_count ()` is `3` in macOS CI, which makes the manpage cram test fail...

zapashcanon:
(@krtab, if you could confirm that it fixes #129)

PR #134, opened 2024-01-27
update for latest encoding that uses Hc for hashconsing

zapashcanon:

PR #136, opened 2024-01-28
fix end-of-stream conflicts

### LINKED ISSUE #84: Fix the conflicts that were introduced recently in the menhir grammar

zapashcanon:
fix #84

`inline_module` was used both inside other rules and as a start symbol. It was lacking an `EOF` token in the start symbol case. Fixed by simply renaming `inline_module` to `inline_module_inner` and by adding `inline_module` which is simply `inline_module_inner` + `EOF`.


zapashcanon:
needs #134 to pass

PR #139, opened 2024-01-29
switch to digestif

### LINKED ISSUE #101: use digestif instead of sha
after #100 is merged

zapashcanon:
Fix #101.

@filipeom, I don't really understand what was the old code doing:

```ocaml
let hash = List.fold_left (fun _ f -> Sha256.file f) Sha256.zero files in
```

It seems like we were discarding the context each time, which does not make sense to me. Also as we are concatenating all the files with: `let file = String.concat " " files in`, I'm not sure we are really computing the right hash. Could you explain why we are doing these two things ?

(This should be rebased after #133 is merged to avoid the `Fpath.v` conversion.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -66,7 +66,7 @@
   (graphics :dev)
   (tiny_httpd :dev)
   (ocamlformat :dev)
-  sha
+  digestif
   xmlm
   pyml
   re2
--- a/owi.opam
+++ b/owi.opam
@@ -34,7 +34,7 @@ depends: [
   "graphics" {dev}
   "tiny_httpd" {dev}
   "ocamlformat" {dev}
-  "sha"
+  "digestif"
   "xmlm"
   "pyml"
   "re2"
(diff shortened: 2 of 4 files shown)


filipeom:
I originally wrote this code for testcomp and later when trying to make the c frondend more generic I ended up generating this tragedy you know seek to understand.

A little context, this code is used to generate the metadata of a benchmark for testcomp and since when it originally was written the input to the c frontend used to only be 1 file. Making this function `metadata` only receive a string which was the `filename`, so the filename was the argument we passed to this function and the hash would simply be `Sha256.file file`.

Later when I make the input to the front

filipeom:
Certainly, here's a refined version:

Some context: Initially, the C frontend exclusively served the purpose of running testcomp, handling only a single argument as input.

Later, when I attempted to make the C frontend more generic by allowing multiple files as input, I gave no consideration to ensuring the correctness of this code. Additionally, since the `metadata` function is specifically relevant to testcomp, where the input is expected to be only one file (albeit now in a list), this code would be acceptable, even though it may not make complete sense.

Considering this, a more appropriate approach might be to provide a list of hashes corresponding to the input files along with a parallel list containing their respective file names?

Issue #116, opened 2024-01-22
something seems to be wrong with br_table


zapashcanon:
See the difference in 48e274b289d706568368d302b154f5bcff6fabc6.

When using a switch (which compiles to a br_table), we do not reach the failure.

When using if then else, it works fine.

Also it seems that executing the `br_table` is very slow (you can see a long pause when running with `--debug`).

PR #140, opened 2024-01-29
Fix symbolic br_table

chambart:
Avoids sharing choice_i32 between different choices...
Also improve the performances by swaping the overflow test and the select_i32
Reverts the maze test to use a br_table

### DIFF
--- a/src/interpret.ml
+++ b/src/interpret.ml
@@ -1422,13 +1422,18 @@ module Make (P : Interpret_intf.P) :
       st stack
     | Br_table (inds, Raw i) ->
       let target, stack = Stack.pop_i32 stack in
-      let* target = Choice.select_i32 target in
-      let target = Int32.to_int target in
-      let target =
-        if target < 0 || target >= Array.length inds then i
+      let> out =
+        Bool.or_
+          I32.(target < const 0l)
+          I32.(target >= const (Int32.of_int (Array.length inds)))
+      in
+      let* target =
+        if out then return i
         else
+          let* target = Choice.select_i32 target in
+          let target = Int32.to_int target in
           let (Raw i) = inds.(target) in
-          i
+          return i
       in
       let state = { state with stack } in
       State.branch state target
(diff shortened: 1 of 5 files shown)


src/interpret.ml
```diff
@@ -1422,13 +1422,18 @@ module Make (P : Interpret_intf.P) :
...
+          I32.(target >= const (Int32.of_int (Array.length inds)))
+      in
+      let* target =
+        if out then return i
         else
+          let* target = Choice.select_i32 target in
+          let target = Int32.to_int target in
           let (Raw i) = inds.(target) in
-          i
+          return i
```
zapashcanon, line 1436: Could you replace the let* by a let+ and remove the return ?

src/symbolic_choice.ml
```diff
@@ -78,11 +78,17 @@ end = struct
         let thread2 = Thread.clone { thread with pc = with_not_v } in
         M.cons (true, thread1) (M.return (false, thread2)) )
 
+  let fresh_symbol_count = ref 0
+  let fresh_symbol name =
+    let n = !fresh_symbol_count in
+    incr fresh_symbol_count;
+    Printf.sprintf "%s_%i" name n
```
zapashcanon, line 85: Shouldn't we use an Atomic here ?

zapashcanon: (also could you replace Printf by Format ? That's what we use everywhere now

zapashcanon:
Fix #116

chambart:
Rebased

krtab:
> Fix #116

I discovered recently that autofix messages **MUST** be in the initial message unfortunately. I've manually linked the issue.

zapashcanon:
@filipeom, is there a way to remove some variables from a model ?

filipeom:
> @filipeom, is there a way to remove some variables from a model ?

We can request the solver for the specific variables we need. With the argument `Solver.model ~symbols`, we can provide a list of symbols for which we want a model. By default, the solver generates a model for all symbols present in the path condition.

I been meaning to improve the way we generate output for models and testsuites to include only the symbols created through our symbolic API. This would also eliminate these auxiliary `choice_i32` variables from being printed. If you want, I can attempt to implement this in a separate PR :smiley:

zapashcanon:
That would be nice, I created an issue so we don't forget about it, we can discuss it there

PR #146, opened 2024-02-05
Solve concrete bounds checks concretely 

### LINKED ISSUE #145: Uncessary solver calls in pointer bounds checks
Some of these may be concrete:

https://github.com/OCamlPro/owi/blob/main/src/symbolic.ml#L172

Results in a 40% slowdown on the Collections-C dataset. Use `Choice.select` instead.

filipeom:
Closes #145


zapashcanon:
Thanks!

PR #147, opened 2024-02-05
simplify choice monad

zapashcanon:


src/choice_intf.ml
krtab, line 1: Most/all of these changes undo things that were done in https://github.com/OCamlPro/owi/pull/125 and are used by @epatrizio for his ongoing fuzzing implem, and hence shouldn't be merged before his work is

krtab (changes requested):
I haven't reviewed everything but I think that this should wait for @epatrizio 's work on fuzzing to be merged anyway

src/choice_intf.ml
zapashcanon: I believe this is completely unrelated changes. The functor is still here in `symbolic.ml`, I restored the `choice_int` to convince you

zapashcanon: The main thing this PR does is removing the useless `Make` functor in `symbolic_choice.ml`, this was instantiated two times (`CList` and `CSeq`), now we only have the `CList` one that has been specialized directly.

zapashcanon:
@krtab, I rebased and refactored a bit more what you added for the fuzzing. Now there's only `Base` and `Complete` in `choice_intf` (no more `Complete_without_run` or `Complete_with_trap`). Both your new `Minimalist` and `MT` (renamed to `Multicore`) implements the `Complete` interface.

Issue #124, opened 2024-01-23
how to make tests more deterministic ?


zapashcanon:
I disabled many tests written by @epatrizio (they are still living in `test/sym/*.disabled`) because the results were changing from one run to the other, making it hard to have them in CI.

We have at least three solutions:

- use the solver in a deterministic mode with a fixed seed, we don't know how easy to do it is or if it can impact performances;
- rewrite all the tests so that they have a clear expected value (I'm not able to tell what is their expected outputs currently as they're quite complicated), it could be done by adding a lot of `assume` but probably requires them to be completely rewritten;
- consider that those tests are simply testing that some instructions are implemented and that we are able to generate a model but that we do not check the symbol values, this would look like #123, unfortunately this is not supported by dune (I opened a feature request https://github.com/ocaml/dune/issues/9812) but this is supported by mdx so we could simply put all those tests in a markdown file and that would do the trick.

filipeom:
I did some investigation, and it appears that by default, Z3 has the `random_seed` parameter disabled. However, non-deterministic behavior might be observed due to the information provided in this answer: https://stackoverflow.com/a/15731771?

According to Leonardo's answer, _Z3 assigns internal unique IDs to expressions. The internal IDs are used to break ties in some heuristics used by Z3._ This implies that the multi-threaded version might be generating expressions in different orders, thereby altering the IDs assigned to expressions between testing runs.

This likely narrows down the options to the remaining two? Unless we execute the tests with the non-multi-threaded version? 

I can help in the effort of rewriting the tests if we chose to do so.

zapashcanon:
Interesting, thanks! Well, I believe we should still use the multi-core version for testing so indeed only the two other options remain, unless we use a different solver.

Out of curiosity, @hra687261 do you know how easy it is to run alt-ergo or colibri2 in a fully deterministic way regarding model generation ?

hra687261:
Afaik, the usage of expression ids in the solver to order expressions in some algorithms is a common practice, but for the same problem the behavior should be the same?
In the Alt-Ergo case, it is a bit particular, as it uses a weak hash table to store expressions, so if your problem has some expressions that disappear and reappear, depending on whether or not they were garbage-collected, the behavior won't be the same, it's something that the Alt-Ergo team wanted to fix for while but it is yet fixed since it rarely causes problems.
[...]

PR #148, opened 2024-02-06
Make tests a bit more deterministic

krtab:
Fix: #124

### DIFF
--- a/example/c/README.md
+++ b/example/c/README.md
@@ -64,6 +64,9 @@ int main() {
   owi_assume(x != 2);
   owi_assume(x != 4);
 
+  // Make model output deterministic
+  owi_assume(x > -2147483646);
+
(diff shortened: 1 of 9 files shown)


krtab:
> did they change for you ?

Yes, they failed my first `dune runtest` :sweat_smile:

zapashcanon:
Do you think you could extend this with an option to display the model but without the value ? That is only the names of the symbols and their types ?

krtab:
New version requires https://github.com/formalsec/encoding/pull/72 to be merged

zapashcanon:
@krtab could you promote the tests ? The CI is failing because of this.

krtab:
I was sure I did...

zapashcanon:
Thanks!

PR #149, opened 2024-02-06
Generate models for every symbol created through the symbolic API

### LINKED ISSUE #143: include only the symbols created through our symbolic API in the model
see #140

filipeom:
Closes #143

Currently, the models displayed by Owi include only the symbols used in the path condition. However, this may result in the display of auxiliary symbols introduced in symbolic execution, as well as the omission of symbols created by the program but not used for symbolic reasoning. Both issues are immediately resolved with this PR.

Nevertheless, the model display might be unsatisfactory, as the model at the end of the program will now include a lot of assignments for values that we don't really care about (see the C maze example). However, when producing the testsuite, these 'useless' assignments still need to be included in the test cases. Maybe we can strike a balance and only output symbols that are relevant to the stdout, while outputting everything in the testcase? What do you think, @zapashcanon?

This solution might also be problematic as we now store a list of symbols in our thread to maintain the symbol type and name. Are there any suggestions on how to improve this?


zapashcanon:
> However, this may result in the display of auxiliary symbols introduced in symbolic execution

The goal was to prevent the `choice_i32` variables from being printed. Here it looks like they are now called `symbol_` like the others but are still printed. So I don't understand how this is fixed ? (I'm looking at the promoted diff)

> as well as the omission of symbols created by the program but not used for symbolic reasoning

Do we really need to print them in the model at the end ? As not being used for symbolic reasoning mean they're not relevant. Although I can see how this could be better to output them so people are not surprised to have one missing symbol in the output

filipeom:
> The goal was to prevent the `choice_i32` variables from being printed. Here it looks like they are now called `symbol_` like the others but are still printed. So I don't understand how this is fixed ? (I'm looking at the promoted diff)

`choice_i32` are correctly no longer printed. The reason why you're seeing more `symbol_` variables is because the program creates `ITER=28` symbolic variables:

```c
  for (int i = 0; i < ITERS; i++) {
    program[i] = owi_i32();
  }
```

So if we generate a model for all symbols created through the symbolic API, all 28 symbols will appear in the model.

> Do we really need to print them in the model at the end ? As not being used for symbolic reasoning mean they're not relevant. Although I can see how this could be better to output them so people are not surprised to have one missing symbol in the output

I agree that printing all 28 symbols may not be necessary if the program only manipulates 10, which could be more user-relevant. However, for the purpose of replaying the execution in a concrete mode, as required for Test-Comp, the final testcase generated through our `testcase-*.xml` file must include all 28 symbols; otherwise, it won't be considered a valid testcase.

I suggest that we limit the printing of relevant symbols in the stdout (i.e., only variables created through the symbolic API and present in the path condition). Still, we should print all symbols in the generated `testcase-*.xml` to ensure the creation of valid testsuites. What are your thoughts on this approach?

zapashcanon:
> choice_i32 are correctly no longer printed. The reason why you're seeing more symbol_ variables is because the program creates ITER=28 symbolic variables:

Ooooh, makes sense. Sorry haha.

Yes, I'm fine with printing all symbols that are created. Otherwise it's not easy for someone to guess they can have any random value.

Thanks!

PR #150, opened 2024-02-06
Adds collections-c symbolic tests

filipeom:
This adds approximately 30 seconds to the total testsuite execution time.

This was my main testsuite when I was developing WASP. It isn't very exhaustive of all symbolic execution primitives, but it's a nice testsuite to check whether the tool can detect a few simple memory bugs and provide early signs of performance issues. Before the hashconsing, this testsuite ran in 10 to 12 seconds; currently, it takes 30 seconds on my computer.

Edit: this should only be merged after #148 as the `bugs` tests are quite random


zapashcanon:
#148 is merged now, could you rebase on it ? after that let's merge :)

zapashcanon:
Thanks!

PR #151, opened 2024-02-07
Add more tests

epatrizio:
Depends on https://github.com/OCamlPro/owi/pull/148
Currently `grep -v "..."` is used in cram tests to compensate for non determinism.
`--no-model` should be used when #148 is merged.


zapashcanon:
#148 is merged now, can you rebase the PR to use it ? :)

PR #152, opened 2024-02-08
Reset solver before model check

filipeom:
This should help make tests more deterministic. Given some arbitrary sequence of paths, if we have only one failing condition, this model is now always generated with the same solver (the empty solver). Therefore, it should always produce the same result. However, for multiple failing conditions in the same program, we might depend on issue #148 for determinism.


filipeom:
Closing as it doesn't actually do anything :sweat_smile:

PR #154, opened 2024-02-08
Add symbolic interpreter for fuzzing

epatrizio:
We have decided with @chambart to fuzz a simpler version of the symbolic interpreter where every branch can only have one possible choice, which is implemented in the Minimalist monad.

### DIFF
--- a/src/symbolic_choice.ml
+++ b/src/symbolic_choice.ml
@@ -144,6 +144,53 @@ end = struct
 end
 [@@inline]
 
+module Minimalist = struct
+  type err =
+    | Assert_fail
+    | Trap of Trap.t
+
+  type 'a t = M of (Thread.t -> ('a, err) Stdlib.Result.t * Thread.t)
+  [@@unboxed]
+
+  let return v = M (fun t -> (Ok v, t))
+
+  let run_minimalist (M v) st = v st
+
+  let bind v f =
+    M
+      (fun init_s ->
+        let v_final, tmp_st = run_minimalist v init_s in
+        match v_final with
+        | Ok v_final -> run_minimalist (f v_final) tmp_st
+        | Error _ as e -> (e, tmp_st) )
+
+  let ( let* ) = bind
+
+  let select (vb : S.vbool) =
+    let v = Expr.simplify vb in
+    match v.node.e with
+    | Val True -> return true
+    | Val False -> return false
+    | _ -> Format.kasprintf failwith "%a" Expr.pp v
+
+  let select_i32 (i : S.int32) =
+    let v = Expr.simplify i in
+    match v.node.e with Val (Num (I32 i)) -> return i | _ -> assert false
+
+  let trap t = M (fun th -> (Error (Trap t), th))
+
+  let assertion (vb : S.vbool) =
+    let v = Expr.simplify vb in
+    match v.node.e with
+    | Val True -> return ()
+    | Val False -> M (fun th -> (Error Assert_fail, th))
+    | _ -> assert false
+
+  let with_thread f = M (fun st -> (Ok (f st), st))
+
+  let add_pc (_vb : S.vbool) = return ()
+end
+
 module CList = Make (struct
   include List
 
--- a/src/symbolic_choice.mli
+++ b/src/symbolic_choice.mli
@@ -1,5 +1,18 @@
 exception Assertion of Encoding.Expr.t * Thread.t
 
+module Minimalist : sig
+  include
+    Choice_intf.Complete_without_run
+      with type thread := Thread.t
+       and module V := Symbolic_value.S
+
+  type err = private
+    | Assert_fail
+    | Trap of Trap.t
+
+  val run_minimalist : 'a t -> Thread.t -> ('a, err) Stdlib.Result.t * Thread.t
+end
+
 module MT :
   Choice_intf.Complete_with_trap
     with type thread := Thread.t
--- a/test/fuzz/interprets.ml
+++ b/test/fuzz/interprets.ml
@@ -1,4 +1,5 @@
 open Owi
+open Syntax
 
 exception Timeout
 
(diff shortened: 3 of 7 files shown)


src/symbolic_choice.ml
```diff
@@ -144,6 +144,55 @@ end = struct
 end
 [@@inline]
 
+module Minimalist = struct
+  type 'a res =
+    | Ok of 'a
+    | Trap of Trap.t
+    | Assert_fail
```
zapashcanon, line 151:
```ocaml
type err =
| Assert_fail
| Trap of Trap.t

type 'a res = ('a, err) Stdlib.Result.t
```

src/symbolic_choice.ml
```diff
@@ -144,6 +144,55 @@ end = struct
...
+
+  let ( let* ) = bind
+
+  let select (vb : S.vbool) =
+    let v = Expr.simplify vb in
+    match v.node.e with
+    | Val True -> return true
+    | Val False -> return false
+    | _ ->
+      failwith @@ Stdlib.Format.asprintf "%a" Expr.pp v (* bof ... garder ? *)
```
zapashcanon, line 176: `Format.kasprintf failwith "%a" Expr.pp v`

src/symbolic_choice.mli
```diff
@@ -1,5 +1,19 @@
...
+module Minimalist : sig
+  include
+    Choice_intf.Complete_without_run
+      with type thread := Thread.t
+       and module V := Symbolic_value.S
+
+  type 'a res =
+    | Ok of 'a
+    | Trap of Trap.t
+    | Assert_fail
```
zapashcanon, line 12: Same as in the `.ml` also this is not used to build values outside of the model so this can me marked as private ?

test/fuzz/interprets.ml
```diff
@@ -73,6 +73,36 @@ module Owi_optimized : INTERPRET = struct
...
+          timeout_call_run (fun () ->
+              let c = Interpret.SymbolicM.modul link_state.envs regular in
+              let init_thread : Thread.t = Thread.create () in
+              let res, _ =
+                Symbolic_choice.Minimalist.run_minimalist c init_thread
+              in
+              match res with
+              | Ok res -> res
+              | Trap _t -> Result.error "symbolic trap"
+              | Assert_fail -> Result.error "symbolic assert_fail" ) ) )
```
zapashcanon, line 101:
```ocaml
let run modul = 
  open Syntax in
  let* simplified = ... in
  let* () = ... in
  let* (regular, link_state) = ... in
  ...
  
let run modul =
  match run modul with
  | Error e -> failwith e
  | Ok v -> v
```

zapashcanon: (and it looks like the second function is not even necessary)

epatrizio: sorry, I don't understand :(

zapashcanon: If you open the `Owi.Syntax` module, you can rewrite all your code with a `let*` instead of having a lot of (hard to read) nested `match`.

src/symbolic_choice.ml
```diff
@@ -144,6 +144,55 @@ end = struct
 end
 [@@inline]
 
+module Minimalist = struct
+  type err =
+    | Assert_fail
+    | Trap of Trap.t
+
+  type 'a res = ('a, err) Stdlib.Result.t
```
zapashcanon, line 152: And now I believe this can be removed.

src/symbolic_choice.ml
```diff
@@ -144,6 +144,55 @@ end = struct
...
+  let run_minimalist (M v) st = v st
+
+  let bind v f =
+    M
+      (fun init_s ->
+        let v_final, tmp_st = run_minimalist v init_s in
+        match v_final with
+        | Ok v_final -> run_minimalist (f v_final) tmp_st
+        | Error (Trap t) -> (Error (Trap t), tmp_st)
+        | Error Assert_fail -> (Error Assert_fail, tmp_st) )
```
zapashcanon, line 167: Please, try to remember using the `as` pattern to avoid extra-allocations :) :

```ocaml
match v_final with
| Ok v_final -> run_minimalist (f v_final) tmp_st
| Error (Trap _t) as e -> (e, tmp_st)
| Error Assert_fail as e -> (e, tmp_st) )
```

Moreover, the last two cases can be merged:

```ocaml
match v_final with
| Ok v_final -> run_minimalist (f v_final) tmp_st
| Error _ as e -> (e, tmp_st)
```

src/symbolic_choice.mli
epatrizio: You mean delete `type 'a res` from the `.mli` ?
It's used in
```ocaml
val run_minimalist : 'a t -> Thread.t -> 'a res * Thread.t
```

zapashcanon: And I meant marking the `err` type as `private` in the `.mli`:

```ocaml
type err = private
    | Assert_fail
 ...
```

The private keywords allows other modules to match on a value of type `err` but not to create one.

test/fuzz/interprets.ml
epatrizio: > If you open the `Owi.Syntax` module, you can rewrite all your code with a `let*` instead of having a lot of (hard to read) nested `match`.

And i used the same approach in `Owi_optimized` and `Owi_unoptimized` modules

epatrizio: > (and it looks like the second function is not even necessary)

Is this part of the review ok now?

src/symbolic_choice.mli
epatrizio: > It is simply an alias. You can delete it and replace it by its definition.

I couldn't delete `val run_minimalist` in `.mli` because it's used in `Interprets.Owi_symbolic` module.
Maybe there's a better way to do it ?

zapashcanon: You should delete the `res` type, not the `run_minimalist` function.

test/fuzz/interprets.ml
zapashcanon: Yes. :)

krtab:
Rebasing on main should fix the CI @epatrizio

krtab:
I did the rebase :)

zapashcanon:
Thanks!

PR #158, opened 2024-02-12
Refactorize the solver out of Thread.t

krtab:
@epatrizio : be advised that once merged this will change the Minimalist monad a bit and you'll have to adapt.

### DIFF
--- a/src/choice_intf.ml
+++ b/src/choice_intf.ml
@@ -35,6 +35,10 @@ module type Complete = sig
 
   val with_thread : (thread -> 'b) -> 'b t
 
+  val solver : Solver.solver t
+
+  val thread : thread t
+
   val add_pc : V.vbool -> unit t
 
   val run : workers:int -> 'a t -> thread -> 'a run_result
--- a/src/solver.ml
+++ b/src/solver.ml
@@ -1,0 +1,13 @@
+type 'a solver_module = (module Encoding.Solver_intf.S with type t = 'a)
+
+type solver = S : ('a solver_module * 'a) -> solver [@@unboxed]
+
+module Z3Batch = Encoding.Solver.Batch (Encoding.Z3_mappings)
+
+let solver_mod : Z3Batch.t solver_module = (module Z3Batch)
+
+let fresh_solver () =
+  let module Mapping = Encoding.Z3_mappings.Fresh.Make () in
+  let module Batch = Encoding.Solver.Batch (Mapping) in
+  let solver = Batch.create ~logic:QF_BVFP () in
+  S ((module Batch), solver)
--- a/src/symbolic.ml
+++ b/src/symbolic.ml
@@ -4,22 +4,17 @@
 
 open Types
 open Hc
-module Solver = Thread.Solver
+open Solver
+module Solver = Solver.Z3Batch
 
 module type Thread = sig
   type t
-
-  type 'a solver_module = (module Encoding.Solver_intf.S with type t = 'a)
-
-  type solver = S : 'a solver_module * 'a -> solver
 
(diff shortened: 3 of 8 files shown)


src/symbolic.ml
```diff
@@ -129,28 +124,31 @@ struct
...
-          | _ ->
-            (* TODO: can we remove this check? We should be in a SAT branch *)
-            assert (Solver.check solver @@ Thread.pc thread);
-            let concrete_addr = Solver.get_value solver a in
-            let cond = Relop (Eq, a, concrete_addr) @: a.node.ty in
-            (* TODO: this should go to the pc *)
-            Solver.add solver [ cond ];
-            concrete_addr )
+      let* thread = get_thread in
+      let* (S (solver_mod, solver)) = get_solver in
```
zapashcanon, line 130: `let+` (and you can remove the `return` below)

src/choice_intf.ml
```diff
@@ -35,6 +35,10 @@ module type Complete = sig
 
   val with_thread : (thread -> 'b) -> 'b t
 
+  val get_solver : Solver_intf.solver t
+
+  val get_thread : thread t
```
zapashcanon, line 40: `thread` and `solver` would be better than `get_thread` and `get_solver`

src/choice_intf.ml
```diff
@@ -35,6 +35,10 @@ module type Complete = sig
 
   val with_thread : (thread -> 'b) -> 'b t
```
zapashcanon, line 36: is this still used after your PR ?

src/solver_intf.ml
```diff
@@ -0,0 +1,13 @@
...
+
+module Solver = Encoding.Solver.Batch (Encoding.Z3_mappings)
+
+let solver_mod : Solver.t solver_module = (module Solver)
+
+let fresh_solver () =
+  let module Mapping = Encoding.Z3_mappings.Fresh.Make () in
+  let module Batch = Encoding.Solver.Batch (Mapping) in
+  let solver = Batch.create ~logic:QF_BVFP () in
+  S ((module Batch), solver)
```
zapashcanon, line 13: I don't understand why this module is named `solver_intf` and has functions like this one. See [the intf trick](https://www.craigfe.io/posts/the-intf-trick/) for an explanation of the `_intf` idiom.

src/symbolic.ml
zapashcanon: Also, if you rename `get_thread` to `thread` you'll be able to simply write `let* thread in`

zapashcanon:
Regarding the CI failure, could you add a `-w 1` to the failing test to make it deterministic ?

src/solver_intf.ml
krtab: I don't know much about OCaml idioms, so please let me know how you think it would be better organized.

src/choice_intf.ml
krtab: Yes, I may try to delete it in a next PR

src/solver_intf.ml
zapashcanon: `Solver` would be a better name

zapashcanon: And the `module Solver = ...` could be `module Z3 = ...` for now ?

src/solver.ml
```diff
@@ -0,0 +1,13 @@
+type 'a solver_module = (module Encoding.Solver_intf.S with type t = 'a)
+
+type solver = S : 'a solver_module * 'a -> solver
```
zapashcanon, line 3: Could you add an unbox annotation ?

chambart (approved):
:wave:

PR #160, opened 2024-02-13
refactor and add tests

zapashcanon:


zapashcanon:
@filipeom, we were not testing `--test-comp`so I added a test. It seems to fail on macOS, do you know if this is expected ?

filipeom:
> @filipeom, we were not testing `--test-comp`so I added a test. It seems to fail on macOS, do you know if this is expected ?

I wouldn't expect it to fail, but I never had anymore run the c the frontend on macos with `--testcomp`. 

I looks like an issue with `pyml`, and I am unable to reproduce it. If you want we can drop the `--testcomp` instrumentation until we implement a better one in ocaml

PR #170, opened 2024-02-15
Multicore monap v2

krtab:
A cleaner and easier to understand version of the multicore monad, that can easily be extended with a notion of priority.
I want to add comments before merging.


krtab:
First commit is in #171

src/symbolic_choice.ml
```diff
@@ -232,83 +125,88 @@ module WQ = struct
...
+    let return x : _ t = fun _arg -> Now x
+
+    let rec bind_aux (x : _ aux) (f : _ -> _ aux) : _ aux =
+      match x with
+      | Now x -> f x
+      | Yield (prio, lx) -> Yield (prio, fun arg -> bind_aux (lx arg) f)
+      | Choice (mx1, mx2) -> Choice (bind_aux mx1 f, bind_aux mx2 f)
+      | Stop -> Stop
+
+    and bind (mx : ('a, 'arg) t) (f : 'a -> ('b, 'arg) t) : _ t =
```
zapashcanon, line 175: `let rec` instead of `and` should be enough ? It makes reasoning easier as it shows they are not mutually recursive.

src/symbolic_choice.ml
```diff
@@ -232,83 +125,88 @@ module WQ = struct
...
+    let yield prio _ = Yield (prio, fun _ -> Now ())
+
+    let choose a b = Choice (a, b)
+
+    let stop _arg = Stop
+
+    let thread_local : ('a, 'a) t = fun arg -> Now arg
+  end
+
+  type 'a out = ('a, solver) Out.t
```
zapashcanon, line 191: Is it possible to remove this type alias ?

src/symbolic_choice.ml
```diff
@@ -232,83 +125,88 @@ module WQ = struct
...
-    | Trap t -> Trap t
-    | Retv v -> f v
-    | Assert _ | Ret _ | Choice _ | Choice_i32 _ | Bind _ -> Bind (v, f)
+   fun mv f ->
+    St
+      (fun st ->
+        Out.bind (run mv st) (fun (v, news_st) ->
+            bind_lift_eval v
+              (fun v -> run (f v) news_st)
+              (fun x -> Out.return (x, news_st)) ) )
```
zapashcanon, line 209: Do we want to provide `let*` and `let+` in `Out` and open it here ?

src/symbolic_choice.ml
```diff
@@ -232,83 +125,88 @@ module WQ = struct
...
+      match mx arg with
+      | Yield (prio, mx) -> Yield (prio, bind mx f)
+      | x -> bind_aux x argumented_f
 
-  let return v = Retv v [@@inline]
+    let yield prio _ = Yield (prio, fun _ -> Now ())
+
+    let choose a b = Choice (a, b)
+
+    let stop _arg = Stop
```
zapashcanon, line 186: `let stop = Fun.const Stop`

src/symbolic_choice.ml
```diff
@@ -319,128 +217,198 @@ module Multicore = struct
...
-    match cond.node.e with
-    | Val True -> Retv true
-    | Val False -> Retv false
-    | _ -> Choice cond
-  [@@inline]
+  let with_thread f = St (fun st _arg -> Now (EVal (f st), st))
 
-  let select_i32 (i : S.int32) =
-    match i.node.e with Val (Num (I32 v)) -> Retv v | _ -> Choice_i32 i
+  let thread = with_thread (fun st -> st)
```
zapashcanon, line 222: `with_thread Fun.id`

src/symbolic_choice.ml
```diff
@@ -319,128 +217,198 @@ module Multicore = struct
...
+      let model = Solver.model ~symbols:[ symbol ] s in
+      match model with
+      | None ->
+        failwith "Unreachable: The problem is sat so a model should exist"
+      | Some model -> begin
+        match Model.evaluate model symbol with
+        | None ->
+          failwith
+            "Unreachable: The model exists so this symbol should evaluate"
+        | Some v -> Some v
```
zapashcanon, line 275: `Some _v as v -> v` to avoid an allocation.

src/symbolic_choice.ml
```diff
@@ -319,128 +217,198 @@ module Multicore = struct
...
+  let select (cond : S.vbool) =
+    let v = Expr.simplify cond in
+    match v.node.e with
+    | Val True -> return true
+    | Val False -> return false
+    | Val (Num (I32 _)) -> failwith "unreachable (type error)"
+    | _ ->
+      let true_branch =
+        let* () = add_pc v in
+        let* () = check_reachability in
```
zapashcanon, line 292: let+

src/symbolic_choice.ml
```diff
@@ -319,128 +217,198 @@ module Multicore = struct
...
+        return true
+      in
+      let false_branch =
+        let* () = add_pc (S.Bool.not v) in
+        let* () = check_reachability in
```
zapashcanon, line 297: let+

src/symbolic_choice.ml
```diff
@@ -319,128 +217,198 @@ module Multicore = struct
...
+  [@@inline]
+
+  let summary_symbol (e : Expr.t) =
+    let* thread in
+    match e.node.e with
+    | Symbol sym -> return (None, sym)
+    | _ ->
+      let choices = thread.choices in
+      let symbol_name = Format.sprintf "choice_i32_%i" choices in
+      let+ () = modify_thread (fun t -> { t with choices = choices + 1 }) in
```
zapashcanon, line 310: `choices = succ choices`

src/symbolic_choice.ml
```diff
@@ -319,128 +217,198 @@ module Multicore = struct
...
+      let* assign, symbol = summary_symbol sym_int in
+      let* () =
+        match assign with Some assign -> add_pc assign | None -> return ()
+      in
+      let rec generator () =
+        let* possible_value = get_model_or_stop symbol in
+        let i =
+          match possible_value with
+          | Num (I32 i) -> i
+          | _ -> failwith "Unreachable: found sumbol must be a value"
```
zapashcanon, line 329: s/sumbol/symbol/

src/symbolic_choice.ml
```diff
@@ -319,128 +217,198 @@ module Multicore = struct
...
+        in
+        let this_value_cond = Expr.Bitv.I32.(Expr.mk_symbol symbol = v i) in
+        let not_this_value_cond =
+          Expr.Bitv.I32.(Expr.mk_symbol symbol != v i)
```
zapashcanon, line 333: `!=` is the negation of `==` (physical equality), you probably want `<>` instead ?

src/symbolic_choice.ml
```diff
@@ -319,128 +217,198 @@ module Multicore = struct
...
 
   let run ~workers t thread =
     let global = init_global () in
     push_first_work global thread t;
-    let producers = Array.init workers (spawn_producer global) in
-    Counter.wait_n global.start_counter workers;
-    loop_and_do (WQ.read_as_seq global.r) (fun () ->
-        Array.iter Domain.join producers )
+    for _ = 0 to workers do
+      let _ = spawn_producer global in
```
zapashcanon, line 410: do not use `let _ =`, if you change the type of `spawn_producer`, you will do a partial application without noticing and this is usually hard to debug, adding a type annotation is enough to fix the potential issue

src/symbolic_choice.ml
krtab: It's actually Expr.Vitv.I32's `!=` : https://github.com/formalsec/encoding/blob/main/lib/expr.ml#L359

src/symbolic_choice.ml
```diff
@@ -162,21 +75,20 @@ module WQ = struct
     { mutex : Mutex.t
     ; cond : Condition.t
     ; queue : 'a Queue.t
-    ; mutable producers : int
+    ; mutable pledges : int
```
zapashcanon, line 78: Maybe we could use:

```ocaml
; pledges : int Atomic.t
```

Then `make_pledge` would simply be `Atomic.incr`.

src/symbolic_choice.ml
```diff
@@ -162,21 +75,20 @@ module WQ = struct
...
     ; mutable failed : bool
```
zapashcanon, line 79: `bool Atomic.t` ?

zapashcanon:
Thanks!

Issue #169, opened 2024-02-14
Unsound All OK


krtab:
May be related to #168 

```wat
(module
  (import "symbolic" "i32_symbol" (func $i32_symbol (result i32)))
  (import "symbolic" "assume" (func $assume (param i32)))
  (import "symbolic" "assert" (func $assert (param i32)))

  (memory $m 1)
  (data (offset (i32.const 0)) "\00\01\02\00")

  (func $start

    (local $n i32)

    (local.set $n (call $i32_symbol))

    (call $assume   ;; 0 <= i32 < 4
      (i32.and
        (i32.ge_u (local.get $n) (i32.const 0))
        (i32.lt_u (local.get $n) (i32.const 4))
        )
    )

    (call $assert
      (i32.eq
        (i32.load8_u (local.get $n))
        (local.get $n)
      )
    )
  )

  (start $start)
)
```

gives

```
All OK
```

should be

```
Assert failure: (i32.eq (i32 0) symbol_0)
Model:
  (model
    (symbol_0 (i32 3)))
Reached problem!
```

krtab:
Same issue when doing the memory initialization by hand

```wat
(module
  (import "symbolic" "i32_symbol" (func $i32_symbol (result i32)))
  (import "symbolic" "assume" (func $assume (param i32)))
  (import "symbolic" "assert" (func $assert (param i32)))

  (memory $m 1)
  ;; (data (offset (i32.const 0)) "\00\01\02\03")  

  (func $start

    (local $n i32)
 
    (i32.store8 (i32.const 0) (i32.const 0))
    (i32.store8 (i32.const 1) (i32.const 1))
    (i32.store8 (i32.const 2) (i32.const 2))
    (i32.store8 (i32.const 3) (i32.const 0))    

    (local.set $n (call $i32_symbol))

    (call $assume   ;; 0 <= i32 < 4
      (i32.and
        (i32.ge_u (local.get $n) (i32.const 0))
        (i32.lt_u (local.get $n) (i32.const 4))
        )
    )

    (call $assert
      (i32.eq
        (i32.load8_u (local.get $n))
        (local.get $n)
      )
    )
  )

  (start $start)
)
```

filipeom:
This looks like an issue from `concretise`: https://github.com/OCamlPro/owi/blob/feafe94d59a108c7a3a1fdfdc44ab255cd333203/src/symbolic.ml#L126-L149

We're not branching when we concretise pointers

PR #172, opened 2024-02-16
Fix Symbolic.concretize

krtab:
Fixes: https://github.com/OCamlPro/owi/issues/168
Fixes: https://github.com/OCamlPro/owi/issues/169

@filipeom This changes the output of some of the C tests so I'm requiring your review as well

### DIFF
--- a/src/symbolic.ml
+++ b/src/symbolic.ml
@@ -59,31 +59,19 @@ struct
   module Memory = struct
     include Symbolic_memory
 
-    let concretise a =
+    let concretise (a : Encoding.Expr.t) : Encoding.Expr.t Choice.t =
       let open Choice in
-      let open Encoding.Expr in
-      let* thread in
-      let+ (S (solver_mod, solver)) = solver in
-      let module Solver = (val solver_mod) in
-      let open Hc in
+      let open Encoding in
       match a.node.e with
-      | Val _ | Ptr (_, { node = { e = Val _; _ }; _ }) -> a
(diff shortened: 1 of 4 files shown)


filipeom:
Wow, this looks great! And much more concise. However, in concretizing, we still need to pattern match on the `Ptr` value. Otherwise, `select_i32` will start discharging concrete queries to the solver like: `fix_symbol: (i32.eq choice_i32_271 (Ptr (i32 8389712) (i32 20)))`. And, we lose the chunks' bounds information in the function `check_within_bounds` just below, making it unable to identify memory bugs (the C tests that changed).

So, making `concretize` something like:

```ocaml
    let concretize (a : Expr.t) : int32 Choice_monad.t =
      let open Choice in
      match a.node.e with
      | Ptr (base, offset) ->
        let+ offset = select_i32 offset in
        Expr.(Ptr (base, Value.const_i32 offset) @: Ty_bitv S32)
      | _ ->
        let+ v = select_i32 a in
        Value.const_i32 v
```

Should be enough to solve both issues :grinning:

filipeom (approved):

zapashcanon:
@krtab, I rebased your PR as I merged a big refactoring.

Otherwise, I'm fine with merging with Filipe's review being applied if it allows to recover the finding of the bugs in the tests.

krtab:
Should be good to do.
I had to deactivate the store.wat test that will just enumerate all possible i32.

zapashcanon:
Thanks!

PR #174, opened 2024-02-17
refactor symbolic and interpret interface

zapashcanon:


zapashcanon:
The CI failure is probably a bug in odoc, see https://github.com/ocaml/odoc/issues/1083. Funnily, it does not happen on macOS CI...

Issue #128, opened 2024-01-25
symbolic div_s should raise Division_by_zero


zapashcanon:
```wat
(module
  (import "symbolic" "i32_symbol" (func $i32_symbol (result i32)))
  (func $start
    i32.const 1
    call $i32_symbol
    i32.div_s
    drop)

  (start $start)
)
```

PR #181, opened 2024-02-21
Adapt check that divisor is not zero for the symbolic case

krtab:
Fixes: #128

### DIFF
--- a/src/interpret.ml
+++ b/src/interpret.ml
@@ -45,6 +45,10 @@ module Make (P : Interpret_intf.P) :
 
     let ( <> ) = ne
 
+    let ( = ) = eq
+
+    let eqz v = v = zero
+
     let min_int = const_i32 Int32.min_int
   end
 
@@ -70,6 +74,10 @@ module Make (P : Interpret_intf.P) :
     let ( ~- ) x = const_i64 0L - x
 
     let ( <> ) = ne
+
+    let ( = ) = eq
+
+    let eqz v = v = zero
 
     let min_int = const_i64 Int64.min_int
   end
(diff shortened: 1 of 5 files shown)


krtab:
3376219 is a rider to appease the CI

src/interpret.ml
```diff
@@ -87,6 +95,10 @@ module Make (P : Interpret_intf.P) :
     let* v = select v in
     f v
 
+  let ifthenelse cond t e =
+    let* cond = select cond in
+    if cond then t () else e ()
```
zapashcanon, line 100: I would prefer not to add this, could you inline it directly ?

src/interpret.ml
```diff
@@ -150,21 +162,21 @@ module Make (P : Interpret_intf.P) :
...
         | Mul -> Choice.return @@ mul n1 n2
         | Div s -> begin
-          try
-            match s with
-            | S ->
-              let> overflow = Bool.and_ (eq n1 min_int) @@ eq n2 ~-(const 1l) in
-              if overflow then Choice.trap Integer_overflow
-              else Choice.return @@ div n1 n2
-            | U -> Choice.return @@ unsigned_div n1 n2
-          with Division_by_zero -> Choice.trap Integer_divide_by_zero
```
zapashcanon, line 160: I'm surprised that we don't need to catch the exception anymore in the concrete case

src/interpret.ml
zapashcanon: You could use `let>` instead, which is already defined and used in this file.

src/interpret.ml
krtab: Select works in both cases anyway so there is now an explicit guard marking the exception raising path unreachable

zapashcanon:
Thanks!

PR #189, opened 2024-02-23
Symbolic fuzzing

epatrizio:
Symbolic execution fuzzing first step : concrete context
Wasm modules have only concrete values and fuzzing checks that symbolic execution has the same behavior as concrete execution


test/fuzz/basic.ml
```diff
@@ -87,10 +87,10 @@ let ibinop =
     ; const (Xor : ibinop)
     ; const (Shl : ibinop)
     ; shr
-    ; const (Rotl : ibinop)
-    ; const (Rotr : ibinop)
     ]
 
+let ibinop_rot = choose [ const (Rotl : ibinop); const (Rotr : ibinop) ]
```
zapashcanon, line 92: Why are we separating these now ?

test/fuzz/basic.ml
```diff
@@ -169,14 +177,9 @@ let extend_64_i64 : text instr gen =
...
-    ; const Neg
-    ; const Sqrt
-    ; const Ceil
-    ; const Floor
-    ; const Trunc
-    ; const Nearest
-    ]
+    [ const Abs; const Neg; const Sqrt; const Ceil; const Floor; const Nearest ]
+
+let funop_trunc = const Trunc
```
zapashcanon, line 182: Same question here

test/fuzz/cond.ml
```diff
@@ -0,0 +1,66 @@
...
+let bind (cond : _ cond) (fcond : _ -> _ cond) : _ cond =
+ fun c -> Option.bind (cond c) (fun ag -> fcond ag c)
+
+let ( let* ) = bind
+
+let run (cond : _ cond) conf : 'a option = cond conf
+
+let lift1 f ma =
+  let* a = ma in
+  return @@ f a
```
zapashcanon, line 20:
```ocaml
let lift1 f ma =
  let* ma in
  return @@ f ma
```

zapashcanon:
IIUC, all the complexity with the new weird operators is here to allow ignoring some instructions in the symbolic case ? I guess it is because they have not been implemented yet ? 

If that is the case, I'd prefer them to be actually implemented (even in an inefficient way) rather than adding all this complex machinery..

test/fuzz/gen.ml
```diff
@@ -214,58 +263,88 @@ let rec expr ~block_type ~stack ~locals env =
...
+    let* expr_available_2_i64_i32 = expr_available_2_i64_i32 env in
+    let* expr_available_2_f32_i32 = expr_available_2_f32_i32 env in
+    let* expr_available_2_f64_i32 = expr_available_2_f64_i32 env in
+    let* expr_available_3_i32 = expr_available_3_i32 env in
+    let* expr_available_1_i64 = expr_available_1_i64 env in
+    let* expr_available_2_i64 in
+    let* expr_available_1_f32 = expr_available_1_f32 env in
+    let* expr_available_2_f32 in
+    let* expr_available_1_f64 = expr_available_1_f64 env in
+    let* expr_available_2_f64 in
```
zapashcanon, line 277: Why are you using `let* e in` for some and `let* e = e in` for others ? Could you switch all of them with `let* e in` ?

test/fuzz/gen.ml
```diff
@@ -67,137 +76,177 @@ let expr_available_2_i64_i32 (env : Env.t) =
...
-      [ B.memory_copy; B.memory_fill ] @ B.memory_init env
+      [ return @@ B.memory_copy; not_symbolic @@ B.memory_fill ]
+      @ List.map (fun e -> return @@ e) (B.memory_init env)
     else []
   in
-  let table_expr = B.table_init env @ B.table_copy env in
-  mem_expr @ table_expr
+  let table_expr =
+    List.map (fun e -> return @@ e) (B.table_init env)
+    @ List.map (fun e -> return @@ e) (B.table_copy env)
```
zapashcanon, line 100: `List.map return (B.table_init env @ B.table_copy env)`

test/fuzz/gen.ml
```diff
@@ -67,137 +76,177 @@ let expr_available_2_i64_i32 (env : Env.t) =
...
   if B.memory_exists env then [ pair B.f64_store (const [ S.Pop; S.Pop ]) ]
   else []
 
-let expr_available_3_i32 env =
+let expr_available_3_i32 env : _ condgen =
   let mem_expr =
     if B.memory_exists env then
-      [ B.memory_copy; B.memory_fill ] @ B.memory_init env
+      [ return @@ B.memory_copy; not_symbolic @@ B.memory_fill ]
+      @ List.map (fun e -> return @@ e) (B.memory_init env)
```
zapashcanon, line 95: `@ List.map return (B.memory_init env)`

(Also, using `return @@ e` instead of `return e` is a little bit weird)

test/fuzz/gen.ml
```diff
@@ -67,137 +76,177 @@ let expr_available_2_i64_i32 (env : Env.t) =
...
-    and+ rt_descr = const @@ List.rev_map (fun t -> S.Push t) rt in
-    (instr, pt_descr @ rt_descr)
-  end
-  | _ -> assert false
+  Cond.with_unwrapping_configurator (fun configurator ->
+      let open Crowbar.Syntax in
+      match stack with
+      | Num_type I32 :: _stack -> begin
+        let* rt = list B.val_type in
+        let* pt = B.stack_prefix (List.tl stack) in
```
zapashcanon, line 179: This is not because of this PR but the code would be better written like this:

```ocaml
      match stack with
      | Num_type I32 :: stack -> begin
        let* rt = list B.val_type in
        let* pt = B.stack_prefix stack in
```

test/fuzz/gen.ml
```diff
@@ -298,33 +381,46 @@ let global env =
...
+  let open Crowbar.Syntax in
   let* () = const () in
   Env.reset_locals env;
   Env.refill_fuel env;
   let* locals = list (local env) in
   let* type_f = B.block_type env in
   let (_name : string) = Env.add_block env type_f Env.Func in
-  let+ body = expr ~block_type:type_f ~stack:[] ~locals env in
+  let+ body =
+    Option.get @@ run (expr ~block_type:type_f ~stack:[] ~locals env) fuzz_conf
```
zapashcanon, line 393: Could you explicit a little bit why `Option.get` is fine here ? By simply matching on it + a proper error message for instance.

test/fuzz/gen.ml
```diff
@@ -298,33 +381,46 @@ let global env =
...
   let+ start_code =
     let* () = const () in
     Env.reset_locals env;
     Env.refill_fuel env;
     let type_f = Bt_raw (None, ([], [])) in
     let id = Some "start" in
-    let+ body = expr ~block_type:type_f ~stack:[] ~locals:[] env in
+    let+ body =
+      Option.get
+      @@ run (expr ~block_type:type_f ~stack:[] ~locals:[] env) fuzz_conf
```
zapashcanon, line 422: Same here.

test/fuzz/gen.ml
```diff
@@ -298,33 +381,46 @@ let global env =
...
+  let* memory =
+    match fuzz_conf with
+    | Concrete -> option (memory env)
+    | Symbolic -> const None
+  in
   let* datas = list (data env) in
   let* types = list (typ env) in
-  let* tables = list (table env) in
+  let* tables =
+    match fuzz_conf with Concrete -> list (table env) | Symbolic -> const []
```
zapashcanon, line 409: Why are we not creating tables in the symbolic case ? Too much stuff is unimplemented ?

test/fuzz/gen.ml
```diff
@@ -298,33 +381,46 @@ let global env =
...
   MFunc { type_f; locals; body; id }
 
-let fields env =
-  let* memory = option (memory env) in
+let fields env fuzz_conf =
+  let open Crowbar.Syntax in
+  let* memory =
+    match fuzz_conf with
+    | Concrete -> option (memory env)
+    | Symbolic -> const None
```
zapashcanon, line 404: I guess this is to avoid getting errors due to the unsupported memory model ? A comment with a `(* TODO *)` saying it should be enabled at some point would be great.

test/fuzz/basic.ml
epatrizio: Same reason as above.

test/fuzz/gen.ml
zapashcanon: Oh indeed, sorry

test/fuzz/basic.ml
zapashcanon: OK, see my comment on the PR below then.

test/fuzz/gen.ml
epatrizio: An oversight during refactoring

test/fuzz/gen.ml
epatrizio: An oversight during refactoring

test/fuzz/basic.ml
filipeom: This is my fault, it should be easy to implement these as it's only the concrete evaluation case that is missing https://github.com/formalsec/encoding/issues/79

zapashcanon: Oh they exist in encoding ? I though it was like the `copysign` one, which was easy to implement symbolically directly in owi (it is exactely the same code as the concrete version but with symbolic primitives)

filipeom: This is the reverse case, they are mapped to `z3` but not resolved concretely in `unop` in `symbolic_value.ml` which calls encoding's `eval_numeric.ml` but I will push a fix in a bit for `Rot(l|r)` and `Trunc`.

zapashcanon: For the concrete case, you could just copy what's in `owi/src/int32.ml` right ?

filipeom: > For the concrete case, you could just copy what's in owi/src/int32.ml right ?
 
Yes I will just plagiarize it :sweat_smile:

test/fuzz/gen.ml
zapashcanon: I created #190.

zapashcanon:
OK so, before merging, I'd like two things to be fixed:

- let's implement all basics instructions (either waiting for them to be in encoding or adding them in Owi when possible) like `rotl`, `rotr` etc.
- once this is done, I believe most of the complexity added in the PR from `cond.ml` can be removed (even without supporting `tables` in symbolic mode);
- let's try to see if completing the table support is hard or not (#190), I expect it to be quite straightforward, but if it's not, I'm fine with merging without it.

filipeom:
> let's implement all basics instructions (either waiting for them to be in encoding or adding them in Owi when possible) like rotl, rotr etc.

I fixed `rotl`, `rotr`, and `trunc` here: https://github.com/formalsec/encoding/pull/80. 

`ctz` and `popcnt` are still missing in owi and in the encoding, so I suggest that for the time being we concretely evaluate them in `symbolic_value.ml`.

krtab:
> `ctz` and `popcnt` are still missing in owi and in the encoding, so I suggest that for the time being we just concretely evaluate them in `symbolic_value.ml`.

There may even be clever tricks where we make a decision tree based on `Choice.choose` to compute those.

filipeom:
> > `ctz` and `popcnt` are still missing in owi and in the encoding, so I suggest that for the time being we just concretely evaluate them in `symbolic_value.ml`.
> 
> There may even be clever tricks where we make a decision tree based on `Choice.choose` to compute those.

Yes, this can be an efficient way to solve this. I'm worried it could cause too much branching? I can think of one unclever way to implement it without branching; we'd create a massive if-then-else expression:

```smt2
 (ite (n = 0) 
   32
   (ite (not (n & 1 = 0))
     31
     (ite (not (n & 2 = 0)) 
       30
       ... 
         0))))
```

But this would probably grind z3 to a halt :sweat_smile:

test/fuzz/basic.ml
```diff
@@ -130,6 +130,14 @@ let ibinop_64 : text instr gen =
   let+ ibinop in
   I_binop (S64, ibinop)
 
+let ibinop_rot_32 : text instr gen =
+  let+ ibinop_rot in
+  I_binop (S32, ibinop_rot)
+
+let ibinop_rot_64 : text instr gen =
+  let+ ibinop_rot in
+  I_binop (S64, ibinop_rot)
```
zapashcanon, line 139: They can be deleted

test/fuzz/basic.ml
zapashcanon, line 182: Same with funop

test/fuzz/basic.ml
```diff
@@ -208,6 +211,14 @@ let funop_64 : text instr gen =
...
   F_unop (S64, funop)
 
+let funop_trunc_32 : text instr gen =
+  let+ funop_trunc in
+  F_unop (S32, funop_trunc)
+
+let funop_trunc_64 : text instr gen =
+  let+ funop_trunc in
+  F_unop (S64, funop_trunc)
+
```
zapashcanon, line 221: Same

test/fuzz/gen.ml
```diff
@@ -35,18 +35,19 @@ let expr_available_1_i32 if_else expr ~locals ~stack env =
...
   ; pair B.f32_convert_i32 (const [ S.Pop; S.Push (Num_type F32) ])
   ; pair B.f64_convert_i32 (const [ S.Pop; S.Push (Num_type F64) ])
   ; pair B.f32_reinterpret_i32 (const [ S.Pop; S.Push (Num_type F32) ])
-  ; if_else expr ~locals ~stack env
   ]
   @ B.local_set_i32 env @ B.local_tee_i32 env @ B.global_set_i32 env
-  @ (if B.memory_exists env then B.memory_grow :: load_instr else [])
-  @ B.expr_br_if env stack @ B.table_get env
+  @
+  if B.memory_exists env then [ B.memory_grow ] @ load_instr
```
zapashcanon, line 49: `B.memory_grow :: load_instr` was more readable

test/fuzz/gen.ml
```diff
@@ -55,7 +56,10 @@ let expr_available_2_i32 (env : Env.t) =
     ; pair B.i32_store16 (const [ S.Pop; S.Pop ])
     ]
   in
-  [ pair B.ibinop_32 (const [ S.Pop ]); pair B.irelop_32 (const [ S.Pop ]) ]
+  [ pair B.ibinop_rot_32 (const [ S.Pop ])
```
zapashcanon, line 59: This one can be removed

test/fuzz/gen.ml
```diff
@@ -86,30 +90,32 @@ let expr_available_3_i32 env =
...
-  ; pair B.extend_64_i64 (const [ S.Nothing ])
   ; pair B.f32_convert_i64 (const [ S.Pop; S.Push (Num_type F32) ])
   ; pair B.f64_convert_i64 (const [ S.Pop; S.Push (Num_type F64) ])
   ; pair B.f64_reinterpret_i64 (const [ S.Pop; S.Push (Num_type F64) ])
   ]
   @ B.local_set_i64 env @ B.local_tee_i64 env @ B.global_set_i64 env
 
 let expr_available_2_i64 =
-  [ pair B.ibinop_64 (const [ S.Pop ])
+  [ pair B.ibinop_rot_64 (const [ S.Pop ])
```
zapashcanon, line 103: same

test/fuzz/gen.ml
```diff
@@ -86,30 +90,32 @@ let expr_available_3_i32 env =
...
+  ; pair B.ibinop_64 (const [ S.Pop ])
   ; pair B.irelop_64 (const [ S.Pop; S.Pop; S.Push (Num_type I32) ])
   ]
 
 (* let expr_available_3_i64 = [] *)
 
 let expr_available_1_f32 env =
-  [ pair B.funop_32 (const [ S.Nothing ])
-  ; pair B.i32_trunc_f32 (const [ S.Pop; S.Push (Num_type I32) ])
+  [ pair B.i32_trunc_f32 (const [ S.Pop; S.Push (Num_type I32) ])
```
zapashcanon, line 111: same

test/fuzz/gen.ml
```diff
@@ -86,30 +90,32 @@ let expr_available_3_i32 env =
...
   ; pair B.i64_trunc_f32 (const [ S.Pop; S.Push (Num_type I64) ])
   ; pair B.i32_trunc_sat_f32 (const [ S.Pop; S.Push (Num_type I32) ])
   ; pair B.i64_trunc_sat_f32 (const [ S.Pop; S.Push (Num_type I64) ])
   ; pair B.f64_promote_f32 (const [ S.Pop; S.Push (Num_type F64) ])
   ; pair B.i32_reinterpret_f32 (const [ S.Pop; S.Push (Num_type I32) ])
+  ; pair B.funop_trunc_32 (const [ S.Nothing ])
```
zapashcanon, line 117: same

test/fuzz/gen.ml
```diff
@@ -121,28 +127,31 @@ let expr_available_2_f32 =
 (* let expr_available_3_f32 = [] *)
 
 let expr_available_1_f64 env =
-  [ pair B.funop_64 (const [ S.Nothing ])
-  ; pair B.i32_trunc_f64 (const [ S.Pop; S.Push (Num_type I32) ])
+  [ pair B.i32_trunc_f64 (const [ S.Pop; S.Push (Num_type I32) ])
```
zapashcanon, line 130: same

test/fuzz/gen.ml
```diff
@@ -121,28 +127,31 @@ let expr_available_2_f32 =
...
   ; pair B.i64_trunc_f64 (const [ S.Pop; S.Push (Num_type I64) ])
   ; pair B.i32_trunc_sat_f64 (const [ S.Pop; S.Push (Num_type I32) ])
   ; pair B.i64_trunc_sat_f64 (const [ S.Pop; S.Push (Num_type I64) ])
   ; pair B.f32_demote_f64 (const [ S.Pop; S.Push (Num_type F32) ])
   ; pair B.i64_reinterpret_f64 (const [ S.Pop; S.Push (Num_type I64) ])
+  ; pair B.funop_trunc_64 (const [ S.Nothing ])
```
zapashcanon, line 136: same

test/fuzz/gen.ml
```diff
@@ -121,28 +127,31 @@ let expr_available_2_f32 =
...
+  ; pair B.funop_64 (const [ S.Nothing ])
   ]
   @ B.local_set_f64 env @ B.local_tee_f64 env @ B.global_set_f64 env
 
 let expr_available_2_f64 =
   [ pair B.fbinop_64 (const [ S.Pop ])
+  ; pair B.fbinop_64 (const [ S.Pop ])
```
zapashcanon, line 143: this is duplicated and should be removed

test/fuzz/gen.ml
```diff
@@ -121,28 +127,31 @@ let expr_available_2_f32 =
...
   ; pair B.frelop_64 (const [ S.Pop; S.Pop; S.Push (Num_type I32) ])
   ]
 
 (* let expr_available_3_f64 = [] *)
 
 let if_else expr ~locals ~stack env =
+  let open Crowbar.Syntax in
```
zapashcanon, line 150: We should probably bring back the top-level `open Crowbar.Syntax` now

test/fuzz/gen.ml
```diff
@@ -214,58 +226,77 @@ let rec expr ~block_type ~stack ~locals env =
...
+    let expr_available_2_i64_i32 = expr_available_2_i64_i32 env in
+    let expr_available_2_f32_i32 = expr_available_2_f32_i32 env in
+    let expr_available_2_f64_i32 = expr_available_2_f64_i32 env in
+    let expr_available_3_i32 = expr_available_3_i32 env in
+    let expr_available_1_i64 = expr_available_1_i64 env in
+    let expr_available_1_f32 = expr_available_1_f32 env in
+    let expr_available_1_f64 = expr_available_1_f64 env in
+    let expr_always_available =
+      expr_always_available block loop expr ~locals ~stack env
+    in
```
zapashcanon, line 239: I would simply remove all of this. (Except maybe the `expr_always_available` one)

test/fuzz/gen.ml
```diff
@@ -214,58 +226,77 @@ let rec expr ~block_type ~stack ~locals env =
...
+        let expr_available_1_i32 =
+          expr_available_1_i32 if_else expr ~stack ~locals env
+        in
+        expr_available_1_any @ expr_available_1_i32 @ expr_available_2_i32
+        @ expr_available_3_i32
       | Num_type I32 :: Num_type I32 :: _tl ->
-        expr_available_1_any
-        @ expr_available_1_i32 if_else expr ~stack ~locals env
-        @ expr_available_2_i32 env
+        let expr_available_1_i32 =
```
zapashcanon, line 250: same

test/fuzz/gen.ml
```diff
@@ -214,58 +226,77 @@ let rec expr ~block_type ~stack ~locals env =
...
+      | Num_type F32 :: Num_type I32 :: _tl -> expr_available_2_f32_i32
+      | Num_type F64 :: Num_type I32 :: _tl -> expr_available_2_f64_i32
       | Num_type I64 :: Num_type I64 :: _tl ->
-        expr_available_1_any @ expr_available_1_i64 env @ expr_available_2_i64
+        expr_available_1_any @ expr_available_1_i64 @ expr_available_2_i64
       | Num_type I32 :: _tl ->
-        expr_available_1_any
-        @ expr_available_1_i32 if_else expr ~stack ~locals env
-      | Num_type I64 :: _tl -> expr_available_1_any @ expr_available_1_i64 env
+        let expr_available_1_i32 =
```
zapashcanon, line 264: same

test/fuzz/gen.ml
epatrizio: I also removed `expr_always_available` ;)

krtab:
Before merging, could you rebase it as one single commit ?

test/fuzz/basic.ml
```diff
@@ -91,7 +91,11 @@ let ibinop =
     ; const (Rotr : ibinop)
     ]
 
-let iunop = choose [ const Clz; const Ctz; const Popcnt ]
+(* TODO :
+   temp comment for symbolic context "const Popcnt"
+   iunop_popcnt_32 / iunop_popcnt_64 in concrete contexte
+*)
+let iunop = choose [ const Clz; const Ctz (*; const Popcnt*) ]
```
zapashcanon, line 98: Could you put `popcnt` back as it was before ?

test/fuzz/basic.ml
```diff
@@ -138,6 +142,16 @@ let iunop_64 : text instr gen =
...
+(* TODO: check comment above *)
+let iunop_popcnt_32 : text instr gen =
+  let+ popcnt = const Popcnt in
+  I_unop (S32, popcnt)
+
+(* TODO: check comment above *)
+let iunop_popcnt_64 : text instr gen =
+  let+ popcnt = const Popcnt in
+  I_unop (S64, popcnt)
+
```
zapashcanon, line 154: These two functions could be removed I guess.

test/fuzz/gen.ml
```diff
@@ -36,17 +37,22 @@ let expr_available_1_i32 if_else expr ~locals ~stack env =
...
+        ; pair B.i64_extend_i32 (const [ S.Pop; S.Push (Num_type I64) ])
+        ; pair B.iunop_popcnt_32 (const [ S.Nothing ])
+        ]
+      else [] )
   @ B.local_set_i32 env @ B.local_tee_i32 env @ B.global_set_i32 env
-  @ (if B.memory_exists env then B.memory_grow :: load_instr else [])
-  @ B.expr_br_if env stack @ B.table_get env
+  @
+  if B.memory_exists env then B.memory_grow :: load_instr
+  else [] @ B.expr_br_if env stack @ B.table_get env
```
zapashcanon, line 55: I believe this is wrong (or confusing) due to wrong parenthesis, it should be:

```ocaml
( if B.memory_exists env then B.memory_grow :: load_instr
  else [] )
@ B.expr_br_if env stack @ B.table_get env
```

test/fuzz/gen.ml
```diff
@@ -103,15 +116,18 @@ let expr_available_2_i64 =
...
   ]
-  @ B.local_set_f32 env @ B.local_tee_f32 env @ B.global_set_f32 env
+  @
+  if env.Env.conf = Concrete then
+    [ pair B.i32_trunc_f32 (const [ S.Pop; S.Push (Num_type I32) ])
+    ; pair B.i64_trunc_f32 (const [ S.Pop; S.Push (Num_type I64) ])
+    ; pair B.i32_trunc_sat_f32 (const [ S.Pop; S.Push (Num_type I32) ])
+    ; pair B.i64_trunc_sat_f32 (const [ S.Pop; S.Push (Num_type I64) ])
+    ]
+  else [] @ B.local_set_f32 env @ B.local_tee_f32 env @ B.global_set_f32 env
```
zapashcanon, line 130: Same here:

```ocaml
  ( if env.Env.conf = Concrete then
    [ pair B.i32_trunc_f32 (const [ S.Pop; S.Push (Num_type I32) ])
    ; pair B.i64_trunc_f32 (const [ S.Pop; S.Push (Num_type I64) ])
    ; pair B.i32_trunc_sat_f32 (const [ S.Pop; S.Push (Num_type I32) ])
    ; pair B.i64_trunc_sat_f32 (const [ S.Pop; S.Push (Num_type I64) ])
    ]
  else [] )
  @ B.local_set_f32 env @ B.local_tee_f32 env @ B.global_set_f32 env
```

test/fuzz/gen.ml
```diff
@@ -121,15 +137,18 @@ let expr_available_2_f32 =
...
   ]
-  @ B.local_set_f64 env @ B.local_tee_f64 env @ B.global_set_f64 env
+  @
+  if env.Env.conf = Concrete then
+    [ pair B.i32_trunc_f64 (const [ S.Pop; S.Push (Num_type I32) ])
+    ; pair B.i64_trunc_f64 (const [ S.Pop; S.Push (Num_type I64) ])
+    ; pair B.i32_trunc_sat_f64 (const [ S.Pop; S.Push (Num_type I32) ])
+    ; pair B.i64_trunc_sat_f64 (const [ S.Pop; S.Push (Num_type I64) ])
+    ]
+  else [] @ B.local_set_f64 env @ B.local_tee_f64 env @ B.global_set_f64 env
```
zapashcanon, line 151: Same:

```ocaml
  ( if env.Env.conf = Concrete then
    [ pair B.i32_trunc_f64 (const [ S.Pop; S.Push (Num_type I32) ])
    ; pair B.i64_trunc_f64 (const [ S.Pop; S.Push (Num_type I64) ])
    ; pair B.i32_trunc_sat_f64 (const [ S.Pop; S.Push (Num_type I32) ])
    ; pair B.i64_trunc_sat_f64 (const [ S.Pop; S.Push (Num_type I64) ])
    ]
  else [] )
  @ B.local_set_f64 env @ B.local_tee_f64 env @ B.global_set_f64 env
```

test/fuzz/basic.ml
epatrizio: same as above ...

test/fuzz/basic.ml
epatrizio: gen.ml :
```ocaml
if env.Env.conf = Concrete then
        [ pair B.extend_32_i32 (const [ S.Nothing ])
        ; pair B.i64_extend_i32 (const [ S.Pop; S.Push (Num_type I64) ])
        ; pair B.iunop_popcnt_32 (const [ S.Nothing ])
        ]
      else []
```

test/fuzz/gen.ml
epatrizio: Parenthesis like this: that's always been my intention ;)

test/fuzz/basic.ml
zapashcanon: I know it is not implemented, but I would like to get this PR ready and fix popcount implementation later.

test/fuzz/basic.ml
zapashcanon: same answer :)

test/fuzz/gen.ml
```diff
@@ -36,14 +37,18 @@ let expr_available_1_i32 if_else expr ~locals ~stack env =
     ]
   in
   [ pair B.iunop_32 (const [ S.Nothing ])
+  ; if_else expr ~locals ~stack env
```
zapashcanon, line 40: Could you put it as before to reduce the diff?

test/fuzz/gen.ml
```diff
@@ -36,14 +37,18 @@ let expr_available_1_i32 if_else expr ~locals ~stack env =
...
   ; pair B.f64_convert_i32 (const [ S.Pop; S.Push (Num_type F64) ])
   ; pair B.f32_reinterpret_i32 (const [ S.Pop; S.Push (Num_type F32) ])
-  ; if_else expr ~locals ~stack env
   ]
+  @ ( if env.Env.conf = Concrete then
+        [ pair B.extend_32_i32 (const [ S.Nothing ])
+        ; pair B.i64_extend_i32 (const [ S.Pop; S.Push (Num_type I64) ])
+        ; pair B.iunop_popcnt_32 (const [ S.Nothing ])
+        ]
+      else [] )
```
zapashcanon, line 51: Could you put all of them as before and let the fuzzer fail if they're not implemented ? We'll implemented them later.

test/fuzz/gen.ml
```diff
@@ -88,11 +93,16 @@ let expr_available_1_i64 env =
...
   ; pair B.f32_convert_i64 (const [ S.Pop; S.Push (Num_type F32) ])
   ; pair B.f64_convert_i64 (const [ S.Pop; S.Push (Num_type F64) ])
   ; pair B.f64_reinterpret_i64 (const [ S.Pop; S.Push (Num_type F64) ])
   ]
+  @ ( if env.Env.conf = Concrete then
+        [ pair B.iunop_popcnt_64 (const [ S.Nothing ]) ]
+      else [] )
+  @ ( if env.Env.conf = Concrete then
+        [ pair B.extend_64_i64 (const [ S.Nothing ]) ]
+      else [] )
```
zapashcanon, line 105: Could you put all of them as before and let the fuzzer fail if they're not implemented ? We'll implemented them later.

test/fuzz/gen.ml
```diff
@@ -103,14 +113,17 @@ let expr_available_2_i64 =
...
   ; pair B.i32_reinterpret_f32 (const [ S.Pop; S.Push (Num_type I32) ])
+  ; pair B.funop_32 (const [ S.Nothing ])
   ]
+  @ ( if env.Env.conf = Concrete then
+        [ pair B.i32_trunc_f32 (const [ S.Pop; S.Push (Num_type I32) ])
+        ; pair B.i64_trunc_f32 (const [ S.Pop; S.Push (Num_type I64) ])
+        ; pair B.i32_trunc_sat_f32 (const [ S.Pop; S.Push (Num_type I32) ])
+        ; pair B.i64_trunc_sat_f32 (const [ S.Pop; S.Push (Num_type I64) ])
+        ]
+      else [] )
```
zapashcanon, line 126: Could you put all of them as before and let the fuzzer fail if they're not implemented ? We'll implemented them later.

test/fuzz/gen.ml
```diff
@@ -121,14 +134,17 @@ let expr_available_2_f32 =
...
   ; pair B.i64_reinterpret_f64 (const [ S.Pop; S.Push (Num_type I64) ])
+  ; pair B.funop_64 (const [ S.Nothing ])
   ]
+  @ ( if env.Env.conf = Concrete then
+        [ pair B.i32_trunc_f64 (const [ S.Pop; S.Push (Num_type I32) ])
+        ; pair B.i64_trunc_f64 (const [ S.Pop; S.Push (Num_type I64) ])
+        ; pair B.i32_trunc_sat_f64 (const [ S.Pop; S.Push (Num_type I32) ])
+        ; pair B.i64_trunc_sat_f64 (const [ S.Pop; S.Push (Num_type I64) ])
+        ]
+      else [] )
```
zapashcanon, line 147: Could you put all of them as before and let the fuzzer fail if they're not implemented ? We'll implemented them later.

PR #191, opened 2024-02-23
simplify rotl and rotr

zapashcanon:


krtab:
Ideally: 
- the code would have comments detailing the semantic of each operation.
- There would be tests checking a) that rotl and rotr are indeed inverse of each other b) that rotating by 32 or 64 is the identity c) that rotating doesn't change popcount d) a few constant patterns

krtab (approved):

zapashcanon:
I believe there is already some tests in `test/script/` somewhere

PR #193, opened 2024-02-23
Implement the remaining functions in symbolic_table

### LINKED ISSUE #190: complete symbolic table implementation
Five functions are missing from `symbolic_table.ml`, I believe it should not be too hard to do but I'm not completely sure. Once it is done, we would need to re-enable the symbolic fuzzing for tables.

krtab:
For now this implementation needs to concretize every value. A possible improvement would be to make it aware of symbolic values.

Closes #190


src/symbolic_table.ml
```diff
@@ -2,34 +2,47 @@
...
+open Types
+
 module ITbl = Hashtbl.Make (struct
   include Int
 
   let hash x = x
 end)
 
-type t = Symbolic_value.ref_value array
+type table = Symbolic_value.ref_value array
```
zapashcanon, line 13: This is used only in one place so let's inline it.

zapashcanon:
LGTM modulo a minor nitpick.

zapashcanon:
Thanks!

Issue #162, opened 2024-02-13
Trouble with float sqrt instructions (in symbolic context)


epatrizio:
`f32.sqrt` and `f64.sqrt` instructions do not end.
So, in symbolic tests, these instructions have been commented out.

https://github.com/OCamlPro/owi/blob/2ef083c191e8fa1da71ceb6b46f5d2dc19334841/test/sym/unop_f32.wat#L37

https://github.com/OCamlPro/owi/blob/2ef083c191e8fa1da71ceb6b46f5d2dc19334841/test/sym/unop_f64.wat#L37

zapashcanon:
Small reproduction case:

```wat
(module
  (import "symbolic" "f32_symbol" (func $f32_symbol (result f32)))

  (func $start
    (local $f32 f32)

    (local.set $f32 (call $f32_symbol))

    (f32.eq (f32.sqrt (local.get $f32)) (f32.const 0))
    (if (then unreachable))

  )

  (start $start)
)
```

Actually, this is hanging because encoding is raising `Encoding.Solver.Unknown` and we do not properly handle this in `Symbolic_choice`.

So two questions:

- @filipeom, it is expected that Z3 can not handle sqrt ? After a quick search online, I believe we would have to use the non-linear solver or something like this ?
- @chambart, @krtab, could you have a look at how we could fix this in `Symbolic_choice` ? This is hopefully similar to ed033dfb7d665b941b19be3345834543a8c01231.

Also, on a related note, @filipeom, do you think we should change the type of `Encoding.Solver.check` (to a `bool option` for instance) to avoid using exceptions ?

filipeom:
>     * @filipeom, is it expected that Z3 can not handle `sqrt` ? After a quick search online, I believe we would have to use the non-linear solver or something like this ?

Z3 should be able to handle the `sqrt` (but really slooowly, see https://github.com/formalsec/encoding/commit/92ae62526cfb1fc9bb7b872622271fad61493971)
 
> Also, on a related note, @filipeom, do you think we should change the type of `Encoding.Solver.check` (to a `bool option` for instance) to avoid using exceptions ?

Yes, I agree this should be done :smiley: Either an option or a ADT like [type satisfiabilty](https://github.com/formalsec/encoding/blob/92ae62526cfb1fc9bb7b872622271fad61493971/lib/mappings_intf.ml#L1) to be more expressive

filipeom:
Also, using `assert`:

```wast
    (call $assert
      (f32.ne
[...]

PR #194, opened 2024-02-26
refacto code coverage specific tests

epatrizio:
Based on `div_zero.wat` example 
https://github.com/OCamlPro/owi/blob/main/test/sym/div_zero.wat


epatrizio:
in `binop_f64.wat`, some tests have been deactivated for now (`div`, `mul`, `copysign`). Reason: very long computing time!
Same problem in copysign.wat : https://github.com/OCamlPro/owi/blob/main/test/sym/copysign.t

zapashcanon:
OK, I expect it to be the same than for `sqrt` and to be caused by bit-blasting performed by the SMT. Could list the precise list of instructions that are affected in #162 ?

epatrizio:
> in `binop_f64.wat`, some tests have been deactivated for now (`div`, `mul`, `copysign`). Reason: very long computing time! Same problem in copysign.wat : https://github.com/OCamlPro/owi/blob/main/test/sym/copysign.t

After further testing, I reactivated `copysign` (~ 1sec calculation). `div` and `mul` operations are very time-consuming (~ 25 sec !)
Maybe we could reactivate `copysign_f64.wat` (~ 5sec)

epatrizio:
CI is broken in macos context !? :thinking:

zapashcanon:
It is not really broken, but our tests are not deterministic enough.

epatrizio:
@krtab suggested me adding the option `-w1` in cram tests command line to obtain determinism.
It's not enough ?
For example :
```console
dune exec owi -- sym unop_i32.wat --no-stop-at-failure --no-value -w1
```

zapashcanon:
No, this makes it more deterministic but not fully.

epatrizio:
Waiting for merging PR #197 (determinism WIP)

epatrizio:
> Waiting for merging PR #197 (determinism WIP)

Rebase done!

epatrizio:
> Maybe we could reactivate `copysign_f64.wat` (~ 5sec)

Done!

zapashcanon:
Did you rebase and promote tests properly ? CI is not happy

krtab:
I have figured out why tests are not deterministic.

Even if we try to sort the results using the breadcrumbs mechanism introduced in #210 , the name of the symbols themselves depend on the execution order, and are hence non deterministic. This can be solved but not immediately.

@epatrizio I'm afraid you'll have to split these tests in single files.

krtab:
Could you check that the coverage does not decrease between right before your commit and right after it?

epatrizio:
The coverage increases slightly : 86.13% > 86.26%
This comes from the code coverage of the `src/symbolic.value.ml` file : 83% (202 / 242) > 85% (206 / 242)

krtab:
> The coverage increases slightly : 86.13% > 86.26% This comes from the code coverage of the `src/symbolic.value.ml` file : 83% (202 / 242) > 85% (206 / 242)

Thanks a lot!

zapashcanon:
Thanks!

PR #197, opened 2024-02-29
make test outputs lazy

epatrizio:
Step 1 = make test outputs lazy (first commit)

Step 2 = make test outputs deterministic (second commit)
`--no-value` and `--w1` command options doesn't seem to be enough.


epatrizio:
In commit https://github.com/OCamlPro/owi/pull/197/commits/24c45330be48696060e024ce04c3c70f7a583f3c : Add `--no-sort-sym-result` command line option.
For example :
`dune exec owi -- sym ./test/sym/unop_f32.wat --no-stop-at-failure --no-sort-sym-results -w1` >> (Exec. time = ~10s)
`dune exec owi -- sym ./test/sym/unop_f32.wat --no-stop-at-failure -w1` >> (Exec. time = ~20s) Results are sorted before display. The `lazy` approach is lost, but we hope for output determinism.

Nb. In cram tests, `--no-sort-sym-results` option must not be used.

epatrizio:
CI is again broken in macos context !
But this doesn't seem to be linked to a problem of determinism ... ?

epatrizio:
@krtab review : reverse command line option interpretation (so change its name)
`--no-sort-sym-results` becomes `--deterministic-result-order`

src/bin/owi.ml
```diff
@@ -35,6 +35,12 @@ let no_values =
   let doc = "do not display a value for each symbol" in
   Cmdliner.Arg.(value & flag & info [ "no-value" ] ~doc)
 
+let deterministic_result_order =
+  let doc =
+    "sort symbolic results before display to obtain a deterministic order"
```
krtab, line 40:
```suggestion
    "Guarantee a fixed deterministic order of found failures. "
```

src/cmd_sym.ml
```diff
@@ -197,50 +197,58 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
...
+  let rec print_and_count_failures count results =
+    match results () with
+    | Seq.Nil -> Ok count
+    | Seq.Cons ((result, thread), tl) ->
+      let pc = Thread.pc thread in
+      let symbols = thread.symbol_set in
+      let model = get_model ~symbols solver pc in
+      let* count, err =
+        match result with
+        | Symbolic_choice.Multicore.EAssert assertion ->
```
krtab, line 223: Would this work?
```suggestion
       let open Symbolic_choice.Multicore in
        match result with
        | EAssert assertion ->
```

src/cmd_sym.ml
```diff
@@ -197,50 +197,58 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
...
+      Format.pp_flush Stdlib.Format.std_formatter ()
+  in
+  let rec print_and_count_failures count results =
+    match results () with
+    | Seq.Nil -> Ok count
+    | Seq.Cons ((result, thread), tl) ->
+      let pc = Thread.pc thread in
+      let symbols = thread.symbol_set in
+      let model = get_model ~symbols solver pc in
+      let* count, err =
```
krtab, line 221: `err` -> `is_err` to hint that is is a boolean

src/cmd_sym.ml
```diff
@@ -197,50 +197,58 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
...
+        match result with
+        | Symbolic_choice.Multicore.EAssert assertion ->
+          print_bug (`EAssert (assertion, model));
+          Ok (succ count, true)
+        | ETrap tr ->
+          print_bug (`ETrap (tr, model));
+          Ok (succ count, true)
+        | EVal (Ok ()) -> Ok (count, false)
```
krtab, line 229: `count` is incremented if and only if `is_err == true` so this could be factored out, which would makes the intent clearer IMO.

krtab (changes requested):
Thanks!

On top of my comments, I think it would be better to make it so that the deterministic order implies no-stop-at-failure, because we anyway have to go through all possible executions.

krtab:
> Example:
> `dune exec owi -- sym ./test/sym/unop_f32.wat --no-stop-at-failure` (non-deterministic, Exec. time = ~10s, manual context)
> `dune exec owi -- sym ./test/sym/unop_f32.wat --no-stop-at-failure --deterministic-result-order` (deterministic, Exec. time = ~20s, automatic context - cram tests)

Could you check both manually? The runtime should really not be that different. If they remain so far apart we will have to bench to understand what is going on. (I would guess the compare when ordering the results sequence)

krtab:
Also, I don't see any test checking errors output in out test base. Could you add one, for example one with a file with bad syntax, and one with a non existing file like in the tests for run and validate?

src/cmd_sym.ml
```diff
@@ -239,6 +239,15 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
       if no_stop_at_failure || not err then print_and_count_failures count tl
       else Ok count
   in
+  let sort_results deterministic_result_order res =
+    if deterministic_result_order then
+      let res = List.of_seq res in
```
zapashcanon, line 244: Not introduced in your PR but can we make `res` a `list` instead of a `seq` ? I would not be surprise if there was no reason for it to be a `seq`...

src/cmd_sym.ml
```diff
@@ -197,50 +197,49 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
...
-      [] (List.of_seq results)
-  in
   let print_bug = function
     | `ETrap (tr, model) ->
       Format.pp_std "Trap: %s@\n" (Trap.to_string tr);
       Format.pp_std "Model:@\n  @[<v>%a@]@\n"
         (Encoding.Model.pp ~no_values)
-        model
+        model;
+      Format.pp_flush Stdlib.Format.std_formatter ()
```
zapashcanon, line 206: This should not be needed.

src/cmd_sym.ml
```diff
@@ -197,50 +197,49 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
...
+          Ok (succ count, true)
+        | ETrap tr ->
+          print_bug (`ETrap (tr, model));
+          Ok (succ count, true)
+        | EVal (Ok ()) -> Ok (count, false)
+        | EVal (Error e) -> Error e
+      in
+      if not no_values then
+        let testcase =
+          List.sort compare (Encoding.Model.get_bindings model) |> List.map snd
```
zapashcanon, line 234: I believe `Encoding` is already sorting the elements given by `get_bindings` (I wrote the code to do it IIRC).

src/cmd_sym.ml
```diff
@@ -197,50 +197,49 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
...
+        in
+        let _ = write_testcase ~dir:workspace ~err testcase in
```
zapashcanon, line 236: Please, no `let _ =`.

src/cmd_sym.ml
```diff
@@ -197,50 +197,49 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
...
+        ()
+      else ();
```
zapashcanon, line 238: This can be removed.

src/cmd_sym.ml
krtab: There is very much reason to do. A Seq is lazy and hence allows outputting issues as soon as they are found, and not wait for the whole program to have been explored (which may not even terminate)

krtab:
Also @epatrizio. I find that the code base really lacks comments, so from now on we should all strive to add comments to parts of the code we're touching. A good rule of thumb to add comments is that they should contain the info you would say to someone if you were reading the code with them. Could you add some comments to the code you wrote in this PR ? Thanks

epatrizio:
> > Example:
> > `dune exec owi -- sym ./test/sym/unop_f32.wat --no-stop-at-failure` (non-deterministic, Exec. time = ~10s, manual context)
> > `dune exec owi -- sym ./test/sym/unop_f32.wat --no-stop-at-failure --deterministic-result-order` (deterministic, Exec. time = ~20s, automatic context - cram tests)
> 
> Could you check both manually? The runtime should really not be that different. If they remain so far apart we will have to bench to understand what is going on. (I would guess the compare when ordering the results sequence)

The 2 outputs are very different (with the `--no-value` option, they are the same).
I'm not quite sure I understand the question, It would be better to have a work session together.

src/cmd_sym.ml
zapashcanon: Fine. I have a vague remembering of wanting to remove it at some point as it seemed not needed, but I'm fine with keeping it.

src/cmd_sym.ml
epatrizio: Seq is lazy. When I add a `flush`, the display is progressive. Is there a better way to do this?

zapashcanon: I don't understand what this has to do with `Seq` being lazy ? The display should be strict even without the flush here.

src/cmd_sym.ml
epatrizio: This `List.sort` was present before this PR.
Has `Encoding` evolved in the meantime? If so, should this part be removed? (See next comment ...)

src/cmd_sym.ml
epatrizio: Same, see comment above ...

src/cmd_sym.ml
zapashcanon: Oh indeed. You can try to remove it and see if it changes anything.

src/cmd_sym.ml
zapashcanon: I mean, writing:

```ocaml
if .. then
  ... ()
else ();
```

is the same as:

```ocaml
if ... then
  ... ();
```

src/cmd_sym.ml
epatrizio: Ok, the `List.sort` can be removed. Also the `write_testcase` call just after ? (I don't know what it's for)
I've just tried deleting everything: it's all ok!

epatrizio:
> Also, I don't see any test checking errors output in out test base. Could you add one, for example one with a file with bad syntax, and one with a non existing file like in the tests for run and validate?

`not_exists.t` cram test already exists

src/cmd_sym.ml
epatrizio: I don't know :( Without the `flush`, the display is not progressive and appears all at once at the end.
If there is a better way to do this, no problem ...

filipeom: It's also my experience that `@\n` is buffered and the output only appears at the end. I tend to use `@.` to insert newlines when I'm terminating a formatter, it supposedly also does a `flush`. 

To this end, we could terminate the model print with `@.` (i.e., `Format.pp_std "Model:@\n  @[<v>%a@]@."`) 

It has the same intended format of being a progressive print, but its exactly the same to what is being done here.

src/cmd_sym.ml
filipeom: When this was done encoding wouldn't sort elements. So it can be removed. However, this code is still wrong. 

The outputed testcase  is given is in a lexicographical order, making stuff like `symbol_10` appearing before `symbol_2` which is obviously wrong.

filipeom: But we can use `symbols` to return the correct testcase, something like:

```ocaml
          let testcase =
            List.map
              (fun sym ->
                match Encoding.Model.evaluate model sym with
                | None -> assert false (* unreachable due to line 205 *)
                | Some v -> v )
              symbols
            |> List.rev
          in

```

filipeom: Oh I didn't see the comment about `write_testcase`. We probably don't want to delete it entirely as these are the concrete inputs for the symbolic program. But we should put it under some flag so that they are only generated when its on?

epatrizio:
> Also, I don't see any test checking errors output in out test base. Could you add one, for example one with a file with bad syntax, and one with a non existing file like in the tests for run and validate?

Done! Add `syntax_error.t` cram test.

src/cmd_sym.ml
epatrizio: Thanks @filipeom !
Works great with `@.`

epatrizio:
> On top of my comments, I think it would be better to make it so that the deterministic order implies no-stop-at-failure, because we anyway have to go through all possible executions.

I understand the intent. But could it be confusing to implicitly link two command line options?

zapashcanon:
> Also, I don't see any test checking errors output in out test base. Could you add one, for example one with a file with bad syntax, and one with a non existing file like in the tests for run and validate?

We have tests for that, they're part of the Wasm script format (see test/script/reference/).

krtab:
@epatrizio 
> I understand the intent. But could it be confusing to implicitly link two command line options?

No this is rather classic, and should be make explicit in the option's documentation. Search for example the word "implies" in https://manpages.debian.org/testing/ripgrep/rg.1.en.html

krtab:
> We have tests for that, they're part of the Wasm script format (see test/script/reference/).

Well my point is we should have a regression test as part of the suite that is launched with dune runtest. Currently if I do on master 

```diff
diff --git a/src/cmd_sym.ml b/src/cmd_sym.ml
index 6fe5ecad..b9a34a50 100644
--- a/src/cmd_sym.ml
+++ b/src/cmd_sym.ml
@@ -209,7 +209,7 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
           | EAssert assertion ->
             Ok (`EAssert (assertion, model) :: failing, true)
           | ETrap tr -> Ok (`ETrap (tr, model) :: failing, true)
-          | EVal (Error e) -> Error e
+          | EVal (Error _e) -> Ok (failing, false)
         in
         if not no_values then
           let testcase =
```

And launch `dune runtest` I don't have any error, and we had this regression in this refactoring at some point with @epatrizio.

zapashcanon:
Oh yes, we are missing a lot of those (easy to see by looking at the coverage report), I thought you were talking about bad syntax errors.

epatrizio:
> No this is rather classic, and should be make explicit in the option's documentation. Search for example the word "implies" in https://manpages.debian.org/testing/ripgrep/rg.1.en.html

OK, done!

src/cmd_sym.ml
krtab: I'm not sure I've followed what you guys wanted around the write_testcase code but given that this was here before this PR and is pretty orthogonal I suggest that we make an issue for it and fix it later

zapashcanon: It was changed to `let _ = write_testcase ...` and I was saying it should stay as before: `let+ () = write_testcase ...` (which is the case now).

src/cmd_sym.ml
```diff
@@ -197,50 +203,59 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
...
         (Encoding.Model.pp ~no_values)
         model
   in
-  let rec print_and_count_failures count = function
-    | [] -> count
-    | hd :: tl ->
-      print_bug hd;
-      let count = succ count in
-      if no_stop_at_failure then print_and_count_failures count tl else count
+  (* Solver results analysis *)
```
krtab, line 218:
```suggestion
```

krtab (approved):

src/cmd_sym.ml
```diff
@@ -197,50 +203,58 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
...
+          ()
+        else Ok ()
+      in
+      if (not is_err) || no_stop_at_failure then
+        print_and_count_failures count_acc tl
+      else Ok count_acc
+  in
+  let results =
+    if deterministic_result_order then
+      List.to_seq @@ List.sort compare @@ List.of_seq results
```
zapashcanon, line 254: I don't understand why we are going back to `Seq` here. The `List.to_seq` could be removed (and this would allow to write `print_and_count_failures` as `list_fold_left` as it was before (adding the count to the accumulator ofc)

src/cmd_sym.ml
```diff
@@ -197,50 +203,58 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
...
+      in
+      let count_acc = if is_err then succ count_acc else count_acc in
+      let* () =
+        if not no_values then
+          let testcase =
+            List.sort compare (Encoding.Model.get_bindings model)
+            |> List.map snd
+          in
+          let+ () = write_testcase ~dir:workspace ~err:is_err testcase in
+          ()
```
zapashcanon, line 245: `let+ () = e in ()` is equivalent to `e`.

src/cmd_sym.ml
krtab: To maintain laziness if we don't need the deterministic order.

zapashcanon: I don't understand.

First, here we need the deterministic order for sure because we are in the `then` branch of `if deterministic_result_order then ...`.

Then, running `List.of_seq` already breaks the laziness. And there is no point in going back to `Seq` after that in this case ?

zapashcanon: Hum OK. Got it. Sorry...

PR #198, opened 2024-03-01
add a flag to print the solver statistics

### LINKED ISSUE #184: add a flag to print the solver statistics
the quick&dirty way to do it for reference:

```patch
diff --git a/src/symbolic_choice.ml b/src/symbolic_choice.ml
index e799358..534a023 100644
--- a/src/symbolic_choice.ml
+++ b/src/symbolic_choice.ml
@@ -19,7 +19,10 @@ let check sym_bool thread (S (solver_module, solver)) =
   | _ ->
     let check = no :: pc in
+
     let r = Solver.check solver check in
+        Solver.pp_statistics Stdlib.Format.std_formatter solver;
+        Stdlib.Format.pp_print_flush Stdlib.Format.std_formatter ();
     not r
 
 let list_select b ({ Thread.pc; _ } as thread) (S (solver_module, s)) =
@@ -34,6 +37,9 @@ let list_select b ({ Thread.pc; _ } as thread) (S (solver_module, s)) =
     let with_not_v = Bool.not v :: pc in
     let sat_true = Solver.check s with_v in
     let sat_false = Solver.check s with_not_v in
+
+        Solver.pp_statistics Stdlib.Format.std_formatter s;
+        Stdlib.Format.pp_print_flush Stdlib.Format.std_formatter ();
     match (sat_true, sat_false) with
     | false, false -> []
     | true, false | false, true -> [ (sat_true, thread) ]
@@ -78,6 +84,8 @@ let list_select_i32 sym_int thread (S (solver_module, solver)) =
       if not (Solver.check solver (additionnal @ pc)) then []
       else begin
         let model = Solver.model ~symbols:[ symbol ] solver in
+        Solver.pp_statistics Stdlib.Format.std_formatter solver;
+        Stdlib.Format.pp_print_flush Stdlib.Format.std_formatter ();
         match model with
         | None -> assert false (* ? *)
         | Some model -> (
```

epatrizio:
fixes #184

### DIFF
--- a/example/c/README.md
+++ b/example/c/README.md
@@ -35,7 +35,7 @@ $ owi c ./poly.c -w1
 ...
 Model:
   (model
-    (symbol_0 (i32 2)))
+    (symbol_0 (i32 4)))
 Reached problem!
 [13]
 ```
@@ -318,7 +318,8 @@ OPTIONS
            optimize mode
 
        -p, --profiling
-           profiling mode
+           profiling mode (including solver statistics display in the context
+           of symbolic execution)
 
        --property=VAL
            property file
@@ -333,9 +334,8 @@ OPTIONS
            skip typechecking pass
 
        -w VAL, --workers=VAL (absent=n)
-           number of workers for symbolic execution. Defaults to a
-           machine-specific value given by the OCaml
-           Domain.recommended_domain_count function.
+           number of workers for symbolic execution. Defaults to the number
+           of physical cores.
 
 COMMON OPTIONS
        --help[=FMT] (default=auto)
--- a/example/conc/README.md
+++ b/example/conc/README.md
@@ -63,7 +63,8 @@ OPTIONS
            optimize mode
 
        -p, --profiling
-           profiling mode
+           profiling mode (including solver statistics display in the context
+           of symbolic execution)
 
        -s VAL, --solver=VAL (absent=Z3)
            SMT solver to use
@@ -72,9 +73,8 @@ OPTIONS
            skip typechecking pass
 
        -w VAL, --workers=VAL (absent=n)
-           number of workers for symbolic execution. Defaults to a
-           machine-specific value given by the OCaml
-           Domain.recommended_domain_count function.
+           number of workers for symbolic execution. Defaults to the number
+           of physical cores.
 
        --workspace=VAL (absent=owi-out)
            path to the workspace directory
--- a/example/run/README.md
+++ b/example/run/README.md
@@ -68,7 +68,8 @@ OPTIONS
            optimize mode
 
        -p, --profiling
-           profiling mode
+           profiling mode (including solver statistics display in the context
+           of symbolic execution)
 
        -u, --unsafe
            skip typechecking pass
--- a/example/sym/README.md
+++ b/example/sym/README.md
@@ -67,7 +67,8 @@ OPTIONS
            optimize mode
 
        -p, --profiling
-           profiling mode
+           profiling mode (including solver statistics display in the context
+           of symbolic execution)
 
        -s VAL, --solver=VAL (absent=Z3)
            SMT solver to use
@@ -76,9 +77,8 @@ OPTIONS
            skip typechecking pass
 
        -w VAL, --workers=VAL (absent=n)
-           number of workers for symbolic execution. Defaults to a
-           machine-specific value given by the OCaml
-           Domain.recommended_domain_count function.
+           number of workers for symbolic execution. Defaults to the number
+           of physical cores.
 
        --workspace=VAL (absent=owi-out)
            path to the workspace directory
--- a/src/bin/owi.ml
+++ b/src/bin/owi.ml
@@ -56,7 +56,10 @@ let optimize =
   Cmdliner.Arg.(value & flag & info [ "optimize" ] ~doc)
 
 let profiling =
-  let doc = "profiling mode" in
+  let doc =
+    "profiling mode (including solver statistics display in the context of \
+     symbolic execution)"
+  in
   Cmdliner.Arg.(value & flag & info [ "profiling"; "p" ] ~doc)
 
 let solver =
--- a/src/cmd/cmd_sym.ml
+++ b/src/cmd/cmd_sym.ml
@@ -112,6 +112,8 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
     else results
   in
   let* count = print_and_count_failures 0 results in
+  if profiling then
+    Format.pp_std "%a@\n" Smtml.Solver_dispatcher.pp_solver_type solver;
   if count > 0 then Error (`Found_bug count)
   else begin
     Format.pp_std "All OK";


epatrizio:
Command line example :
```console
dune exec owi -- sym ./test/sym/table.wat --no-stop-at-failure --print-solver-statistics
```
Display :
```console
(decisions 3)                          
(propagations 62)
(final checks 1)
(mk clause 98)
(num checks 1)
(mk bool var 66)
(arith-make-feasible 1)
(arith-max-columns 4)
(num allocs 149343)
(rlimit count 240)
(max memory 84.94)
(memory 68.2)(decisions 3)
(propagations 129)
(final checks 2)
(added eqs 5)
(mk clause 199)
(del clause 98)
(num checks 2)
(mk bool var 169)
(arith-make-feasible 2)
(arith-max-columns 4)
(bv bit2core 32)
(num allocs 153416)
(rlimit count 665)
(max memory 84.94)
(memory 68.3)(decisions 34)
(propagations 197)
(final checks 3)
(added eqs 8)
(mk clause 302)
(del clause 199)
(num checks 3)
(mk bool var 272)
(arith-make-feasible 3)
(arith-max-columns 4)
(bv dynamic diseqs 2)
(bv bit2core 32)
(bv->core eq 1)
(num allocs 153416)
(rlimit count 1350)
(max memory 84.94)
(memory 68.3)Trap: unreachable
Model:
  (model
    (symbol_0 (i32 0))
    (symbol_1 (i32 0)))
Trap: undefined element
Model:
  (model
    (symbol_0 (i32 0))
    (symbol_1 (i32 -2147483648)))
Reached 2 problems!
```
Does the `Owi.Solver.pp_statistics` function need to be improved?
``` ocaml
let pp_statistics solver =
  Z3Batch.pp_statistics Stdlib.Format.std_formatter solver;
  Stdlib.Format.pp_print_flush Stdlib.Format.std_formatter ()
```

src/cmd_sym.ml
```diff
@@ -202,7 +206,7 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
       (fun failing (result, thread) ->
         let pc = Thread.pc thread in
         let symbols = thread.symbol_set in
-        let model = get_model ~symbols solver pc in
+        let model = get_model ~symbols solver pc print_solver_statistics in
```
krtab, line 209: I think I would prefer that the printing is done directly in cmd rather than in get_model. If it is meant as summary statistics, it should be done once at the end.

krtab (changes requested):
Thanks!

Before merging, I think we must check and clarify what statistics are displayed exactly. If there is n worker threads, then there is a total of n+1 solver objects. One for each worker thread and one in the main thread (the one you're using). I don't know whether the statistics are computed by z3 for all solver objects combined or only for each one (my guess is that it is only for each one). If it is only for each one, the meaningful thing to do would be combining all these statistics as one, and display that.

src/solver.ml
```diff
@@ -11,3 +11,7 @@ let fresh_solver () =
   let module Batch = Encoding.Solver.Batch (Mapping) in
   let solver = Batch.create ~logic:QF_BVFP () in
   S ((module Batch), solver)
+
+let pp_statistics solver =
+  Z3Batch.pp_statistics Stdlib.Format.std_formatter solver;
+  Stdlib.Format.pp_print_flush Stdlib.Format.std_formatter ()
```
zapashcanon, line 17: You can probably remove this function.

src/cmd_sym.ml
```diff
@@ -151,11 +151,15 @@ let run_file ~unsafe ~optimize pc filename =
...
   assert (Solver.Z3Batch.check solver pc);
-  match Solver.Z3Batch.model ~symbols solver with
-  | None -> assert false
-  | Some model -> model
+  let model =
+    match Solver.Z3Batch.model ~symbols solver with
+    | None -> assert false
+    | Some model -> model
+  in
+  if print_statistics then Solver.pp_statistics solver;
```
zapashcanon, line 161:
```ocaml
if ... then
  Format.pp_std "%a@\n" Z3.pp_statistics solver;
model
```

krtab:
Could you rebase on #199?

zapashcanon:
I'm wondering if we would like to merge this with the `--profiling` flag ?

filipeom:
> I don't know whether the statistics are computed by z3 for all solver objects combined or only for each one (my guess is that it is only for each one).

To clarify, what `pp_statistics` does is print the statistics related to each solver object (Your guess is correct). You can confirm this looking at the number of solver checks printed in the output Eric gave. The maximum is `(num checks 3)` which only represents the checks done for by the solver in the main thread. 

>  If it is only for each one, the meaningful thing to do would be combining all these statistics as one, and display that.

It doesn't seem trivial to do here because a solver object allocated in a thread will be collected when the thread finishes (correct me if I'm wrong). Therefore, we would need to incrementally combine solver statistics as threads finish.

epatrizio:
> I'm wondering if we would like to merge this with the `--profiling` flag ?

```console
dune exec owi -- sym ./test/sym/table.wat --no-stop-at-failure --profiling
```
displays :
```text
Exec module anonymous                  
enter 1
intrs 5
Exec module anonymous
enter 1
intrs 1
calls
  id 0 start
    enter 1
    intrs 3
    calls
      id 1 f
        enter 1
        intrs 3
Trap: undefined element
Model:
  (model
    (symbol_0 (i32 0))
    (symbol_1 (i32 -2147483648)))
Trap: unreachable
Model:
  (model
    (symbol_0 (i32 0))
    (symbol_1 (i32 0)))
(decisions 34)
(propagations 197)
(final checks 3)
(added eqs 8)
(mk clause 302)
(del clause 201)
(num checks 3)
(mk bool var 272)
(arith-make-feasible 3)
(arith-max-columns 4)
(bv dynamic diseqs 2)
(bv bit2core 32)
(bv->core eq 1)
(num allocs 153329)
(rlimit count 1350)
(max memory 85.6)
(memory 68.3)
Reached 2 problems!%
```

krtab:
> It doesn't seem trivial to do here because a solver object allocated in a thread will be collected when the thread finishes (correct me if I'm wrong). Therefore, we would need to incrementally combine solver statistics as threads finish.

Though not trivial, it shouldn't be too hard either. There is one solver per worker, and a defined number of workers. At least in the nominal case where no worker encounters an unrecoverable error, it should be doable for each worker too record its own solver's statistics somewhere and for the main thread to collect these aggregated statistics once every worker is finished. 

This depends of formalsec/encoding#85 to be done but there is always the possibility of doing it "dirty" by simply parsing the output so we should manage to get something that works anyway.

krtab:
I think that even with `-w1` the results are meaningless. `-w1` doesn't deactivate the multicore monad, it simply spans only one worker iirc.

zapashcanon:
Oh so, even inside a single worker, the Z3 statistics are not shared?

zapashcanon:
I'm closing this one. We'll need to gather all solver statistics and merge them somehow.

PR #203, opened 2024-03-05
In tests, change dune exec for direct binary call

krtab:
Using `dune exec`` means that the dune-project file needs to be present in the dune sandbox, which
would sometimes not happen and lead to the very
annoying workflow of having to dune clean before
every dune runtest.

I tried adding the dune-project file as a dependency but couldn't get it to work.

A direct call to the binary seems to solve the issue.

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -18,7 +18,7 @@
 ## Man page
 
 ```sh
-$ dune exec -- owi --help=plain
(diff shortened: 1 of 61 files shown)


zapashcanon:
Nice! Thanks! I don't remember why I used `dune exec` in the first place...

PR #204, opened 2024-03-06
Lifts most simplification concerns to encoding

filipeom:
This focuses on removing simplifications from `Symbolic_value` to encoding. Please let me know if I should also move any other code to encoding.

Since this depends on formalsec/encoding#86, CI will fail until it is merged. I haven't done it to avoid breaking any ongoing work on owi. However, I've run the tests locally, and everything is okay, including performance, which seems to be the same.

### DIFF
--- a/src/symbolic_value.ml
+++ b/src/symbolic_value.ml
@@ -5,41 +5,6 @@
 open Encoding
 open Ty
 open Expr
-open Hc
-
-let ( let+ ) o f = Option.map f o
-
-let unop ty op e =
-  match e.node.e with
-  | Val (Num n) -> Val (Num (Eval_numeric.eval_unop ty op n)) @: e.node.ty
-  | _ -> Unop (op, e) @: ty
-
-let binop ty op e1 e2 =
-  match (e1.node.e, e2.node.e) with
-  | Val (Num n1), Val (Num n2) ->
-    Val (Num (Eval_numeric.eval_binop ty op n1 n2)) @: ty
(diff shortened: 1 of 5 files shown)


src/symbolic_value.ml
```diff
@@ -113,63 +78,34 @@ module Ref = struct
...
-  let and_ e1 e2 =
-    match (of_val e1.node.e, of_val e2.node.e) with
-    | Some b1, Some b2 -> const (b1 && b2)
-    | Some true, _ -> e2
-    | _, Some true -> e1
-    | Some false, _ | _, Some false -> const false
-    | _ -> Binop (And, e1, e2) @: Ty_bool
+  let or_ e1 e2 = Bool.or_ e1 e2
+
+  let and_ e1 e2 = Bool.and_ e1 e2
```
zapashcanon, line 87: We could `include Bool` to avoid redefining all these ? I also prefer using `v` to `const` so I'm fine with changing it in Owi

filipeom: Certainly, I can do that. However, I'm concerned about simply changing `Value.Bool.const` to `Value.Bool.v` because it would no longer align with the consistency of value constructors for constants, such as `Value.const_i32`, `Value.const_i64`, `Value.const_f32`, and `Value.const_f64`.

One solution is to move these constructors into their respective modules, transforming, for example, `Value.const_i32` into `Value.I32.v`. 

What would you prefer?

chambart (approved):
Good to go after `dune fmt`

Issue #201, opened 2024-03-05
use proper `conf-*` packages instead of depexts


zapashcanon:
for instance `conf-clang` instead of the `clang` depext

krtab:
The only one that can currently be done seems to be clang, and maybe llvm, if using version 16 is fine. Both lld and wabt still have to be explicit depexts.

filipeom:
Also, if we could adapt the reference interpreter's binary decoder we wouldn't need `wabt` https://github.com/WebAssembly/spec/blob/main/interpreter/binary/decode.ml

zapashcanon:
I added the missing packages in https://github.com/ocaml/opam-repository/pull/27381 ; we should be able to try it now

PR #205, opened 2024-03-06
Change clang depext for conf-clang dependency

krtab:
conf-clang allows for more platforms to be supported


zapashcanon:
Related to #201

PR #206, opened 2024-03-06
Change llvm depext for conf-llvm dependency

krtab:


.github/workflows/build.yml
```diff
@@ -51,15 +51,9 @@ jobs:
...
         run: |
-          if [ "$RUNNER_OS" == "macOS" ]; then
-            export PATH="$(brew --prefix llvm@15)/bin/:$PATH" # macossucks
-          fi
           opam exec -- dune build @install
       - name: test
         run: |
-          if [ "$RUNNER_OS" == "macOS" ]; then
-            export PATH="$(brew --prefix llvm@15)/bin/:$PATH" # macossucks
-          fi
```
krtab, line 62: Curious to see if the conf- package indeed allows us to remove these lines

krtab:
Hm, this also breaks the dependency on `opt`... I don't think it will end up being much better than what we have, which is at least working.

.github/workflows/build.yml
filipeom: Did you try using `conf-llvm` without removing this? It seems that `llvm-conf` has the same `depext` has us. So it would probably still need this

https://github.com/ocaml/opam-repository/blob/3f84051bba2dea78fb8f004d75421604ad53b166/packages/conf-llvm/conf-llvm.16/opam#L14

PR #209, opened 2024-03-11
Document Env_id

krtab:


krtab:
I'm realizing now that there is also a Func_id module, highly similar, but I guess my comments should be changed to reflect that

Issue #208, opened 2024-03-11
Investigate non determinism of CI


krtab:
In the following CI run https://github.com/OCamlPro/owi/actions/runs/8158780403/job/22301476759, there seems to be non determinism despite `--deterministic-result-order`. 

To do:

- [ ] Check that the discrepancy is indeed an order issue. To do so, apply the CI's git diff, and manually compare.
- [ ] If it is indeed an order issue, this is a bug of `deterministic-result-order- and should be fixed.

krtab:
Assigning high priority because it is failing the CI

PR #210, opened 2024-03-11
Add breadcrumbs to Thread.t

krtab:
As identified in https://github.com/OCamlPro/owi/issues/208, the
deterministic order option is not deterministic across architectures.

A likely culprit is the use of the polymorphic compare function which
may not give the same result in different environments or on different
targets.

This commit resolves this by keeping track of each choice made ('select')
through "breadcrumbs" and sorting results on these breadcrumbs.

Closes: https://github.com/OCamlPro/owi/issues/208


chambart (approved):
Seems good

PR #221, opened 2024-03-15
binary format deserializer

epatrizio:


krtab:
- src/cmd_run.ml
- src/named.ml
- src/named.mli 

Have been moved in #229 and #226

src/binary_basic.ml
```diff
@@ -0,0 +1,471 @@
...
+
+  let get n input =
+    if n < input.size then Some (Bytes.get input.bytes (input.pt + n)) else None
+
+  let get0 = get 0
+end
+
+let read_byte input =
+  match Input.get0 input with
+  | None -> failwith "read_byte error"
```
zapashcanon, line 39: Could you start using the error monad ? Returning `Error "read_byte error"` will be enough to start.

src/binary_basic.ml
```diff
@@ -0,0 +1,471 @@
...
+  | Some c ->
+    let next_input = Input.sub_suffix 1 input in
+    let i = Char.code c in
+    let i = if i < 64 then i else i - 128 in
+    (* 64=2⁶ *)
+    (i, next_input)
+
+let read_S64 = read_S32
+
+let deserialize_indice = read_U32
```
zapashcanon, line 69: I feel like writing:


```ocaml
let deserialize_indice input =
  let indice, input = read_U32 input in
  (Raw indice), input
```

would simplify the code in many places.

src/binary_basic.ml
```diff
@@ -0,0 +1,471 @@
...
+  | '\x7F' -> (Some (Num_type I32), next_input)
+  | '\x7E' -> (Some (Num_type I64), next_input)
+  | '\x7D' -> (Some (Num_type F32), next_input)
+  | '\x7C' -> (Some (Num_type F64), next_input)
+  | _ -> (None, input)
+
+let deserialize_vectype input =
+  let b, _next_input = read_byte input in
+  match b with (* | '\x7B' -> (Some V128, input) *)
+  | _ -> (None, input)
```
zapashcanon, line 83:
```ocaml
  match b with
  | '\x7B' -> assert false (* (Some V128, input) *)
  | _ -> (None, input)
```

src/binary_basic.ml
```diff
@@ -0,0 +1,471 @@
...
+
+let deserialize_reftype input =
+  let b, _next_input = read_byte input in
+  match b with
+  (* | '\x70' -> (Some FUNCREF, input) *)
+  (* | '\x6F' -> (Some EXTERNREF, input) *)
```
zapashcanon, line 89: They are implemented in Owi so you should be able to find them somewhere in the AST.

src/binary_basic.ml
```diff
@@ -0,0 +1,471 @@
...
+let deserialize_valtype input =
+  match deserialize_numtype input with
+  | Some t, input -> (t, input)
+  | None, _ -> (
+    match deserialize_vectype input with
+    | Some t, input -> (t, input)
+    | None, _ -> (
+      match deserialize_reftype input with
+      | Some t, input -> (t, input)
+      | None, _ -> failwith "deserialize_valtype error" ) )
```
zapashcanon, line 101: It could be possible to get something like:


```ocaml
let deserialize_valtype input =
  match deserialize_numtype input with
  | Ok _ as v -> v
  | Error () ->
    match deserialize_vectype input with
    | Ok _ as v -> v
    | Error () ->
      match deserialize_reftype input with
      | Ok _ as v -> v
      | Error () -> Error "deserialize_valtype error"
```

src/binary_deserializer.ml
```diff
@@ -0,0 +1,359 @@
...
+  | Mem of limits
+  | Global of mut * 'b val_type
+
+let consume_to_end x input = (x, Input.sub ~pos:0 ~len:0 input)
+
+let magic_check bytes =
+  let magic = Bytes.sub_string bytes 0 4 in
+  match String.equal magic "\x00\x61\x73\x6d" with
+  | true -> Ok ()
+  | false -> assert false
```
zapashcanon, line 21:
```ocaml
match e1 with
| true -> e2
| false -> e3
```

is `if e1 then e2 else e3` ;) (same in other places in this file)

src/binary_deserializer.ml
```diff
@@ -0,0 +1,359 @@
...
+         let params = List.map (fun param -> (None, param)) params in
+         (Bt_raw (Some (Raw id), (params, results)), input) ) )
+
+let section_import input =
+  print_endline "import";
+  section_parse input ~expected_id:'\x02' []
+  @@ vector (fun _id input ->
+         let modul, input = vector (fun _id input -> read_byte input) input in
+         let name, input = vector (fun _id input -> read_byte input) input in
+         let modul = String.of_seq (List.to_seq modul) in
```
zapashcanon, line 77: A function `string_of_char_list` would be nice to have defined separately (it can also probably be implemented in a better way ;))

src/binary_deserializer.ml
```diff
@@ -0,0 +1,359 @@
...
+  section_parse input ~expected_id:'\x00' () @@ consume_to_end ()
+
+let section_type input =
+  print_endline "type";
+  section_parse input ~expected_id:'\x01' []
+    (vector (fun id input ->
+         let fcttype, input = read_byte input in
+         assert (fcttype = '\x60');
+         let params, input =
+           vector (fun _id input -> deserialize_valtype input) input
```
zapashcanon, line 63: A function `let vector_no_id f input = vector (fun _id -> f) input` would be nice as it seems we ignore the `id` in most cases. Then this could be rewritten `vector_no_id deserialize_valtype input`

src/binary_deserializer.ml
```diff
@@ -0,0 +1,359 @@
...
+          { exports with global }
+        | _ -> failwith "deserialize_exportdesc error" )
+      empty_exports exports
+  in
+  let modul =
+    { modul with global; mem; elem; func; table; start; data; exports }
+  in
+  Ok modul
+
+let empty_modul () : Simplified.modul =
```
zapashcanon, line 334: Could you define this as a value in `simplified.ml` directly ? `Simplified.empty_modul` for instance. We could probably use it in other parts of Owi. :)

src/compile.ml
```diff
@@ -40,3 +40,14 @@ let until_interpret link_state ~unsafe ~optimize ~name m =
   let* m, link_state = until_link link_state ~unsafe ~optimize ~name m in
   let* () = Interpret.Concrete.modul link_state.envs m in
   Ok link_state
+
+let simplified_interpret link_state ~unsafe ~name m =
+  let* m =
+    if unsafe then Ok m
+    else
+      let* () = Typecheck.modul m in
+      Ok m
```
zapashcanon, line 49:
```ocaml
let+ () = Typecheck.modul m in
m
```

:-)

src/binary_basic.ml
```diff
@@ -0,0 +1,471 @@
+open Types
+
+module Input = struct
+  type t =
+    { bytes : bytes
```
zapashcanon, line 5: Using `string` is probably better here as we never mutate it ? In OCaml `string` is simply `bytes` but made immutable...

src/binary_deserializer.ml
epatrizio: I don't know if this implementation would be better (in terms of complexity)  :
```ocaml
let string_of_char_list char_list =
  (* String.of_seq (List.to_seq char_list) *)
  let char_arr = Array.of_list char_list in
  String.init (Array.length char_arr) (fun i -> Array.get char_arr i)
```

zapashcanon: Something like:

```ocaml
let buf = Buffer.create 64 in
List.iter (Buffer.add_char buf) char_list;
Buffer.contents buf
```

would be better. :)

epatrizio: OCaml Buffer module :ok_hand:

src/binary_deserializer.ml
```diff
@@ -0,0 +1,338 @@
...
+open Binary_basic
+
+type ('a, 'b) import =
+  | Func of int
+  | Table of limits * 'a ref_type
+  | Mem of limits
+  | Global of mut * 'b val_type
+
+let consume_to_end x input =
+  let* input = Input.sub ~pos:0 ~len:0 input in
```
zapashcanon, line 16: `let+`

src/binary_basic.ml
```diff
@@ -0,0 +1,538 @@
...
+end
+
+let string_of_char_list char_list =
+  let buf = Buffer.create 64 in
+  List.iter (Buffer.add_char buf) char_list;
+  Buffer.contents buf
+
+let read_byte input =
+  let* c = Input.get0 input in
+  let* next_input = Input.sub_suffix 1 input in
```
zapashcanon, line 46: `let+`

src/binary_basic.ml
```diff
@@ -0,0 +1,534 @@
...
+  Ok (c, next_input)
+
+(* https://en.wikipedia.org/wiki/LEB128#Unsigned_LEB128 *)
+let rec read_UN n input =
```
krtab, line 50: `n` is not checked to be > 0

src/binary_basic.ml
```diff
@@ -0,0 +1,534 @@
...
+  let* b, input = read_byte input in
+  let b = Char.code b in
+  let x = Int64.of_int (b land 0x7f) in
+  if b land 0x80 = 0 then Ok (x, input)
+  else
+    let* i64, input = read_UN (n - 7) input in
+    Ok (Int64.logor x (Int64.shl i64 7L), input)
```
krtab, line 57: Depending on our perf requirements, we may want to make this tail-rec

src/binary_basic.ml
```diff
@@ -0,0 +1,534 @@
...
+let read_U32 input =
+  let* i64, input = read_UN 32 input in
+  Ok (Int64.to_int i64, input)
+
+let read_U64 input =
+  let* i64, input = read_UN 64 input in
+  Ok (Int64.to_int i64, input)
+
+(* https://en.wikipedia.org/wiki/LEB128#Signed_LEB128 *)
+let rec read_SN n input =
```
krtab, line 68: n should be checked

src/binary_basic_decode.ml
```diff
@@ -0,0 +1,640 @@
...
+    { bytes : string
+    ; pt : int
+    ; size : int
+    }
+
+  let _pp input =
+    print_int input.pt;
+    print_string " ";
+    print_int input.size;
+    print_newline ()
```
krtab, line 15: Could you either:
- base it on Format and remove the underscore,
- or remove it altogether?

src/binary_basic_decode.ml
```diff
@@ -0,0 +1,640 @@
...
+
+  let is_empty input = input.size = 0
+
+  let from_bytes bytes =
```
krtab, line 19: awkward "bytes" when it is actually a string

epatrizio: Keeping `bytes` in some function names seems to make sense: it refers to the binary format of the source file.

src/compile.ml
zapashcanon: Actually, none of them should be redefined and we should simply `open Syntax`, it probably was written a long time ago. See #240 :)

src/binary_basic_decode.ml
```diff
@@ -0,0 +1,639 @@
...
+    else
+      let* acc_elt, input = parse_elt loop_id input in
+      let acc = acc_elt :: acc in
+      loop (loop_id + 1) input acc
+  in
+  loop 0 input []
+
+let vector_no_id f input = vector (fun _id -> f) input
+
+let deserialize_indice input =
```
zapashcanon, line 139: If you make the type more precise by using something along the lines of:

```ocaml
let deserialize_indice input : Types.simplified Types.indice =
  ...
```

Then it should simplify some other parts of the code such as:

```ocaml
let tableidx =
  match tableidx with Raw i -> i | _ -> assert false
```

Because the typechecker will understand that the other case is impossible so you'll be able to simply write:

```ocaml
let tableidx =
  match tableidx with Raw i -> i
```

Or even:

```ocaml
let (Raw tableidx) = tableidx in
```

epatrizio: Complete type with `result`
```ocaml
(Types.simplified Types.indice * Input.t, Error.err) result
```

src/binary_basic.ml
epatrizio: Add TODO

zapashcanon:
Thanks!

PR #227, opened 2024-03-21
Add a CI check that owi.opam has been correctly generated

krtab:


krtab:
You can see it fail successfully here: https://github.com/OCamlPro/owi/actions/runs/8373647726/job/22927156227#step:11:30

PR #233, opened 2024-04-17
Little project cleanup

lefessan:
* Fix source file headers: Copyright to OCamlPro, Written by Leo and Pierre
* Add a basic Makefile
* Fix some missing dependencies in dune-project


.gitignore
```diff
@@ -1,3 +1,7 @@
-_build
-_coverage
+_*
 owi-out
+*~
+headers-cc.txt
+headers-ml.txt
+.header-check-more
+check-headers.undo
```
krtab, line 7: Which files were ignored but generated by our workflow? If they are specific to your own tooling, wouldn't they be better suited to be in .git/info/exclude?

dune-project
```diff
@@ -74,5 +74,10 @@
   pyml
   re2
   (hc (>= 0.3))
+  crunch
+  mdx
+  crowbar
+  tiny_httpd
+  graphics
```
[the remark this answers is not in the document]
lefessan: So, could you give me the `Makefile` rules to use to call `dune` to only build test targets, and separately, dev targets ?

.gitignore
lefessan: What do you mean by **our workflow** ? Isn't **my workflow** part of **our workflow** ?

Also, do you really care about any **other** contributor adding a few lines in `.gitignore` ? In general, it's a small concession, but having more contributors to an open-source tool is usually worth the trouble...

krtab: > What do you mean by our workflow ? Isn't my workflow part of our workflow ?

Taking "our workflow' to mean the one implemented in this repository, I don't think yours is part of ours, even after merging this PR.

> Also, do you really care about any other contributor adding a few lines in .gitignore

Yes, most projects I know of try to keep their .gitignore light, I believe it is good practice to do the same.

That being said, I don't think this discussion will turn into anything more than bikeshedding, so let's just say that without more justification for these additions, I am against, and see what others think.

lefessan:
I won't spend more time on this. However, the headers and README **must** be updated with correct copyright information. I let you decide whether to merge this request in full, or create another one with only some parts of it.

zapashcanon:
See #234 as an alternative. Thanks!

PR #235, opened 2024-04-18
Accommodate `encoding` library name change

filipeom:
It was decided that the `encoding` library would now be named `smtml`. Therefore, in this PR, I make the necessary adjustments to accommodate this change.

Additionally, I took advantage of this opportunity to use more smart constructors during symbolic expression creation, rather than building these expressions manually. This allows for greater expression optimization, due to the simplification applied at construction time.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -60,7 +60,7 @@
   (odoc :with-doc)
   (crunch :dev)
   bos
-  encoding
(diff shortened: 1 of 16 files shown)


zapashcanon:
Thanks a lot! :)

PR #238, opened 2024-04-22
Concolic execution

chambart:
I could have been productively writing some part of the article, instead I procrastinated to make that.

This is quite early, but maybe not that far from the real deal: It only run once. But it is too late to continue that this 'evening'.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,10 +3,10 @@
 - `owi opt` and `owi sym` can run on `.wasm` files directly
 - remove dependency on `wabt`
 - better API for `Parse`, `Compile` and `Simplified` (renamed to `Binary`), added a `Binary_to_tex` module
+- add `owi conc` subcommands and `owi c --concolic`: concolic mode
 
 ## 0.2 - 2024-04-24
 
-- add `owi conc` subcommands and `owi c --concolic`: concolic mode
 - use a subcommand system for the `owi` binary
 - add `owi c`, `owi fmt`, `owi opt`, `owi sym`, `owi validate` and `owi wasm2wat` subcommands
 - add a fuzzer
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 ## 0.2 - 2024-04-24
 
+- add `owi conc` subcommands and `owi c --concolic`: concolic mode
 - use a subcommand system for the `owi` binary
 - add `owi c`, `owi fmt`, `owi opt`, `owi sym`, `owi validate` and `owi wasm2wat` subcommands
 - add a fuzzer
--- a/src/bin/owi.ml
+++ b/src/bin/owi.ml
@@ -115,11 +115,15 @@ let c_cmd =
     let doc = "write results to dir" in
     Arg.(value & opt string "owi-out" & info [ "output"; "o" ] ~doc)
   in
+  let concolic =
+    let doc = "concolic mode" in
+    Arg.(value & flag & info [ "concolic" ] ~doc)
+  in
   Cmd.v info
     Term.(
       const Cmd_c.cmd $ debug $ arch $ property $ testcomp $ output $ workers
       $ opt_lvl $ includes $ files $ profiling $ unsafe $ optimize
-      $ no_stop_at_failure $ no_values $ deterministic_result_order )
+      $ no_stop_at_failure $ no_values $ deterministic_result_order $ concolic )
 
 let fmt_cmd =
   let open Cmdliner in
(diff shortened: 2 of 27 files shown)


src/bin/owi.ml
```diff
@@ -183,6 +183,19 @@ let sym_cmd =
       $ no_stop_at_failure $ no_values $ deterministic_result_order $ workspace
       $ files )
 
+let conc_cmd =
```
zapashcanon, line 186: I'm not sure what's best here, having `owi sym` and `owi conc`, which mean we'll have to have `owi c` and `owi concc` for instance (and so on for every future language). Or make this an option of the symbolic command, which will give `owi sym --concolic` and `owi c --concolic`. Do you have any opinion about it?

src/cmd/cmd_conc.ml
```diff
@@ -0,0 +1,240 @@
+open Syntax
+module Expr = Smtml.Expr
+module Value = Concolic_value.V
+module Choice = Concolic.P.Choice
+
+let () = Random.self_init ()
```
zapashcanon, line 6: We should maybe store this in some kind of cache file, in order to be able to get some deterministic results when needed?

src/cmd/cmd_conc.ml
```diff
@@ -0,0 +1,240 @@
...
+      (*   , Symbolic.P.Extern_func.Extern_func (Func (Arg (I32, Res), R0), free) ) *)
+      (* ; ("abort", Symbolic.P.Extern_func.Extern_func (Func (UArg Res, R0), abort)) *) ]
+  in
+  { functions }
+
+let ( let*/ ) (t : 'a Result.t) (f : 'a -> 'b Result.t Choice.t) :
+  'b Result.t Choice.t =
+  match t with Error e -> Choice.return (Error e) | Ok x -> f x
+
+let simplify_then_link_then_run ~unsafe ~optimize (pc : unit Result.t Choice.t)
```
zapashcanon, line 116: I feel like most of this code could be shared with `cmd_sym.ml` right?

src/cmd/cmd_conc.ml
```diff
@@ -0,0 +1,240 @@
...
+       Or we can fix this by sharing the result of Concolic.convert_env
+    *)
+    Env_id.map Concolic.convert_env link_state.envs
+  in
+  let c = (Interpret.Concolic.modul link_state_envs) to_run in
+  Choice.bind pc (fun r ->
+      match r with Error _ -> Choice.return r | Ok () -> c )
+
+let run_file ~unsafe ~optimize pc filename =
+  let*/ m0dule = Parse.Module.from_file filename in
```
zapashcanon, line 158: I believe we took the `modul` convention (although I have to admit `m0dule` is funny).

src/concolic/concolic_value.ml
```diff
@@ -0,0 +1,462 @@
+module type T = sig
+  type t
+end
+
+type ('c, 's) cs =
```
zapashcanon, line 5: Could you find a better name ? :sweat_smile: When reading `concolic_choice` it was hard for me to understand.

src/bin/owi.ml
zapashcanon: @krtab maybe you have ideas about this

src/cmd/cmd_conc.ml
filipeom: Yeah, probably at some point we could use a default seed, in my experience it makes debugging easier. 

For now, since it only runs 1 path its good because we can see different paths with different executions :smiley:

chambart: Yes I fixed the default seed. That was mainly to be able to test before having some proper runner

src/concolic/concolic_value.ml
chambart: Yes, certainly. I'm not sure it is really better than a pair anyway.

chambart:
I pushed a few more things, now it has some kind of runner that can try to find some branch to run somewhat randomly (This is very naive and very bad, but it has to start somewhere...).

src/bin/owi.ml
zapashcanon: OK, keeping `owi conc` and having `owi c --concolic` LGTM. :)

CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## 0.2 - 2024-04-24
 
+- add `owi conc` subcommands and `owi c --concolic`: concolic mode
```
zapashcanon, line 5: This should go in the `unreleased` section, 0.2 has already been published on opam :)

src/concolic/concolic_value.ml
chambart: I changed the name of the fields

chambart:
We should be able to test it now. I'm removing the draft status

PR #244, opened 2024-04-24
Add a wasm2wat subcommand

zapashcanon:


src/cmd/cmd_wasm2wat.ml
```diff
@@ -0,0 +1,15 @@
...
+
+open Syntax
+
+let cmd_one file =
+  let ext = Fpath.get_ext file in
+  match ext with
+  | ".wasm" ->
+    let* m = Binary_deserializer.from_file file in
+    Ok (Format.pp_std "%a@\n" Simplified.Pp.modul m)
+  | ext -> Error (`Msg (Format.sprintf "invalid extension: `%s`" ext))
```
krtab, line 13: Debatable we want to enforce extension but not blocking for merge.
We could instead check the magic bytes.

test/fmt/locals.t
krtab, line 1: What's happening here?

src/binary_parser/binary_deserializer.ml
krtab, line 1: Assuming this is simply the merge of two files that's ok for me but I didn't review more

zapashcanon: I merged the two files and then the compiler told me there was some unused stuff that I removed.

src/cmd/cmd_wasm2wat.ml
zapashcanon: Yes, I'm not sure yet what's best. I guess it's better to see how we handle this in other places (parse.ml, compile.ml, owi sym, owi run, owi opt etc.) later and do the same everywhere

test/fmt/locals.t
zapashcanon: I realized these tests were not calling the right command. They were calling `owi opt` instead of `owi fmt` so I fixed them. I noticed because I reused all the `fmt` tests in the `wasm2wat` ones. :-)

PR #251, opened 2024-04-27
enable binary tests and fix parts of the binary parser

zapashcanon:


test/script/reference.t
```diff
@@ -1,7 +1,11 @@
   $ owi script --no-exhaustion reference/address.wast
   $ owi script --no-exhaustion reference/align.wast
   $ owi script --no-exhaustion reference/binary-leb128.wast
+  expected integer representation too long but there was no error
+  [7]
```
zapashcanon, line 5: This should be fixed in the binary parser.

test/script/reference.t
```diff
@@ -1,7 +1,11 @@
...
   $ owi script --no-exhaustion reference/binary.wast
+  expected integer too large but there was no error
+  [7]
```
zapashcanon, line 8: This should be fixed in the binary parser.

test/script/reference.t
```diff
@@ -13,7 +17,11 @@
   $ owi script --no-exhaustion reference/const.wast
   $ owi script --no-exhaustion reference/conversions.wast
   $ owi script --no-exhaustion reference/custom.wast
+  expected unexpected end but there was no error
+  [7]
   $ owi script --no-exhaustion reference/data.wast
+  expected unknown memory 0 but there was no error
+  [7]
```
zapashcanon, line 24: This is because we don't validate simplified modules (see the comment in `script.ml`), as validation is currently done while doing the text to simplified transformation. We should split them in the future.

test/script/reference.t
```diff
@@ -107,6 +115,12 @@
...
   $ owi script --no-exhaustion reference/unwind.wast
   $ owi script --no-exhaustion reference/utf8-custom-section-id.wast
+  expected malformed UTF-8 encoding but there was no error
+  [7]
   $ owi script --no-exhaustion reference/utf8-import-field.wast
+  expected malformed UTF-8 encoding but there was no error
+  [7]
   $ owi script --no-exhaustion reference/utf8-import-module.wast
+  expected malformed UTF-8 encoding but there was no error
+  [7]
```
zapashcanon, line 125: This should be fixed in the binary parser.

test/script/reference.t
zapashcanon: This is because we ignore custom sections instead of validating them before ignoring them. I gave a minimal example to reproduce as a comment in `binary_parser.ml`.

zapashcanon:
It seems to be working well now. A few error messages are not those expected by the spec but this can be fixed later, see #255. I'm going to merge in order to be able to make progress on other PR after rebasing on this one.

Issue #258, opened 2024-05-06
wasm-ld: error: duplicate symbol: read


zapashcanon:
on some c files, for instance:

```shell-session
$ owi c
owi c sv-benchmarks/c/busybox-1.22.0/ls-incomplete-2.i
owi-out/ls-incomplete-2.c:2973:14: warning: array index 15 is past the end of the array (that has type 'char[2]') [-Warray-bounds]
      return "\001"[(signed long int)(61440 >> 12 & 15)];
             ^      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
owi-out/ls-incomplete-2.c:3412:14: warning: array index 15 is past the end of the array (that has type 'char[9]') [-Warray-bounds]
      return "\037##%\"%##"[(signed long int)(61440 >> 12 & 15)];
             ^              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
owi-out/ls-incomplete-2.c:5251:1: warning: non-void function does not return a value in all control paths [-Wreturn-type]
}
^
3 warnings generated.
wasm-ld: error: duplicate symbol: read
>>> defined in owi-out/ls-incomplete-2.o
>>> defined in [...]/owi/binc/libc.wasm
run ['/usr/bin/wasm-ld' '-z' 'stack-size=8388608' '--export=_start' '-o'
     'owi-out/a.out.wasm' 'owi-out/ls-incomplete-2.o'
     '[...]/owi/binc/libc.wasm']: exited with 1%                                
```

PR #261, opened 2024-05-06
Make C stdlib read weak

krtab:
TestComp's c/busybox-1.22.0/ test files redefine C stdlib's read which lead to wasm-ld failling with a "duplicate symbol" error. Making the symbol weak resolves the issue.

Closes https://github.com/OCamlPro/owi/issues/258


zapashcanon:
Thanks!

PR #263, opened 2024-05-06
- provide exit function in summaries module

### LINKED ISSUE #256: unknown import "summaries" "exit" with `owi c sv-benchmarks/c/eca-rers2012/Problem06_label00.c`

zapashcanon:
- do not fail when the memory model can not handle some cases but abort instead, so that we keep exploring other paths

fix #256


zapashcanon:
I'm getting better results on test-comp with this patch. Less `Bad` results and more `Nothing` (which correspond to the cases where we are aborting). @filipeom said it seems like the right thing to do so I'm merging.

PR #266, opened 2024-05-07
allows `owi conc` to run on `.wasm` files

zapashcanon:


zapashcanon:
Worked well on test-comp so I'm merging.

PR #283, opened 2024-05-26
better report for benchmarks

zapashcanon:


zapashcanon:
[Preview](https://fs.zapashcanon.fr/tmp/results-report/) on a small local run with a 3 seconds timeout.

PR #295, opened 2024-06-05
A handful of tentative performance improvements

krtab:
@zapashcanon let me know if you want them in separate PRs

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -74,5 +74,6 @@
   digestif
   xmlm
   (hc (>= 0.3))
+  (processor (>= 0.1))
   dune-site)
  (sites (share pyc) (share binc) (share libc)))
--- a/src/bin/owi.ml
+++ b/src/bin/owi.ml
@@ -60,12 +60,12 @@ let unsafe =
 
 let workers =
   let doc =
-    "number of workers for symbolic execution. Defaults to a machine-specific \
-     value given by the OCaml Domain.recommended_domain_count function."
+    "number of workers for symbolic execution. Defaults to the number of \
+     physical cores."
   in
   Cmdliner.Arg.(
     value
-    & opt int (Domain.recommended_domain_count ())
+    & opt int Processor.Query.core_count
     & info [ "workers"; "w" ] ~doc ~absent:"n" )
 
 let workspace =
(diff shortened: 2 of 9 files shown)


src/symbolic/symbolic_choice.mli
```diff
@@ -13,4 +13,11 @@ include
...
      and module V := Symbolic_value
 
-val run : workers:int -> 'a t -> Thread.t -> 'a run_result
+val run :
+     workers:int
+  -> 'a t
+  -> Thread.t
+  -> ('b -> unit)
+  -> (unit -> unit)
+  -> (unit -> unit)
```
zapashcanon, line 22: Could you use named parameters for the three parameters ?

src/cmd/cmd_sym.ml
```diff
@@ -244,7 +244,21 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
   let pc = Choice.return (Ok ()) in
   let result = List.fold_left (run_file ~unsafe ~optimize) pc files in
   let thread : Thread.t = Thread.create () in
-  let results = Symbolic_choice.run ~workers result thread in
+  let res_queue = Wq.init () in
+  let callback v =
+    match fst v with
+    | Symbolic_choice.EVal (Ok ()) -> ()
+    | _ -> Wq.push v res_queue
```
zapashcanon, line 251:
```ocaml
let callback = function
  | Symbolic_choice.EVal (Ok ()), _ -> ()
  | _ -> Wq.push v res_queue
```

dune-project
```diff
@@ -74,5 +74,6 @@
   digestif
   xmlm
   (hc (>= 0.3))
+  (processor (= 0.1))
```
zapashcanon, line 77: `>= 0.1`

src/bin/owi.ml
```diff
@@ -65,7 +65,7 @@ let workers =
   in
```
zapashcanon, line 65: Could you update the doc to say we use the total number of physical cores and not the value provided by Domain.recommended_domain_count anymore ?

src/symbolic/symbolic_choice.ml
```diff
@@ -366,7 +366,7 @@ let get_model_or_stop symbol =
     | Some v -> return v
   end
 
-let select (cond : Symbolic_value.vbool) =
+let select_inner ?(explore_first = true) (cond : Symbolic_value.vbool) =
```
zapashcanon, line 369: Could you use `~explore_first` instead ? I don't really like optional arguments TBH, and here the function is used in only two places so it's not that tedious

zapashcanon:
LGTM once we checked that the 1st and 3rd one are actually leading to better results, which I would expect (+ my nitpicks comments).

zapashcanon:
@krtab, I wanted to run the benchmarks on the server, but the code actually does not typecheck

krtab:
Split in #320 #321 #322 #323

PR #297, opened 2024-06-05
Concolic: Don't try to explore branches with nothing to explore

chambart:
Avoid randomly picking a branch that has nothing to run


zapashcanon:
Included in #296.

PR #299, opened 2024-06-06
add `--solver` cli option

zapashcanon:
requires dev version of smtml


filipeom:
This should now be available at `smtml 0.1.1`. However, to use colibri2 we might need to change the way we're interacting with the solver when we generate models.

PR #301, opened 2024-06-07
implement the extended const-proposal

Laplace-Demon:


zapashcanon:
The CI is failing for an unrelated error (we probably broke a concolic execution test with the recent commits but it doesn't matter as this is quite experimental), I'm merging.

zapashcanon:
Thanks!

PR #310, opened 2024-06-14
Simplify file

Laplace-Demon:


src/libc/src/test-comp.c
```diff
@@ -2,7 +2,7 @@
 
 _Bool __VERIFIER_nondet_bool(void) {
   _Bool var = owi_i32();
-  owi_assume(or_(var == 0, var == 1));
+  owi_assume(var == 0 || var == 1);
```
zapashcanon, line 5: @filipeom, I thought we were not using `or_` and `and_` anymore but actually we are using it there, do you think this is going to have a negative impact in some way to change it like this ? Maybe, we could also expose a `owi_bool()` function instead, what do you think ?

filipeom: It will depend on how the compiler decides to compile this. But most likely it will try to lazy evaluate the condition, possibly causing unnecessary branches. Creating  a `owi_bool` is cleaner. But, using a bitwise-or or chaining two `owi_assume` would most-likely also work to reduce branching also.

zapashcanon: @Laplace-Demon, do you want to try to add a `owi_bool` function ?

src/concolic/concolic_wasm_ffi.ml
```diff
@@ -69,6 +69,19 @@ module M :
...
+        let n =
+          match forced_value with
+          | None -> Int32.logand 1l (Random.bits32 ())
+          | Some (Num (I32 n)) -> n
+          | _ -> assert false
+        in
+        let sym_expr =
+          Expr.make (Cvtop (Ty_bitv 32, Zero_extend 31, Expr.mk_symbol sym))
+        in
+        (I32 n, Value.pair n sym_expr) )
```
zapashcanon, line 83: @filipeom what do you think about this ?

src/symbolic/symbolic_wasm_ffi.ml
```diff
@@ -48,6 +48,8 @@ module M :
 
   let symbol_f64 = symbol (Ty_fp 64)
 
+  let symbol_bool = symbol (Ty_bitv 32)
+
```
zapashcanon, line 52: @Laplace-Demon, here I believe you have to constrain the symbol further otherwise it can be any `int32`

src/concolic/concolic_wasm_ffi.ml
filipeom: This won't work because of the: 

```ocaml
Choice.with_new_symbol (Ty_bitv 32) (fun sym forced_value ->
```

At the beginning of the function, so It will create a 63-bit vector. For this to be correct we would need to create a `Ty_bitv 1` symbol. But I don't think it's supported atm :sweat_smile: 

We can probably get away with just `let sym expr = Expr.mk_symbol sym in`

src/concolic/concolic_wasm_ffi.ml
```diff
@@ -69,6 +69,19 @@ module M :
...
+  let symbol_bool () : Value.int32 Choice.t =
+    Choice.with_new_symbol (Ty_bitv 32) (fun sym forced_value ->
+        let n =
+          match forced_value with
+          | None -> Int32.logand 1l (Random.bits32 ())
+          | Some (Num (I32 n)) -> n
+          | _ -> assert false
+        in
+        let sym_expr =
+          Expr.make (Cvtop (Ty_bitv 32, Zero_extend 31, Expr.mk_symbol sym))
```
filipeom, line 81:
```suggestion
    Choice.with_new_symbol (Ty_bool) (fun sym forced_value ->
        let b =
          match forced_value with
          | None -> Random.bool ()
          | Some True -> true
          | Some False -> false
          | _ -> assert false
        in
        let n = V.Bool.int32 b in
        (I32 n, Value.(Bool.int32 (pair b (Expr.mk_symbol sym)))))
```
You could probably try to do something similar to what you did on the symbolic version here. (I don't know if this code compiles)

Laplace-Demon: Thanks for the proposition, it compiles and works :)

src/libc/src/owi.c
```diff
@@ -1,58 +1,27 @@
...
-                       "i32.const 0;"
-                       "i32.ne;"
-                       "local.get 1;"
-                       "i32.const 0;"
-                       "i32.ne;"
-                       "i32.or;"
-                       "return;");
-}
+__attribute__((weak, import_module("symbolic"), import_name("assume"))) void
+assume(int);
```
zapashcanon, line 27: missing final line here

zapashcanon: also `assume` is present two times

zapashcanon: @krtab could you remind me why was the `weak` attribute needed ?

zapashcanon: @Laplace-Demon also all attributes can be safely removed from the `.h` files (but should be kept in the `.c` file).

zapashcanon:
@Laplace-Demon oops, I meant, removed from the `.c` file but kept in the header, sorry !

Laplace-Demon:
> @Laplace-Demon oops, I meant, removed from the `.c` file but kept in the header, sorry !

dw, it's fixed

src/symbolic/symbolic_wasm_ffi.ml
```diff
@@ -48,6 +48,11 @@ module M :
 
   let symbol_f64 = symbol (Ty_fp 64)
 
+  let symbol_bool () =
+    let open Choice in
+    let+ sym = symbol Ty_bool () in
+    Expr.make (Cvtop (Ty_bool, Smtml.Ty.OfBool, sym))
```
zapashcanon, line 54: You'll need the following patch to rebase:

```patch
diff --git a/src/symbolic/symbolic_wasm_ffi.ml b/src/symbolic/symbolic_wasm_ffi.ml
index 3f4ed0b2..4ad1ba33 100644
--- a/src/symbolic/symbolic_wasm_ffi.ml
+++ b/src/symbolic/symbolic_wasm_ffi.ml
@@ -40,9 +40,7 @@ module M :
   let symbol_f64 () = Choice.with_new_symbol (Ty_fp 64) Expr.symbol
 
   let symbol_bool () =
-    let open Choice in
-    let+ sym = symbol Ty_bool () in
-    Expr.make (Cvtop (Ty_bool, Smtml.Ty.OfBool, sym))
+    Choice.with_new_symbol Ty_bool Expr.symbol
 
   open Expr
 ```

**Comment by zapashcanon** [comment]
I started a benchmark to check if the boolean stuff makes any difference, it's in `results-testcomp-owi_w24_O3_sZ3-2024-07-16_15h37m05s/`

**Comment by zapashcanon** [comment]
@filipeom, I'm getting a lot of:

```txt
     Z3.Error("Sorts Bool and (_ BitVec 32) are incompatible")
```

Is my suggested patch incorrect ? :sweat_smile:

filipeom:
> @filipeom, I'm getting a lot of:
> 
> ```
> owi: internal error, uncaught exception:
>      Z3.Error("Sorts Bool and (_ BitVec 32) are incompatible")
> ```
> 
> Is my suggested patch incorrect ? 😅

It's most likely because of `to_bool`:

https://github.com/OCamlPro/owi/blob/10b939f9d4681de4d5cc6c801c356e9b2ec78494/src/symbolic/symbolic_value.ml#L192-L196

`ToBool` is assuming its being passed something of type `I32`. I can suggest a workaround:

```diff
diff --git a/src/symbolic/symbolic_value.ml b/src/symbolic/symbolic_value.ml
index d32f69fd..ed50be17 100644
--- a/src/symbolic/symbolic_value.ml
+++ b/src/symbolic/symbolic_value.ml
@@ -192,6 +192,7 @@ module I32 = struct
   let to_bool (e : vbool) =
     match view e with
     | Val (Num (I32 i)) -> Bool.const @@ Int32.ne i 0l
+    | Symbol { ty = Ty_bool; _ } -> e
     | Cvtop (_, OfBool, cond) -> cond
     | _ -> make (Cvtop (ty, ToBool, e))
```

But, I'm not sure this is enough. I need to play around with this and most likely fix it directly in `smtml.

zapashcanon:
OK, I'll wait then. If you want to find a reproduction case, you can look on the server with the results mentioned earlier. Then look at which tests are causing `Other` and open the output of one of them to check it has the same error message.

zapashcanon:
@Laplace-Demon, actually, maybe we could do a first PR without the `bool` related changes and make a second one later, WDYT?

filipeom:
> OK, I'll wait then. If you want to find a reproduction case, you can look on the server with the results mentioned earlier. Then look at which tests are causing `Other` and open the output of one of them to check it has the same error message.

Yes, I'm checking it out :smile:

filipeom:
Ok so instead of this https://github.com/OCamlPro/owi/pull/310#discussion_r1679649107

We should use the following patch instead:

```diff
diff --git a/src/cmd/cmd_utils.ml b/src/cmd/cmd_utils.ml
index be7ac19d..e9fb3d1d 100644
--- a/src/cmd/cmd_utils.ml
+++ b/src/cmd/cmd_utils.ml
@@ -8,7 +8,7 @@ let out_testcase ~dst testcase =
   let o = Xmlm.make_output ~nl:true ~indent:(Some 2) dst in
   let tag atts name = (("", name), atts) in
   let atts = [ (("", "coversError"), "true") ] in
-  let to_string v = Format.asprintf "%a" Smtml.Value.pp_num v in
+  let to_string v = Format.asprintf "%a" Smtml.Value.pp v in
   let input v = `El (tag [] "input", [ `Data (to_string v) ]) in
   let testcase = `El (tag atts "testcase", List.map input testcase) in
   let dtd =
diff --git a/src/symbolic/symbolic_wasm_ffi.ml b/src/symbolic/symbolic_wasm_ffi.ml
index 3f4ed0b2..8eaee478 100644
--- a/src/symbolic/symbolic_wasm_ffi.ml
+++ b/src/symbolic/symbolic_wasm_ffi.ml
@@ -40,9 +40,8 @@ module M :
   let symbol_f64 () = Choice.with_new_symbol (Ty_fp 64) Expr.symbol
 
   let symbol_bool () =
-    let open Choice in
-    let+ sym = symbol Ty_bool () in
-    Expr.make (Cvtop (Ty_bool, Smtml.Ty.OfBool, sym))
+    Choice.with_new_symbol (Ty_bitv 1) (fun sym ->
+        Expr.cvtop (Ty_bitv 32) (Smtml.Ty.Zero_extend 31) (Expr.symbol sym) )
 
   open Expr

```

And the latest version of https://github.com/formalsec/smtml/pull/179

Laplace-Demon:
> @Laplace-Demon, actually, maybe we could do a first PR without the `bool` related changes and make a second one later, WDYT?

Yes you have raison :)

src/libc/src/owi.c
```diff
@@ -1 +1,23 @@
...
+int and_(int a, int b) {
+  __asm__ __volatile__("local.get 0;"
+                       "i32.const 0;"
+                       "i32.ne;"
+                       "local.get 1;"
+                       "i32.const 0;"
+                       "i32.ne;"
+                       "i32.and;"
+                       "return;");
+}
```
zapashcanon, line 12: I think you can still remove `and_` and only keep `or_`. :)

zapashcanon:
Thanks!

Issue #312, opened 2024-06-14
concolic failures


zapashcanon:
From the report generated:

```
    run 1151: testcomp/sv-benchmarks/c/seq-pthread/cs_stateful-1.i
    run 1150: testcomp/sv-benchmarks/c/seq-pthread/cs_stack-2.i
    run 1149: testcomp/sv-benchmarks/c/seq-pthread/cs_read_write_lock-2.i
    run 1148: testcomp/sv-benchmarks/c/seq-pthread/cs_queue-2.i
    run 1147: testcomp/sv-benchmarks/c/seq-pthread/cs_fib_longer-2.i
    run 1146: testcomp/sv-benchmarks/c/seq-pthread/cs_fib-2.i
    run 853: testcomp/sv-benchmarks/c/ntdrivers/parport.i.cil-1.c
    run 852: testcomp/sv-benchmarks/c/ntdrivers/kbfiltr.i.cil-2.c
    run 851: testcomp/sv-benchmarks/c/ntdrivers/floppy.i.cil-1.c
    run 850: testcomp/sv-benchmarks/c/ntdrivers/diskperf.i.cil-2.c
    run 849: testcomp/sv-benchmarks/c/ntdrivers/cdaudio.i.cil-1.c
    run 116: testcomp/sv-benchmarks/c/eca-rers2012/Problem18_label00.c
    run 115: testcomp/sv-benchmarks/c/eca-rers2012/Problem16_label00.c
    run 114: testcomp/sv-benchmarks/c/eca-rers2012/Problem15_label00.c
    run 113: testcomp/sv-benchmarks/c/eca-rers2012/Problem12_label00.c
    run 112: testcomp/sv-benchmarks/c/eca-rers2012/Problem11_label00.c
    run 111: testcomp/sv-benchmarks/c/eca-rers2012/Problem06_label00.c
    run 110: testcomp/sv-benchmarks/c/eca-rers2012/Problem05_label00.c
    run 107: testcomp/sv-benchmarks/c/busybox-1.22.0/test-1.i
    run 106: testcomp/sv-benchmarks/c/busybox-1.22.0/printf-3.i
    run 105: testcomp/sv-benchmarks/c/busybox-1.22.0/od-4.i
    run 104: testcomp/sv-benchmarks/c/busybox-1.22.0/ls-incomplete-2.i
```

I had a quick look at the output:

[...]

PR #314, opened 2024-06-15
Implement concolic tables

chambart:
Should fix a few `TODO` from https://github.com/OCamlPro/owi/issues/312


zapashcanon:
Thanks!

PR #318, opened 2024-06-17
Add solver argument to testcomp script

filipeom:
Maybe `solver : Smtml.Solver_dispatcher.solver_type` should be optional?


zapashcanon:
Thanks!

PR #320, opened 2024-06-18
Allow passing a filter callback to the scheduler

krtab:


src/symbolic/symbolic_choice.ml
```diff
@@ -161,19 +158,21 @@ end = struct
...
-              Wq.end_pledge sched.res_writer;
-              Wq.fail sched.work_queue )
+            ~finally:(fun () -> callback_close ())
             (fun () ->
               let wls = wls_init () in
-              work wls sched ) )
+              try work wls sched callback
+              with e ->
+                let bt = Printexc.get_raw_backtrace () in
+                Wq.fail sched.work_queue;
```
zapashcanon, line 174: You can put this in the `~finally` and remove the `try ... with` along with the exeception backtrace stuff

zapashcanon:
I'm going to bench this on Test-Comp first

src/symbolic/symbolic_choice.ml
```diff
@@ -298,17 +297,16 @@ end = struct
...
-      Array.map
-        (fun () -> spawn_worker sched (Solver.fresh solver))
-        (Array.init workers (Fun.const ()))
+      Array.init workers (fun _i ->
+          spawn_worker sched (Solver.fresh solver) callback callback_init
+            callback_end )
     in
-    Wq.read_as_seq sched.res_writer ~finalizer:(fun () ->
-        Array.iter Domain.join join_handles )
+    join_handles
```
zapashcanon, line 309: You can directly return the array without binding it:

```suggestion
    Array.init workers (fun _i ->
        spawn_worker sched (Solver.fresh solver) callback callback_init
          callback_end )
```

zapashcanon:
I started the bench. For reference the current score on Test-Comp for the main branch is:

```yaml
  Nothing:     75    Reached:    613    Timeout:    527    Other:      1    Killed:      0
```

src/symbolic/symbolic_choice.ml
zapashcanon, line 174: Actually, it seems that I introduced a bug by putting the `Wq.fail` intro the `~finally` as it clearly does not have the same behavior...

src/symbolic/symbolic_choice.ml
```diff
@@ -132,25 +135,19 @@ end = struct
...
 
     let add_init_task sched task = Wq.push task sched.work_queue
 
-    let rec work wls sched =
+    let rec work wls sched callback =
       let rec handle_status (t : _ Schedulable.status) sched =
         match t with
         | Stop -> ()
-        | Now x -> Wq.push x sched.res_writer
+        | Now x -> callback x
```
zapashcanon, line 150: Is there actually any value in having all the callback mechanism instead of simply matching on `x` here to push into the queue only when it is not `EVal (Ok ())` ?

krtab: Yes, because we can't access the type here

zapashcanon:
Thanks!

PR #321, opened 2024-06-18
Use physical cores for recommended parralellism

krtab:
Should this have a changelog?


zapashcanon:
I think no Changelog is needed. Thanks!

PR #324, opened 2024-06-19
Integration of E-ACSL specification language in owi

Laplace-Demon:


src/cmd/cmd_c.ml
```diff
@@ -2,6 +2,8 @@
 (* Copyright © 2021-2024 OCamlPro *)
 (* Written by the Owi programmers *)
 
+[@@@warnerror "-unused-var-strict"]
```
zapashcanon: Can you remove this ? :-)

src/cmd/cmd_c.ml
```diff
@@ -35,12 +37,51 @@ let find location file : Fpath.t Result.t =
...
+    let outs =
+      List.map
+        (fun file ->
+          let ext = Fpath.get_ext ~multi:true file in
+          let file_without_ext = Fpath.rem_ext ~multi:true file in
+          let file_without_ext =
+            Fpath.to_string file_without_ext ^ "_instrumented"
+          in
+          let file = file_without_ext ^ ext in
+          Fpath.v file )
```
zapashcanon, line 66: Could you build a list of pairs with the old name and the new one, so you can avoid the `list_iter2` stuff ?

src/libc/include/eacsl.h
```diff
@@ -0,0 +1,277 @@
...
+void __e_acsl_assert_register_union(__e_acsl_assert_data_t *data, const char *name);
+void __e_acsl_assert_register_other(__e_acsl_assert_data_t *data, const char *name);
+
+/* Miscellaneous functions */
+void __e_acsl_assert_copy_values(__e_acsl_assert_data_t *dest,
+                              __e_acsl_assert_data_t *src);
+void __e_acsl_assert_clean(__e_acsl_assert_data_t *data);
+
+
+#endif
```
zapashcanon, line 277: missing newline

test/c/eacsl/integer_dividable.c
```diff
@@ -0,0 +1,22 @@
...
+
+int main(void) {
+  int numerator = owi_i32();
+  int denominator = owi_i32();
+
+  owi_assume(denominator != 0);
+
+  is_dividable(numerator, denominator);
+  return 0;
+}
```
zapashcanon, line 22: missing EOL

test/c/eacsl/integer_dividable.t
```diff
@@ -0,0 +1,11 @@
+  $ owi c --e-acsl integer_dividable.c --no-value
+  [kernel] Parsing integer_dividable.c (with preprocessing)
+  [e-acsl] beginning translation.
+  [kernel] Parsing FRAMAC_SHARE/e-acsl/e_acsl.h (with preprocessing)
+  [e-acsl] translation done in project "e-acsl".
```
zapashcanon, line 5: Could you pass a flag to frama-c when asking for the preprocessing file so that it stops writing this kind of verbose output ? (Even better would be to pass the flag only when the `--debug` one is not present in Owi.)

test/c/eacsl/max.c
```diff
@@ -0,0 +1,22 @@
...
+int max(int x, int y) {
+    return (x > y) ? x : y;
+}
+
+int main(void) {
+    int x = owi_i32();
+    int y = owi_i32();
+    max(x, y);
+    return 0;
+}
```
zapashcanon, line 22: missing EOL (you should configure your editor to add it automatically on save)

test/c/eacsl/integer_dividable.t
Laplace-Demon: now the presence and absence of --debug will set the frama-c verbosity level to 2 and 0, which by default is 1

src/libc/src/eacsl.c
```diff
@@ -0,0 +1,198 @@
...
+void __e_acsl_assert_register_fun(__e_acsl_assert_data_t *data, const char *name) {}
+void __e_acsl_assert_register_struct(__e_acsl_assert_data_t *data, const char *name) {}
+void __e_acsl_assert_register_union(__e_acsl_assert_data_t *data, const char *name) {}
+void __e_acsl_assert_register_other(__e_acsl_assert_data_t *data, const char *name) {}
+void __e_acsl_assert_free_value(__e_acsl_assert_data_value_t *value) {}
+void __e_acsl_assert_copy_value(__e_acsl_assert_data_t *dest,
+                             __e_acsl_assert_data_value_t *value) {}
+void __e_acsl_assert_copy_values(__e_acsl_assert_data_t *dest,
+                              __e_acsl_assert_data_t *src) {}
+void __e_acsl_assert_clean(__e_acsl_assert_data_t *data) {}
```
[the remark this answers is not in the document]
Laplace-Demon: Yes all the function declarations and non-empty function definitions are from Frama-C code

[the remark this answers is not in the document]
Laplace-Demon: https://git.frama-c.com/pub/frama-c/-/tree/master/src/plugins/e-acsl/share/e-acsl?ref_type=heads

zapashcanon:
I believe this on deserve an entry in `CHANGES.md` @Laplace-Demon :)

test/c/eacsl/not-yet-supported/abrupt termination/breaks.c
```diff
@@ -0,0 +1,21 @@
...
+        if (x % 11 == 0) break;
+        x--;
+        if (x % 7 == 0) continue;
+        x--;
+        if (x % 5 == 0) return x;
+        x--;
+    }
+    }
+    return 0;
+}
```
zapashcanon, line 21: Could you avoid using a space in the folder name `abrupt termination` ? `abrupt_termination` would be better :)

zapashcanon:
@Laplace-Demon, could you rebase this one so I can run the CI ? Also, if you had other changes, feel free to add them.

Also, it would be nice to have at least one example in `example/c/README.md` to make the feature a little bit more discoverable.

zapashcanon:
Looks like you need to add `frama-c` as a test dependencies. :)

Laplace-Demon:
> Looks like you need to add `frama-c` as a test dependency. :)

Thanks :)

Laplace-Demon:
I had tests for not-yet-supported E-ACSL features so those tests are bound to fail. Do you know ways to deal with these tests?

zapashcanon:
Thanks for the example. I'm not sure if having a wrongly E-ACSL annotated code and show that Owi can help in detecting it is the best way to present it, I'd rather expect the other way round, but that's probably fine.

I saw you added `frama-c` to the `dune-project`, could you add it as a *test* dependency ? That is, writing `(frama-c :with-test)` instead ? You'll also need to run `dune build` in order to generate the new `.opam` file containing the frama-c dependency.

> I had tests for not-yet-supported E-ACSL features so those tests are bound to fail. Do you know ways to deal with these tests?

It should not be a problem to have test that are failing.

Laplace-Demon:
> Thanks for the example. I'm not sure if having a wrongly E-ACSL annotated code and show that Owi can help in detecting it is the best way to present it, I'd rather expect the other way round, but that's probably fine.
> 
> I saw you added `frama-c` to the `dune-project`, could you add it as a _test_ dependency ? That is, writing `(frama-c :with-test)` instead ? You'll also need to run `dune build` in order to generate the new `.opam` file containing the frama-c dependency. Otherwise the CI is not able to install it for instance.
> 
> > I had tests for not-yet-supported E-ACSL features so those tests are bound to fail. Do you know ways to deal with these tests?
> 
> It should not be a problem to have test that are failing.

What do you mean by "the other way round"? I also find it more or less bizarre to present it this way. But it's not that informative to simply illustrate the execution of a correctly monitored program.

zapashcanon:
Some tests are failing in the following way:

```diff
   $ owi c --e-acsl integer_dividable.c --no-value
-  Assert failure: (bool.eq (i32.of_bool (bool.eq (i32.rem symbol_0 symbol_1) (i32 0))) (i32 1))
+  Assert failure: (bool.eq (i32.rem symbol_0 symbol_1) (i32 0))
   Model:
     (model
       (symbol_0 i32)
```

Could you run `dune runtest` and then `dune promote` ? It should fix most of them (I guess that you've been running them before rebasing and the default optimisation level changed since then, which may explain the mismatch).

If it is not enough, you could rewrite the `.t` files like this:

```cram
  $ owi c --e-acsl integer_dividable.c --no-value -w1
  ...
  Assert failure: (bool.eq (i32.rem symbol_0 symbol_1) (i32 0))
  Model:
    (model
      (symbol_0 i32)
```

Having a `...` line allows to match any output. I'm not sure why they differ here (it could be the optimisations applied by Clang for instance, if the version differ on CI and your machine) but it should not matter. Adding `-w1` will also make stuff more deterministic.

test/c/eacsl/function/multi_behaviors.t
```diff
@@ -1,2 +1,5 @@
   $ owi c --no-value ./multi_behaviors.c --no-value
-  All OK
+  owi: option '--no-value' cannot be repeated
```
zapashcanon, line 2: You should remove one of the `--no-value`

test/c/eacsl/not-yet-supported/abrupt_termination/returns.t
```diff
@@ -1,2 +1,12 @@
...
+  [e-acsl] returns.c:8: User Error: 
+    Term '\at(x,__invalid_label)' was used before being translated.
+    This usually happens when using a label defined after the place
+    where the Term should be translated
+  [kernel] Plug-in e-acsl aborted: invalid user input.
+  run ['/home/laplace_demon/.opam/5.2.0+flambda/bin/frama-c' '-e-acsl'
+       '-no-frama-c-stdlib' '-kernel-warn-key' 'CERT:MSC:38=inactive'
+       '-verbose' '0'
+       '-cpp-extra-args="-I/home/laplace_demon/WorkPlace/owi/_build/install/default/share/owi/libc"'
+       './returns.c' '-then-last' '-print' '-ocode' './returns_instrumented.c']: exited with 1
```
zapashcanon, line 11: Could you replace:

```txt
  [kernel] Plug-in e-acsl aborted: invalid user input.
  run ['/home/laplace_demon/.opam/5.2.0+flambda/bin/frama-c' '-e-acsl'
       '-no-frama-c-stdlib' '-kernel-warn-key' 'CERT:MSC:38=inactive'
       '-verbose' '0'
       '-cpp-extra-args="-I/home/laplace_demon/WorkPlace/owi/_build/install/default/share/owi/libc"'
       './returns.c' '-then-last' '-print' '-ocode' './returns_instrumented.c']: exited with 1
  [26]
```

By:

```txt
  [kernel] Plug-in e-acsl aborted: invalid user input.
  ...
  [26]
```

Otherwise, it contains references to you own laptop that are not going for others or CI when running tests.

test/c/eacsl/not-yet-supported/abrupt_termination/returns.t
```diff
@@ -1,2 +1,12 @@
   $ owi c --e-acsl ./returns.c
-  All OK
+  [e-acsl] returns.c:8: User Error: 
+    Term '\at(x,__invalid_label)' was used before being translated.
+    This usually happens when using a label defined after the place
+    where the Term should be translated
+  [kernel] Plug-in e-acsl aborted: invalid user input.
```
zapashcanon, line 6: Doesn't this mean that the mistake is ours here ? The error message looks suspicious to me

test/c/eacsl/not-yet-supported/data_invariant/global.t
```diff
@@ -1,7 +1,2 @@
   $ owi c --e-acsl ./global.c --no-value
-  Assert failure: %{.*}
-  Model:
-    (model
-      (symbol_0 i32))
-  Reached problem!
-  [13]
+  All OK
```
zapashcanon, line 2: Why is there no warning here?

test/c/eacsl/not-yet-supported/data_invariant/strong_global.t
```diff
@@ -1,7 +1,11 @@
...
-  Assert failure: %{.*}
-  Model:
-    (model
-      (symbol_0 i32))
-  Reached problem!
-  [13]
+  [kernel:annot-error] strong_global.c:4: Warning: unexpected token 'global'
+  [kernel] User Error: warning annot-error treated as fatal error.
+  [kernel] Frama-C aborted: invalid user input.
+  run ['/home/laplace_demon/.opam/5.2.0+flambda/bin/frama-c' '-e-acsl'
```
zapashcanon, line 5: Same here about adding `...`

test/c/eacsl/not-yet-supported/data_invariant/strong_global.t
```diff
@@ -1,7 +1,11 @@
   $ owi c --e-acsl ./strong_global.c --no-value
-  Assert failure: %{.*}
-  Model:
-    (model
-      (symbol_0 i32))
-  Reached problem!
-  [13]
+  [kernel:annot-error] strong_global.c:4: Warning: unexpected token 'global'
```
zapashcanon, line 2: This error message is suspicious ?

test/c/eacsl/not-yet-supported/data_invariant/strong_type.t
```diff
@@ -1,7 +1,11 @@
   $ owi c --e-acsl ./strong_type.c --no-value
-  Assert failure: %{.*}
-  Model:
-    (model
-      (symbol_0 i32))
-  Reached problem!
-  [13]
+  [kernel:annot-error] strong_type.c:4: Warning: unexpected token 'type'
```
zapashcanon, line 2: Same

test/c/eacsl/not-yet-supported/data_invariant/strong_type.t
```diff
@@ -1,7 +1,11 @@
...
+  [kernel] User Error: warning annot-error treated as fatal error.
+  [kernel] Frama-C aborted: invalid user input.
+  run ['/home/laplace_demon/.opam/5.2.0+flambda/bin/frama-c' '-e-acsl'
```
zapashcanon, line 5: Same

test/c/eacsl/not-yet-supported/data_invariant/type.t
```diff
@@ -1,7 +1,2 @@
   $ owi c --e-acsl ./type.c --no-value
-  Assert failure: %{.*}
-  Model:
-    (model
-      (symbol_0 i32))
-  Reached problem!
-  [13]
+  All OK
```
zapashcanon, line 2: No warning ?

test/c/eacsl/not-yet-supported/ghost_struct_field/field.t
```diff
@@ -1,2 +1,23 @@
...
+    15            loop variant length;
+    16        */
+    
+    17        while (node != NULL) {
+    18            node = node->next;
+    19            //@ ghost length --;
+    20        }
+    
+    21    }
+    22
```
zapashcanon, line 16: suspicious errors ?

test/c/eacsl/not-yet-supported/termination/general_measure.t
```diff
@@ -1,2 +1,11 @@
   $ owi c --e-acsl ./general_measure.c
-  All OK
+  [kernel:annot-error] general_measure.c:24: Warning: unexpected token ''
+  [kernel] User Error: warning annot-error treated as fatal error.
+  [kernel] Frama-C aborted: invalid user input.
+  run ['/home/laplace_demon/.opam/5.2.0+flambda/bin/frama-c' '-e-acsl'
```
zapashcanon, line 5: same about `...`

test/c/eacsl/not-yet-supported/abrupt_termination/returns.t
Laplace-Demon: Oh yes, I'll fix it

test/c/eacsl/not-yet-supported/abrupt_termination/returns.t
Laplace-Demon: It's the example on the E-ACSL reference and I guess the support for `returns` is incomplete (in fact the implementation reference marks it as non-supported)

test/c/eacsl/not-yet-supported/data_invariant/strong_global.t
Laplace-Demon: Same as above, `global` is not yet made a keyword

test/c/eacsl/not-yet-supported/data_invariant/type.t
Laplace-Demon: Same as above

test/c/eacsl/not-yet-supported/ghost_struct_field/field.t
Laplace-Demon: The grammar for defining ghost struct fields is to define again the same structure in the ghost code.

```c
struct List {
    int element;
    struct List* next;
};
/*@ ghost struct List {
    int order;
}; */
```

It's not supported so the field seems undefined

test/c/eacsl/not-yet-supported/termination/general_measure.t
```diff
@@ -1,2 +1,11 @@
   $ owi c --e-acsl ./general_measure.c
-  All OK
+  [kernel:annot-error] general_measure.c:24: Warning: unexpected token ''
```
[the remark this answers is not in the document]
Laplace-Demon: Designating a custom decreasing relation for loop variants is not supported, which leads to the error.

```c
/*@ predicate lexico(struct intpair p1, struct intpair p2) =
    (p1.x >  p2.x && p1.x >= 0) ||
    (p1.x == p2.x && p1.y >  p2.y && p1.y >= 0);
*/

...

/*@ loop invariant x >= 0 && y >= 0;
        loop variant   m for lexico
*/
while (...)

```

test/c/eacsl/not-yet-supported/abrupt_termination/returns.t
zapashcanon: OK, maybe it should be good to open an issue on e-acsl to ask about changing this to a warning and to ignore `returns` ?

test/c/eacsl/not-yet-supported/ghost_struct_field/field.t
zapashcanon: OK, maybe it should be good to open an issue on e-acsl to ask about changing this to a warning and to ignore `returns` ?

test/c/eacsl/not-yet-supported/termination/general_measure.t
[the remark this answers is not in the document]
zapashcanon: OK, maybe it should be good to open an issue on e-acsl to ask about changing this to a warning and to ignore `returns` ?

src/cmd/cmd_c.ml
```diff
@@ -33,6 +33,59 @@ let find location file : Fpath.t Result.t =
...
+      let framac_verbosity_level = if debug then "2" else "0" in
+      Cmd.(
+        of_list
+          [ "-e-acsl"
+          ; "-no-frama-c-stdlib"
+          ; "-kernel-warn-key"
+          ; "CERT:MSC:38=inactive"
+          ; "-verbose"
+          ; framac_verbosity_level
+          ; String.concat "" [ "-cpp-extra-args=\""; includes; "\"" ]
```
zapashcanon, line 53: You can rewrite this like this to avoid quoting:

```ocaml
; String.concat "" [ {|-cpp-extra-args="|}; includes; {|"|} ]
```

src/cmd/cmd_c.ml
```diff
@@ -33,6 +33,59 @@ let find location file : Fpath.t Result.t =
...
+
+    let outs =
+      List.map
+        (fun file ->
+          let ext = Fpath.get_ext ~multi:true file in
+          let file_without_ext = Fpath.rem_ext ~multi:true file in
+          let file_without_ext =
+            Fpath.to_string file_without_ext ^ "_instrumented"
+          in
+          let file = file_without_ext ^ ext in
```
zapashcanon, line 68: I don't think this is building with the latest changes adding `prelude`. What do you get when you do `dune build @all` ?

You should avoid the `Fpath.to_string` and `^` stuff, for instance, see how it is done in the `write_file` function from `ast/binary_encoder.ml`.

src/cmd/cmd_c.ml
```diff
@@ -33,6 +33,59 @@ let find location file : Fpath.t Result.t =
...
+          Fpath.v file )
+        files
+    in
+
+    let framac : Fpath.t -> Fpath.t -> Bos.Cmd.t =
+     fun file out ->
+      Cmd.(
+        framac_bin %% flags1 % Fpath.to_string file %% flags2
+        % Fpath.to_string out )
```
zapashcanon, line 77: You can replace this with `% p out` thanks to `val Bos.Cmd.p : Fpath.t -> string`.

src/libc/include/stdarg.h
```diff
@@ -3,10 +3,10 @@
...
 #include <endian.h>
 
-typedef char *va_list;
-#define va_start(ap, parmn) (void)((ap) = (char *)(&(parmn) + 1))
-#define va_end(ap) (void)((ap) = 0)
-#define va_arg(ap, type) (((type *)((ap) = ((ap) + sizeof(type))))[-1])
+typedef __builtin_va_list va_list;
+#define va_start(ap, param) __builtin_va_start(ap, param)
+#define va_end(ap)          __builtin_va_end(ap)
+#define va_arg(ap, type)    __builtin_va_arg(ap, type)
```
zapashcanon, line 9: Do you want to submit this as a patch to E-ACSL where I believe the non-builtin version is hardcoded in `eacsl.h` ?

src/libc/src/eacsl.c
```diff
@@ -0,0 +1,198 @@
...
+                                int is_enum, long value) {}
+void __e_acsl_assert_register_ulong(__e_acsl_assert_data_t *data, const char *name,
+                                 int is_enum, unsigned long value) {}
+void __e_acsl_assert_register_longlong(__e_acsl_assert_data_t *data, const char *name,
+                                    int is_enum, long long value) {}
+void __e_acsl_assert_register_ulonglong(__e_acsl_assert_data_t *data,
+                                     const char *name, int is_enum,
+                                     unsigned long long value) {}
+// void __e_acsl_assert_register_mpz(__e_acsl_assert_data_t *data, const char *name,
+//                                int is_enum, const __e_acsl_mpz_t value) {}
```
zapashcanon, line 172: If those are unused, can we remove the comments ?

src/libc/src/eacsl.c
```diff
@@ -0,0 +1,198 @@
...
+                                        __e_acsl_assert_data_rkind_t kind,
+                                        __e_acsl_assert_data_real_value_t value) {}
+void __e_acsl_assert_register_float(__e_acsl_assert_data_t *data, const char *name,
+                                 float value) {}
+void __e_acsl_assert_register_double(__e_acsl_assert_data_t *data, const char *name,
+                                  double value) {}
+void __e_acsl_assert_register_longdouble(__e_acsl_assert_data_t *data,
+                                      const char *name, long double value) {}
+// void __e_acsl_assert_register_mpq(__e_acsl_assert_data_t *data, const char *name,
+//                                const __e_acsl_mpq_t value) {}
```
zapashcanon, line 184: Same

test/c/eacsl/not-yet-supported/abrupt_termination/returns.t
Laplace-Demon: Ok

src/cmd/cmd_c.ml
Laplace-Demon: Thanks!

src/cmd/cmd_c.ml
Laplace-Demon: Seems I've not rebased to the lastest commits

src/libc/src/eacsl.c
Laplace-Demon: We have to comment them out because we cannot use GMP in webassembly. These functions can be generated by E-ACSL, though.

zapashcanon: Could you remove the comments and add an example where E-ACSL generates such functions ?

src/libc/include/stdarg.h
Laplace-Demon: Yes, the non-builtin MACROs are in `stdarg.h` but I believe they should at least adjust them according to architecture-dependent configuration ?

zapashcanon: Yes, you can simply change their code to use `stdarg.h` :)

Laplace-Demon:
> Some tests are failing in the following way:
> 
> ```diff
>    $ owi c --e-acsl integer_dividable.c --no-value
> -  Assert failure: (bool.eq (i32.of_bool (bool.eq (i32.rem symbol_0 symbol_1) (i32 0))) (i32 1))
> +  Assert failure: (bool.eq (i32.rem symbol_0 symbol_1) (i32 0))
>    Model:
>      (model
>        (symbol_0 i32)
> ```
> 
> Could you run `dune runtest` and then `dune promote` ? It should fix most of them (I guess that you've been running them before rebasing and the default optimisation level changed since then, which may explain the mismatch).
> 
> If it is not enough, you could rewrite the `.t` files like this:
> 
> ```
>   $ owi c --e-acsl integer_dividable.c --no-value -w1
>   ...
>   Model:
>     (model
>       (symbol_0 i32)
> ```
> 
> Having a `...` line allows to match any output. I'm not sure why they differ here (it could be the optimisations applied by Clang for instance, if the version differ on CI and your machine) but it should not matter. Adding `-w1` will also make stuff more deterministic.

There's a minor problem regarding dune's cram tests. According to (dune's documentation)[https://dune.readthedocs.io/en/stable/tests.html#test-output-sanitation], regular expression is not to be implemented and `diff` is verbatim.

> Note: Unlike Dune’s version of Cram, the original specification for Cram supports regular expression and glob filtering for matching output. We chose not to implement this feature because it breaks the test, diff, and accept cycle. With regex or glob matching, the output must now be manually inspected and possibly updated. We consider the postprocessing approach described here as superior and will not introduce output matchers.

zapashcanon:
Oh indeed, this is working with MDX but not dune, I forgot about it...

Did you first tried to:

> Could you run dune runtest and then dune promote ? It should fix most of them (I guess that you've been running them before rebasing and the default optimisation level changed since then, which may explain the mismatch).

I believe it should work ?

Laplace-Demon:
> Oh indeed, this is working with MDX but not dune, I forgot about it...
> 
> Did you first tried to:
> 
> > Could you run dune runtest and then dune promote ? It should fix most of them (I guess that you've been running them before rebasing and the default optimisation level changed since then, which may explain the mismatch).
> 
> I believe it should work ?

Yes `dune promote` works.

What about placing the expected output directly (All OK)?

Laplace-Demon:
@zapashcanon Is a602f8a what we need ?

zapashcanon:
> What about placing the expected output directly (All OK)?

You can but it's tedious, it's much better to always: `dune runtest`, check the output, if it's okay, then, `dune promote`.

test/c/eacsl/not-yet-supported/gmp/explicit.c
```diff
@@ -0,0 +1,15 @@
...
+    owi_var = owi_i32();
+
+    mpz_t gmp_var;
+    mpz_init_set_si(gmp_var, 0);
+
+    int result = mpz_cmp_si(gmp_var, owi_var) == 0;
+    owi_assert(result);
+
+    return 0;
+}
```
zapashcanon, line 15: Missing newline at the end of file :)

src/libc/include/eacsl.h
```diff
@@ -1,3 +1,23 @@
...
+/*                                                                            */
+/*                                                                            */
+/*  SPDX-License-Identifier: AGPL-3.0-or-later                                */
+/*                                                                            */
+/*  Copyright © 2021-2024 OCamlPro                                            */
+/*                                                                            */
+/*  Modified by the Owi programmers                                           */
+/*                                                                            */
+/******************************************************************************/
+
```
zapashcanon, line 20: Thanks! Could you now make it shorter:

```c
// SPDX-License-Identifier: LGPL-2.1-or-later
// Copyright (C) 2012-2024 CEA (Commissariat à l'énergie atomique et aux énergies alternatives
// This file is originally from the Frama-C's E-ACSL plug-in available at https://git.frama-c.com/pub/frama-c/-/tree/master/src/plugins/e-acsl

// SPDX-License-Identifier: AGPL-3.0-or-later
// Copyright © 2021-2024 OCamlPro
// Modified by the Owi programmers
```

zapashcanon:
> @zapashcanon Is https://github.com/OCamlPro/owi/commit/a602f8a5a70cdecd49d9b163bec8ed1b79b4248b what we need ?

Yes, thanks! I added a small review. Also, could you rebase on the main branch so the CI is actually run?

Laplace-Demon:
> > @zapashcanon Is [a602f8a](https://github.com/OCamlPro/owi/commit/a602f8a5a70cdecd49d9b163bec8ed1b79b4248b) what we need ?
> 
> Yes, thanks! I added a small review. Also, could you rebase on the main branch so the CI is actually run?

ok

Laplace-Demon:
Running `dune runtest` can always trigger the segfault mentioned in #387

zapashcanon:
Thanks!

PR #328, opened 2024-06-24
Fixes Ptr smtml change

filipeom:
This together with the `main` branch of `smtml` should allow us to use Bitwuzla without segmentation faults


zapashcanon:
Thanks!

PR #333, opened 2024-06-27
Refactor symbol creation functions

filipeom:
- Use `Thread.symbol` for creating unique symbols, replacing the global atomic variable in `symbolic_wasm_ffi.ml`.

This makes symbol names nicer for each path since now they are sequential. (c.f. [in the promoted test](https://github.com/OCamlPro/owi/pull/333/files#diff-c59bd1833c328a05ea217570fd32fe02ef78b6b7dfab8455bbf01c849604ce9a)) 

- Add `with_new_symbol` to `symbolic_choice` to mirror the approach in `concolic_wasm_ffi`.

- Rename `Thread.incr_choices` to `incr_symbols` and make `Thread.symbol_set` immutable.

### DIFF
--- a/src/symbolic/symbolic_choice.mli
+++ b/src/symbolic/symbolic_choice.mli
@@ -13,6 +13,8 @@ include
      and type 'a run_result = ('a eval * Thread.t) Seq.t
      and module V := Symbolic_value
 
+val with_new_symbol : Smtml.Ty.t -> (Smtml.Symbol.t -> 'b) -> 'b t
+
 val run :
      workers:int
   -> Smtml.Solver_dispatcher.solver_type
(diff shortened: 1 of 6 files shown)


src/symbolic/symbolic_choice.mli
```diff
@@ -13,6 +13,8 @@ include
      and type 'a run_result = ('a eval * Thread.t) Seq.t
      and module V := Symbolic_value
 
+val with_new_symbol : Smtml.Ty.t -> (Smtml.Symbol.t -> 'b) -> 'b t
```
filipeom, line 16: I'm not sure if I should add this to `Choice_intf.Complete` instead?

zapashcanon: It's fine to have it there for now. :)

krtab (commented):
LGTM

PR #342, opened 2024-07-03
Add some comments

krtab:


src/ast/types.ml
```diff
@@ -727,9 +727,12 @@ let import_desc fmt : 'a import_desc -> Unit.t = function
...
 
 type 'a import =
-  { modul : string
-  ; name : string
+  { modul : string  (** The name of the module from which the import is done *)
+  ; name : string  (** The name of the importee in its module of origin *)
   ; desc : 'a import_desc
+      (** If this import_desc first field is Some s, the importee is made
+          available under name s, else it can only be used via its numerical
+          index.*)
```
zapashcanon, line 735: I think the `desc` part is incorrect. It is not an option, it refers to which kind of import this is (func/table/memory/global) along with some "type" information

zapashcanon: Oh, I missed the `first field` part.

PR #350, opened 2024-07-03
change default optimization level to O3

zapashcanon:


zapashcanon:
cc @filipeom, in [test/c/collections-c/pqueue_tests.t](https://github.com/OCamlPro/owi/pull/350/files#diff-95c63926f646198e78f1f511908f32c8ce94f57b27b0d921b52259e710062f05) we are not detecting the heap overflow with `-O3`, I don't know these tests so I would say it's a regression but on Test-Comp we're clearly better with `-O3` so I'd like to still make it the default, what do you think?

filipeom:
> cc @filipeom, in [test/c/collections-c/pqueue_tests.t](https://github.com/OCamlPro/owi/pull/350/files#diff-95c63926f646198e78f1f511908f32c8ce94f57b27b0d921b52259e710062f05) we are not detecting the heap overflow with `-O3`, I don't know these tests so I would say it's a regression but on Test-Comp we're clearly better with `-O3` so I'd like to still make it the default, what do you think?

I think it's ok, its expected when one compiles with optimizations some memory errors might no longer be detected. For Test-Comp we're not focused on verifying memory safety so I agree that we should make the best parameters the default :smile:

PR #353, opened 2024-07-05
Simplify binary modules types

zapashcanon:
depends on #352

### DIFF
--- a/src/ast/binary_encoder.ml
+++ b/src/ast/binary_encoder.ml
@@ -89,15 +89,21 @@ let write_valtype buf vt =
   let c = get_char_valtype vt in
   Buffer.add_char buf c
 
-let encode_vector buf datas encode_func =
+let encode_vector length iter buf datas encode_func =
   let vector_buf = Buffer.create 16 in
-  let len = List.length datas in
-  List.iter (encode_func vector_buf) datas;
+  let len = length datas in
+  iter (encode_func vector_buf) datas;
   write_u32_of_int buf len;
   Buffer.add_buffer buf vector_buf
 
+let encode_vector_list buf datas encode_func =
+  encode_vector List.length List.iter buf datas encode_func
+
+let encode_vector_array buf datas encode_func =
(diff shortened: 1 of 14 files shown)


src/ast/binary_encoder.ml
```diff
@@ -742,19 +741,31 @@ let keep_imported { Named.values; _ } =
...
-  let local_memories = keep_local modul.mem in
-  let local_globales = keep_local modul.global in
-  let imported_funcs = keep_imported modul.func in
-  let imported_tables = keep_imported modul.table in
-  let imported_memories = keep_imported modul.mem in
-  let imported_globals = keep_imported modul.global in
+
+  let indexed_list_of_array a = Array.mapi Indexed.return a |> Array.to_list in
+
+  let func = indexed_list_of_array modul.func in
```
zapashcanon, line 747: note to self: this could be simplified further

src/link/link.ml
```diff
@@ -152,13 +152,16 @@ let eval_global ls env (global : (Binary.global, binary global_type) Runtime.t)
...
 let eval_globals ls env globals : Link_env.Build.t Result.t =
-  Named.fold
-    (fun id global env ->
-      let* env in
-      let* global = eval_global ls env global in
-      let env = Link_env.Build.add_global id global env in
-      Ok env )
-    globals (Ok env)
+  let+ env, _i =
+    (* TODO: use array_fold_left instead *)
```
Laplace-Demon, line 156: `array_fold_left` is already used here

zapashcanon:
This should be correct now. There was some subtle bug about extra `List.rev` hidden by the usage of `Named.fold`... I believe this is the one that caused some debugging troubles to @epatrizio and @krtab some time ago.

This makes the code shorter, easier to understand, more efficient and fix some bugs. Thus I'm going to merge once CI is happy. :)

@Laplace-Demon, could you confirm it it also fixes the issue you are encountering in your other PR ?

PR #356, opened 2024-07-05
Allow passing the memory to external functions

filipeom:
Adds a `Mem` argument to external function to allow direct manipulation of the memory.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ## unreleased
 
+- adds a `Mem` argument to external function to allow direct manipulation of the memory.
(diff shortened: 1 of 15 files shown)


zapashcanon:
Wow, thanks! We wanted to do it for a long time...

Could you add an example, maybe [here](https://github.com/OCamlPro/owi/tree/main/example/define_host_function) ?

filipeom:
> Wow, thanks! We wanted to do it for a long time...
> 
> Could you add an example, maybe [here](https://github.com/OCamlPro/owi/tree/main/example/define_host_function) ?

Yes of course :smiley:

zapashcanon:
Thanks! :partying_face:

PR #360, opened 2024-07-09
Create Synchronzier module

krtab:


krtab:
Because Ocaml does not possess an optimizing compiler, we should check that this does not tank performances and manually add [@@inline] where needed to try and salvage what demands salvation.

zapashcanon:
Can you add a `.mli` ?

src/data_structures/synchronizer.ml
```diff
@@ -0,0 +1,64 @@
+(* SPDX-License-Identifier: AGPL-3.0-or-later *)
+(* Copyright © 2021-2024 OCamlPro *)
+(* Written by the Owi programmers *)
+
+type ('get, 'write) t =
+  { mutex : Mutex.t
+  ; cond : Condition.t
+  ; getter : unit -> 'get option
```
zapashcanon, line 8: It is not clear to me why this has type `unit -> 'get option` and not `'get option` by looking at the code, could you explain ?

krtab: The idea is that this is a closure that when called, gets one element to run, without caring for synchronization. Typically, for a Queue, this is `fun () -> Queue.take_opt q`, and for concolic this will be `fun () -> find_node_to_run tree`.

src/data_structures/synchronizer.ml
```diff
@@ -0,0 +1,64 @@
...
+  Mutex.unlock synchro.mutex
+
+let end_pledge synchro =
+  Mutex.lock synchro.mutex;
+  synchro.pledges <- synchro.pledges - 1;
+  Condition.broadcast synchro.cond;
+  Mutex.unlock synchro.mutex
+
+let write v synchro =
+  Mutex.protect synchro.mutex (fun () -> synchro.writter v synchro.cond)
```
zapashcanon, line 41:
```ocaml
let write v { writer; cond; mutex; _ } =
  Mutex.protect mutex (fun () -> writter v cond)
```

krtab:
Once this is effectively used in concollic, I'll add documentation and close #363

zapashcanon:
I'm asking for a `.mli` not only to add documentation but to make sure the interface is properly constrained to the things that are actually used outside (so that the compiler can tell us about unused stuff and the like).

krtab:
Yes, I have added a mli :)

src/data_structures/synchronizer.ml
```diff
@@ -0,0 +1,64 @@
...
+  Mutex.lock q.mutex;
+  q.closed <- true;
+  Condition.broadcast q.cond;
+  Mutex.unlock q.mutex
+
+let rec work_while f q =
+  match get q true with
+  | None -> ()
+  | Some v ->
+    let () = f v (fun v -> write v q) in
```
zapashcanon, line 62: `f v ...;` was more idiomatic.

krtab: But didn't constrain the return type to be unit.

zapashcanon:
Thanks!

PR #361, opened 2024-07-09
Remove binary modul from Link.modul_to_run

krtab:
Only its id is needed, and it makes for a clearer and less confusing interface.

### DIFF
--- a/src/concolic/concolic.ml
+++ b/src/concolic/concolic.ml
@@ -257,14 +257,14 @@ module P = struct
 
   module Module_to_run = struct
     type t =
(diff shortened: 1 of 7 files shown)


zapashcanon:
I rebased. IIRC I had a doubt about the correctness of this (I was afraid of mixing the identifier of a module coming from `(module $name` and the name used for registering/linking). 

On the other hand, this does not seem to break anything so I'm merging as it is a simplification. I'm merging, if this was wrong there's still the possibility to revert in the future.

Thanks!

PR #362, opened 2024-07-10
Add and use a new compilation API

krtab:


zapashcanon:
I would like to avoid having two different APIs but I like the direction this is going.

I think this is quite easily doable by making a few changes.

Adding a module `Any` (feel free to find a better name) to `compile.mli` which would have the following signature:

```ocaml
module Any : sig
  val until_typecheck : unsafe:bool -> (Text.modul, Binary.modul) Either.t -> (Text.modul, Binary.modul) Either.t Result.t

  val until_optimize : unsafe:bool -> optimize:bool -> (Text.modul, Binary.modul) Either.t ->

  ...


+end
```

Then, change the signature of `Parse.guess_from_file` from `((Text.modul, Text.script) Either.t, Binary.modul) Either.t Result.t` to `((Text.modul, Binary.modul) Either.t, Text.script) Either.t Result.t` (permutting `Binary.modul` and `Text.script` in order to be able to easily pipe the result of this function to the one defined before.

Add `add_main_as_start` everywhere it is needed (in `Compile.{Text,Binary,Any}.*`).

Then if you want to avoid piping the output of `Parse.guess_from_file` to `Compile.Any.*` many time, we could define:

`Compile.File` and `Compile.Files` with the same interface as `Any` but taking `Fpath.t` and `Fpath.t list` as input respectively instead of an `(Text.modul, Binary.modul) Either.t`.

zapashcanon:
Also this is actually fixing #249

PR #357, opened 2024-07-05
[WIP] Make symbolic memory parametric

filipeom:
Needed for #338. Depends on #367

PR #367, opened 2024-07-15
Parameterize Thread on Symbolic Memory and Choice Monad on Thread 

filipeom:
This PR parameterizes the `Thread` module on the symbolic memory and the `Choice_monad` module on a `Thread`, with @krtab doing the heavy lifting.

#### Summary 

This is needed for #357 because we're going to have to use the `Symbolic_choice` within the `Symbolic_memory`. The current implementation of `Symbolic_choice` would not allow us to do so since the original `Thread` module was defined using the symbolic memory itself, causing a circular dependency. Therefore, we parameterized `Thread` and subsequently `Symbolic_choice`, resulting in the following modules:

- `Thread_with_memory`: This is equivalent to the previous implementation of `Thread` in that it contains a reference to the symbolic memory. This module is used to construct `Symbolic_choice_with_memory`.
- `Thread_without_memory`: A thread without references to the symbolic memory.
- `Symbolic_choice_with_memory`: This is equivalent to the previous implementation of `Symbolic_choice` and is the module used to build the symbolic interpreter in `symbolic.ml`.
- `Symbolic_choice_without_memory`: A choice module built with `Thread_without_memory` to be used within the `symbolic_memory` module in #357.

#### Notes

The .mlis probably need more work. I purposefully removed `symbolic_choice.mli` to appease the type checker, but we should probably add it again to avoid having dead code within `symbolic_choice.ml`. Nevertheless, I think this is ready-ish for review :smile:


src/cmd/cmd_sym.ml
```diff
@@ -58,14 +58,18 @@ let cmd profiling debug unsafe optimize workers no_stop_at_failure no_values
   let* _created_dir = Bos.OS.Dir.create ~path:true ~mode:0o755 workspace in
   let pc = Choice.return (Ok ()) in
   let result = List.fold_left (run_file ~unsafe ~optimize) pc files in
-  let thread : Thread.t = Thread.create () in
+  let thread = Thread_with_memory.init () in
   let res_queue = Wq.init () in
-  let callback = function
-    | Symbolic_choice.EVal (Ok ()), _ -> ()
+  (* Hackity hack *)
+  let path_counter = ref 0 in
```
zapashcanon, line 64: It seems like this is incremented in `callback` but never read ?

filipeom: Ah oops, I did this just to debug a thing. I'll remove it

filipeom:
> OK, I read the code a first time and left minor comments. I'll probably have to re-read it in a few days and spend more time on it if you want a proper review. 

Yeah this is not urgent so feel free to take your time :)

> Did you get @chambart's feedback on the change? (We never know if there's a magic way to make this simpler...)

Me and Arthur briefly discussed this with Pierre but we haven't had his take on it just yet.

zapashcanon:
@filipeom, could you add the missing `.mli` and rebase? This is OK to merge for me. :)

filipeom:
Reintroduced `symbolic_choice.mli` in https://github.com/OCamlPro/owi/pull/367/commits/8a20a3987003edf84f1d2928cc8a6772ff4eb963. I think this might now break the isolation of the symbolic choice monad, but I don't see an alternative to exposing `CoreImpl.State` in `symbolic_choice.mli` to make `Symbolic_choice_with_memory.lift_mem` typecheck.

zapashcanon:
Thanks!

PR #368, opened 2024-07-18
Custom annotation

Laplace-Demon:


src/parser/test.wat
```diff
@@ -0,0 +1,3 @@
+(module
+  (@a df dcd feevsd 1 , c;vdaead 2dcecwef .)
+)
```
zapashcanon, line 3: I believe this is a leftover file that should be removed?

src/parser/text_lexer.ml
```diff
@@ -432,12 +435,31 @@ let rec token buf =
...
   | ";;" ->
     single_comment buf;
     token buf
   | "(;" ->
     comment buf;
     token buf
+  (* custom annotation *)
+  | "(@", name ->
+    let annotid = Utf8.lexeme buf in
+    let annotid = String.sub annotid 3 (String.length annotid - 4) in
```
zapashcanon, line 448: Maybe check that the string has the minimum required length before using `String.sub` and raise an error if not ?

src/parser/text_lexer.ml
```diff
@@ -432,12 +435,31 @@ let rec token buf =
...
+    let annotid = mk_string buf annotid in
+    if annotid = "" then Log.err "empty annotation id"
```
zapashcanon, line 450: `if String.equal "" annotid`

src/parser/text_lexer.ml
```diff
@@ -432,12 +435,31 @@ let rec token buf =
...
+    else
+      let items = annot buf in
+      Annot.(record_annot { annotid; items });
+      token buf
+  | "(@", Plus id_char ->
+    let annotid = Utf8.lexeme buf in
+    let annotid = String.sub annotid 2 (String.length annotid - 2) in
```
zapashcanon, line 457: Same here about the length check.

src/parser/text_lexer.ml
```diff
@@ -473,4 +495,56 @@ and single_comment buf =
...
+  | ";;" ->
+    single_comment buf;
+    annot buf
+  | "(;" ->
+    comment buf;
+    annot buf
+  (* custom annotation *)
+  | "(@", name ->
+    let annotid = Utf8.lexeme buf in
+    let annotid = String.sub annotid 3 (String.length annotid - 4) in
```
zapashcanon, line 511: same

src/parser/text_lexer.ml
```diff
@@ -473,4 +495,56 @@ and single_comment buf =
...
+    let annotid = mk_string buf annotid in
+    if annotid = "" then Log.err "empty annotation id"
```
zapashcanon, line 513: same

src/ast/annot.ml
```diff
@@ -1,3 +1,5 @@
+open Format
```
zapashcanon, line 1: `Format` is no more available since the `prelude` related changes

src/ast/annot.ml
```diff
@@ -14,6 +16,24 @@ and item =
...
+    annot.items
+
+and pp_item fmt = function
+  | Atom atom -> pp_string fmt atom
+  | String str -> pp_string fmt str
+  | Id id -> Types.pp_id fmt id
+  | Int i -> pp_string fmt i
+  | Float f -> pp_string fmt f
+  | Parens items -> pp_list ~pp_sep:pp_space pp_item fmt items
+  | Annot annot -> pp_annot fmt annot
```
zapashcanon, line 35: You'll have to replace `pp_string` by `Fmt.string`, `pp_list` by `list` and `~pp_sep` by `~sep`.

src/ast/text.ml
```diff
@@ -100,7 +100,9 @@ type modul =
   }
 
 let pp_modul fmt (m : modul) =
-  pp fmt "(module%a@\n  @[<v>%a@]@\n)" pp_id_opt m.id
+  pp fmt "%a(module%a@\n  @[<v>%a@]@\n)"
+    (pp_list ~pp_sep:pp_newline pp_annot)
```
zapashcanon, line 104: `(Fmt.list ~sep:pp_newline pp_annot)`

src/ast/annot.ml
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+and item =
+  | Atom of string
+  | String of string
+  | Id of string
+  | Int of string
+  | Float of string
+  | Parens of item list
+  | Annot of annot
```
zapashcanon, line 15: Where does the `Annot of annot` comes from ? I don't see this in the spec, I only see something that matches your `Parens` case.

src/ast/annot.ml
```diff
@@ -0,0 +1,55 @@
...
+  | Parens items -> pp_list ~pp_sep:pp_space pp_item fmt items
+  | Annot annot -> pp_annot fmt annot
+
+let annot_recorder : annot list NameMap.t ref = ref NameMap.empty
+
+let find_nil key map = Option.value (NameMap.find_opt key map) ~default:[]
+
+let record_annot annot =
+  let old = find_nil annot.annotid !annot_recorder in
+  annot_recorder := NameMap.add annot.annotid (annot :: old) !annot_recorder
```
zapashcanon, line 43: If you use a `Hashtbl` instead of a `Map` you can avoid having to record previous bindings by yourself as they'll be stored in the table by default when using `Hashtbl.add`, you can recover all bindings with `Hashtbl.find_all`.

src/ast/annot.ml
```diff
@@ -21,17 +21,16 @@ module NameMap = Map.Make (String)
...
+    (list ~sep:sp pp_item) annot.items
 
 and pp_item fmt = function
-  | Atom atom -> pp_string fmt atom
-  | String str -> pp_string fmt str
+  | Atom atom -> Fmt.string fmt atom
+  | String str -> Fmt.string fmt str
   | Id id -> Types.pp_id fmt id
-  | Int i -> pp_string fmt i
-  | Float f -> pp_string fmt f
```
zapashcanon: As you opened `Fmt` at the beginning of the module, you can remove the `Fmt.` qualification on these cases

src/ast/annot.ml
zapashcanon: I believe Hashtbl to be better. :)

src/ast/annot.ml
zapashcanon: Actually I believe we should use a representation close to S-expressions:

```ocaml
type sexp = Atom of string | List of sexp list
```

WDYT ?

Laplace-Demon: After offline discussion: use a representation close to S-expressions

zapashcanon: https://github.com/janestreet/sexplib0/blob/master/src/sexp.mli

zapashcanon:
Could you also change the status of `Custom Annotation Syntax in the Text Format` from `❌` to `Ongoing` in the [Current proposals](https://github.com/ocamlpro/owi?tab=readme-ov-file#current-proposals) section?

src/annot/function_contract.ml
```diff
@@ -0,0 +1,36 @@
...
+
+let parse_clauses = list_fold_left parse_clause ([], [])
+
+let parse Annot.{ annotid; items } =
+  if not (String.equal annotid name) then
+    Error (`Annotation_id_incorrect annotid)
+  else
+    let* func, items' = parse_func items in
+    let* preconditions, postconditions = parse_clauses items' in
+    Ok { func; preconditions; postconditions }
```
zapashcanon, line 36: When you have a sequence:

```ocaml
let* ...
...
let* ...
Ok e
```

You can always replace the last `let*` by a `let+` which allows you to remove the `Ok`. Sometimes it can help in making the code more readable by removing a lot of parenthesis or reduce the indentation level.

Laplace-Demon: Thanks!

src/annot/spec.ml
```diff
@@ -0,0 +1,98 @@
...
+  | UnConnect : unconnect * 'a prop -> 'a prop
+  | BinConnect : binconnect * 'a prop * 'a prop -> 'a prop
+  | Binder : binder * binder_type * string option * 'a prop -> 'a prop
+
+and 'a term =
+  | Int32 : int32 -> 'a term
+  | Var : text indice -> text term
+  | GlobalVar : 'a indice -> 'a term
+  | BinderVar : 'a indice -> 'a term
+  | Result : 'a term
```
zapashcanon, line 51: Why are you making `prop` and `term` mutually recursive ? You could define `term` first and then `prop`, right?

Laplace-Demon: I'm considering using props in if-else expression. But that's incremental and I'm just leaving the place of that.

For the GADT, it's for writing the production `| Var : text indice -> text term`. I think it's good for two reasons:
- It can allow using string ids in text annotations, which may facilitate writing specifications.
- And we don't need to define separate types for text contract and binary contract. Text contract can be of type `text contract` (which additionally allows the constructor `Var`) and binary can be `binary contract`.

(But that also means we have to parse contracts in text format, and transform it to binary format through the group+assign+rewrite process. What do you think of this?)

zapashcanon: > I'm considering using props in if-else expression. But that's incremental and I'm just leaving the place of that.

OK fine with me. I'd still remove the mutual recursion for now to make everything simpler as a start but you can keep it if you want.

Laplace-Demon: I think I can make a "clean" version once I have the whole process running :) I admit these commits are mainly for visualization rather than works finished.

Laplace-Demon: > Yes, that's the plan, we'll have to do this at some point.

In our earlier discussion we said to parse the contract in binary format. I guess it's a slip of the tongue?

zapashcanon: We can start by parsing it it the binary format but it'll produce `binary prop`/`binary term` directly. But it'll be hard to test because I'm not sure you'll want to write binary propositions by hand. :-)

So indeed, it's probably better to start by parsing them in the text format, producing `text prop`/`text term`. Then do the translation (you can also start by not doing the translation of names and only translate everything directly with an identity function that fails on text identifiers). Then we'll do the generation.

I'm not sure if people are likely to write annotations by hand in the binary format so it probably doesn't matter in the beginning. It may be useful if we want to try to compile E-ACSL annotations to Wasm annotations, but's that's probably a full PhD work haha.

zapashcanon:
Closing in favor of #407.

Issue #278, opened 2024-05-23
new test not passing: unexpected token "\"(func (export \"f1\") (result i32)\""


zapashcanon:
After #276:

```shell-session
$ owi script --no-exhaustion reference/comments.wast
unexpected token "\"(func (export \"f1\") (result i32)\""
[40]
```

This is related to UTF-8 definition of newline, see https://github.com/WebAssembly/spec/commit/43d405f5301adefb823b443c6fb9bb6093b1e1a7

epatrizio:
here's a minimalist test :
```wast
(module quote
  "(func )"
)
```

epatrizio:
```wast
(module quote
  "(func"
  "  ;; comment\0a"
  ")"
  "(func"
  "  ;; comment\0d"
  ")"
  "(func"
  "  ;; comment\0d\0a"
  ")"
)
```
there may be 2 different problems because the first minimalist example does not mention `newline` character ... ?!

krtab:
> here's a minimalist test :
> 
> ```webassembly
> (module quote
>   "(func )"
> )
> ```

I'm opening a issue (and trying a fix) for this one.

PR #383, opened 2024-07-23
actually parse inline binary and quoted module in scripts

### LINKED ISSUE #279: new test not passing: unbound name 4294967249
After #276:

```shell-session
$ owi script --no-exhaustion reference/float_literals.wast
unbound name 4294967249
[38]
```

zapashcanon:
Fix #279. The binary parser code was actually correct. The issue was that we were not parsing inline binary module correctly...


zapashcanon:
Also fix #278. @Laplace-Demon it should fix you issue regarding custom annotations tests btw. :)

PR #389, opened 2024-07-26
update to latest prelude and smtml, add missings dev-setup deps

zapashcanon:


filipeom:
I made a release for https://github.com/formalsec/smtml/pull/189. So we should probably use `smtml >= 0.2.4` since 0.2.3 is incompatible with prelude 0.3

zapashcanon:
Having `prelude >= 0.3` will force the right smtml to be picked

PR #391, opened 2024-07-31
fix printing of names (used in data section for instance)

### LINKED ISSUE #37: data string init pretty printer
```ocaml
Format.fprintf fmt {|(data %a %a %S)|} id_opt d.id data_mode d.mode d.init
```
(types.ml - line 950)
Using `%S` tag to display the init string of the data instruction could produce errors.
For example, `%S` escapes special characters : check Issue #36

zapashcanon:
Should fix #37.

I still have to add some tests and to run the fuzzer again.

### DIFF
--- a/src/ast/text.ml
+++ b/src/ast/text.ml
@@ -38,7 +38,7 @@ type data =
   }
 
 let pp_data fmt (d : data) =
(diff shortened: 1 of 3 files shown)


zapashcanon:
Some cases were also missing in the reference interpreter: https://github.com/WebAssembly/spec/pull/1772

PR #392, opened 2024-08-01
Add flags `--fail-on-trap-only` and `fail-on-assertion-only`

### LINKED ISSUE #273: add an option to stop looking for a model only if an assertion failure is found but not a Wasm trap

filipeom:
Closes #273 for symbolic execution only. I'll rerun testcomp benchmarks just to see if the issue is thoroughly fixed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ## unreleased
 
+- adds flags `--fail-on-trap-only` and `fail-on-assertion-only`
 - parameterize the `Thread` module on the symbolic memory and the `Choice_monad` module on a Thread
 - adds a `owi_char` function to create char symbolic value
 - adds a `Mem` argument to external function to allow direct manipulation of the memory.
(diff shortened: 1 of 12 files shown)


zapashcanon:
Thanks!

PR #395, opened 2024-08-02
Add output dir and table in zulip notification

filipeom:
Preview:

image

### DIFF
--- a/bench/report/runs.ml
+++ b/bench/report/runs.ml
@@ -108,6 +108,31 @@ let pp_quick_results fmt results =
     "Nothing: %6i    Reached: %6i    Timeout: %6i    Other: %6i    Killed: %6i"
     !nothing !reached !timeout !other !killed
 
+let pp_table_results fmt results =
+  let nothing = count_nothing results in
+  let reached = count_reached results in
+  let timeout = count_timeout results in
+  let other = count_other results in
+  let killed = count_killed results in
+  let total = count_all results in
+  Format.fprintf fmt
+    "| Nothing | Reached | Timeout | Other | Killed | Total |@\n\
+     |:-------:|:-------:|:-------:|:-----:|:------:|:-----:|@\n\
+     | %6i | %6i | %6i | %6i | %6i | %6i |" nothing reached timeout other killed
+    total
+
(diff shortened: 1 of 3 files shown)


zapashcanon:
Great! Thanks! :)

bench/report/runs.ml
```diff
@@ -100,6 +100,26 @@ let pp_quick_results fmt results =
...
+      | Nothing _ -> incr nothing
+      | Reached _ -> incr reached
+      | Timeout _ -> incr timeout
+      | Signaled _ | Stopped _ -> incr killed
+      | Other _ -> incr other )
+    results;
+  Format.fprintf fmt
+    "| Nothing | Reached | Timeout | Other | Killed |@\n\
+     |:-------:|:-------:|:-------:|:-----:|:------:|@\n\
+     | %6i | %6i | %6i | %6i | %6i |" !nothing !reached !timeout !other !killed
```
zapashcanon, line 121: Could you use `Runs.count_nothing results` and the like instead of recomputing them here ?

Also, should we add tables to display also all the clock/utime/stime statistics ? See `runs.mli` to get the list of available functions. :)

zapashcanon: In `quick_results` it was done with references because we call it after all tasks so I did not want to make it slow but here it probably does not matter. (And actually it probably doesn't matter in `quick_results` either...`

filipeom: > Could you use Runs.count_nothing results and the like instead of recomputing them here ?

Ofc! :)

> Also, should we add tables to display also all the clock/utime/stime statistics ? See runs.mli to get the list of available functions. :)

Good idea! Will add this also :)

zapashcanon: And for the time statistics, I guess we want to compute them on `reached` tasks only, right?

filipeom: Hum idk, that might skew the results in favour of the tasks we perform very well? What do you think?

zapashcanon: Well, the timeout tasks are all going to take 30 sec. and are not very interesting for wall clock time. But I don't mind.

PR #407, opened 2024-08-13
Allow writing function contracts in webassembly code

Laplace-Demon:


Laplace-Demon:
`(module
    (@contract $start
        (ensures (= result (+ (param 0) 3)))
    )
    (func $plus_three
        (param $x i32) (result i32)
        (i32.add (i32.const 3) (local.get $x)))
    (func $f1)
    (func $f2)
    (func $start
        (call $plus_three (i32.const 42))
        drop)
    (start $start)
)`

src/annot/contract.ml
```diff
@@ -0,0 +1,44 @@
...
+let cons_second (l1, l2) x2 = (l1, x2 :: l2)
+
+let parse_contract =
+  let open Sexp in
+  function
+  | List (Atom funcid :: conds) ->
+    let aux acc = function
+      | List [ Atom "requires"; precond ] ->
+        let+ precond = parse_prop precond in
+        cons_first acc precond
```
zapashcanon, line 35: You can write:

```ocaml
    let aux (l1, l2) = function
      | List [ Atom "requires"; precond ] ->
        let+ precond = parse_prop precond in
        (precond :: l1), l2
```

Instead of defining auxiliary functions.

src/annot/contract.ml
```diff
@@ -0,0 +1,44 @@
...
+      | List [ Atom "ensures"; postcond ] ->
+        let+ postcond = parse_prop postcond in
+        cons_second acc postcond
+      | _ as s -> Error (`Unknown_annotation_clause s)
```
zapashcanon, line 39: `| _ as s` is the same as `| s` :-)

src/annot/contract.ml
```diff
@@ -0,0 +1,44 @@
...
+    in
+    let* funcid = parse_indice funcid in
+    let+ preconditions, postconditions = list_fold_left aux ([], []) conds in
+    { funcid; preconditions; postconditions }
+  | _ as s -> Error (`Unknown_annotation_object s)
```
zapashcanon, line 44: `| s`

src/annot/spec.ml
```diff
@@ -0,0 +1,378 @@
...
+  | Result : 'a term
+
+type 'a prop =
+  | Const : bool -> 'a prop
+  | BinPred : binpred * 'a term * 'a term -> 'a prop
+  | UnConnect : unconnect * 'a prop -> 'a prop
+  | BinConnect : binconnect * 'a prop * 'a prop -> 'a prop
+  | Binder : binder * binder_type * string option * 'a prop -> 'a prop
+
+let pp_bool fmt = function true -> pf fmt "true" | false -> pf fmt "false"
```
zapashcanon, line 117: You can use `Fmt.bool` instead. :)

src/annot/spec.ml
```diff
@@ -0,0 +1,378 @@
...
+
+let pp_binpred fmt = function
+  | Ge -> pf fmt "≥"
+  | Gt -> pf fmt ">"
+  | Le -> pf fmt "≤"
+  | Lt -> pf fmt "<"
+  | Eq -> pf fmt "="
+  | Neq -> pf fmt "≠"
```
zapashcanon, line 125: I'm fine about using the nicer `≠`, `≥` etc. But you'll have to make sure that we can parse them also. Is it the case already?

src/annot/spec.ml
```diff
@@ -0,0 +1,378 @@
...
+
+let parse_text_id id =
+  try
+    let len = String.length id in
+    let hd = String.get id 0 in
+    let tl = String.sub id 1 (len - 1) in
+    if Char.equal hd '$' && String.for_all valid_text_indice_char id then
+      Some tl
+    else None
+  with Invalid_argument _ -> None
```
zapashcanon, line 198: You should never catch `Invalid_argument _` when writing OCaml code. `Invalid_argument` denotes a programming error and you should ensure before calling the function that it won't happen. This is really needed because when compiling with `-unsafe`, such exception is not going to be raised and the code will segfault or return a garbage value...

src/annot/spec.ml
```diff
@@ -0,0 +1,378 @@
...
+  try
+    let len = String.length id in
+    let hd = String.get id 0 in
+    let tl = String.sub id 1 (len - 1) in
+    if Char.equal hd '$' && String.for_all valid_text_indice_char id then Ok tl
+    else Error (`Invalid_text_indice id)
+  with Invalid_argument _ -> Error (`Invalid_text_indice id)
+
+let parse_raw_id id =
+  match int_of_string id with Some id -> Some id | None -> None
```
zapashcanon, line 210: This is simply `int_of_string id` right ? You can remove the whole match stuff.

src/annot/spec.ml
```diff
@@ -0,0 +1,378 @@
...
+  with Invalid_argument _ -> None
+
+let parse_text_id_result id =
+  try
+    let len = String.length id in
+    let hd = String.get id 0 in
+    let tl = String.sub id 1 (len - 1) in
+    if Char.equal hd '$' && String.for_all valid_text_indice_char id then Ok tl
+    else Error (`Invalid_text_indice id)
+  with Invalid_argument _ -> Error (`Invalid_text_indice id)
```
zapashcanon, line 207: Same.

src/annot/contract.ml
Laplace-Demon: thanks!

src/annot/spec.ml
```diff
@@ -0,0 +1,378 @@
...
+  | _, _ -> Error (`Invalid_indice id)
+
+let parse_binder_type =
+  let open Sexp in
+  function
+  | Atom "i32" -> ok I32
+  | Atom "i64" -> ok I64
+  | Atom "f32" -> ok F32
+  | Atom "f64" -> ok F64
+  | _ as bt -> Error (`Unknown_binder_type bt)
```
zapashcanon, line 230: ` | bt`

src/annot/spec.ml
```diff
@@ -0,0 +1,378 @@
...
+  | _, Some id -> Ok (Raw id)
+  | _, _ -> Error (`Invalid_indice id)
+
+let parse_binder_type =
+  let open Sexp in
+  function
+  | Atom "i32" -> ok I32
+  | Atom "i64" -> ok I64
+  | Atom "f32" -> ok F32
+  | Atom "f64" -> ok F64
```
zapashcanon, line 229: Why not `Ok` ?

src/annot/spec.ml
```diff
@@ -0,0 +1,378 @@
...
+    let+ tm2 = parse_term tm2 in
+    BinOp (Div, tm1, tm2)
+  | List [ Atom c; tm1; tm2 ] ->
+    let* tm1 = parse_term tm1 in
+    let+ tm2 = parse_term tm2 in
+    BinOp (CustomBinOp c, tm1, tm2)
+  (* Result *)
+  | Atom "result" -> ok Result
+  (* Invalid *)
+  | _ as tm -> Error (`Unknown_term tm)
```
zapashcanon, line 304: `| tm`

src/annot/spec.ml
Laplace-Demon: No, I'm currently using `!=`, `>=` for syntax. I'm not sure we can parse `≠`, `≥`, etc. I'll try it

src/annot/spec.ml
Laplace-Demon: oops

src/annot/spec.ml
```diff
@@ -0,0 +1,378 @@
...
+  | Atom "f32" -> ok F32
+  | Atom "f64" -> ok F64
+  | _ as bt -> Error (`Unknown_binder_type bt)
+
+let rec parse_term =
+  let open Sexp in
+  function
+  (* Int32 *)
+  | Atom i32 when Option.is_some (Int32.of_string_opt i32) ->
+    ok @@ Int32 (Int32.of_string i32)
```
zapashcanon, line 234: I'm not fond of doing two times the `of_string` conversion and having the two `Atom` cases in different places. What about something like:

```ocaml
| Atom a as tm -> begin
  match Int32.of_string_opt a with
  | Some i32 -> Ok (Int32 i32)
  | None ->
  match parse_tex_id a with
  | Some ind -> Ok (Var ind)
  | None ->
  if String.equal "result" a then Ok Result
  else Error (`Unknown_term tm)
end
```

?

src/annot/spec.ml
Laplace-Demon: Ok thanks!

src/parser/text_lexer.ml
```diff
@@ -105,7 +105,7 @@ let id_char =
     | 'a' .. 'z'
     | 'A' .. 'Z'
     | '!' | '#' | '$' | '%' | '&' | '\'' | '*' | '+' | '-' | '.' | '/' | ':'
-    | '<' | '=' | '>' | '?' | '@' | '\\' | '^' | '_' | '`' | '|' | '~' | '*' )]
+    | '<' | '=' | '>' | '?' | '@' | '\\' | '^' | '_' | '`' | '|' | '~' )]
```
zapashcanon, line 108: Why are you removing this case? Are the tests still passing while removing it? I believe it is allowed to write `(func $f*f (param i32) ...` and now it is not anymore?

zapashcanon: Oh this is because it was duplicated, right?

src/annot/spec.ml
Laplace-Demon: I think it's good idea to group `Atom` cases together. It's clearer

src/parser/text_parser.mly
```diff
@@ -992,7 +992,25 @@ let inline_module_inner ==
...
+      match (list_map
+        (fun t ->
+          match t with
+          | Annot t ->
+            let+ c = parse_contract t.items in
+            Contract c
+          | _ -> Error (`Invalid_indice ""))
+      annots) with
+      | Ok annots -> annots
+      | _ -> []
```
zapashcanon, line 1009: I'm not sure we want to ignore the error here.

src/parser/text_lexer.ml
Laplace-Demon: It seems that we already have an `'*'` in the pattern (line 107)

src/primitives/float32.mli
```diff
@@ -62,6 +62,8 @@ val ge : t -> t -> bool
 
 val of_string : string -> t
 
+val of_string_opt : string -> t option
+
```
zapashcanon, line 66: While you're at it, could you rename `of_string` to `of_string_exn` and rename `of_string_opt` to `of_string` ? You'll also have to change it in `float64`, `int32` and `int64` but I expect it to be quite straightforward.

src/text_to_binary/rewrite.ml
```diff
@@ -407,6 +407,131 @@ let rewrite_types (_modul : Assigned.t) (t : binary str_type) :
   let t = [ (None, (Final, [], t)) ] in
   Ok t
 
+let rec rewrite_term (binder_list : string option list)
+  ?(modul : Binary.modul = Binary.empty_modul)
+  ?(func_param_list : string option list = []) :
```
zapashcanon, line 412: Could you use named parameters instead of optional ones here ? This piece of code is already complicated and it's too easy to do a mistake with optional parameters IMO...

src/text_to_binary/rewrite.ml
```diff
@@ -407,6 +407,131 @@ let rewrite_types (_modul : Assigned.t) (t : binary str_type) :
...
+  text Spec.term -> binary Spec.term Result.t =
+  let rec aux error acc id = function
+    | [] -> Error error
+    | Some id' :: bl ->
+      if String.equal id id' then Ok (Raw acc) else aux error (acc + 1) id bl
+    | _ :: bl -> aux error (acc + 1) id bl
```
zapashcanon, line 418: `| None :: bl`

src/text_to_binary/rewrite.ml
```diff
@@ -407,6 +407,131 @@ let rewrite_types (_modul : Assigned.t) (t : binary str_type) :
   let t = [ (None, (Final, [], t)) ] in
   Ok t
 
+let rec rewrite_term (binder_list : string option list)
+  ?(modul : Binary.modul = Binary.empty_modul)
+  ?(func_param_list : string option list = []) :
+  text Spec.term -> binary Spec.term Result.t =
+  let rec aux error acc id = function
```
zapashcanon, line 414: Could you find a better name than `aux` ?

src/parser/text_lexer.ml
Laplace-Demon: > Oh this is because it was duplicated, right?

Yes :) ocaml-lsp-server seems unable to detect it probably because it's in the extension node `[%sedlex.regexp?]`? When I copied this pattern to an OCaml function ocaml-lsp-server complained

src/text_to_binary/rewrite.ml
```diff
@@ -407,6 +407,131 @@ let rewrite_types (_modul : Assigned.t) (t : binary str_type) :
...
+  function
+  | Int32 i32 -> Ok (Int32 i32)
+  | Int64 i64 -> Ok (Int64 i64)
+  | Float32 f32 -> Ok (Float32 f32)
+  | Float64 f64 -> Ok (Float64 f64)
+  | Var ind -> (
+    match
+      ( find_binder binder_list ind
+      , find_param func_param_list ind
+      , find_global modul ind )
```
zapashcanon, line 455: OK we'll have to think about this more. This seems reasonable but it may be a little bit dangerous too...

Laplace-Demon: In what sense? efficiency?

src/text_to_binary/rewrite.ml
```diff
@@ -407,6 +407,131 @@ let rewrite_types (_modul : Assigned.t) (t : binary str_type) :
...
+    UnOp (u, tm1)
+  | BinOp (b, tm1, tm2) ->
+    let* tm1 = rewrite_term binder_list ~modul ~func_param_list tm1 in
+    let+ tm2 = rewrite_term binder_list ~modul ~func_param_list tm2 in
+    BinOp (b, tm1, tm2)
+  | Result -> Ok Result
+
+let rec rewrite_prop (binder_list : string option list)
+  ?(modul : Binary.modul = Binary.empty_modul)
+  ?(func_param_list : string option list = []) :
```
zapashcanon, line 481: Same.

test/conc/plus.wat
```diff
@@ -0,0 +1,8 @@
+(module
```
zapashcanon, line 1: Do not put this in `test/conc/` either in `test/sym/contract/` or in `test/weasl/` for instance (For WEbAssembly Specification Language hahaha) but if you have any other idea, go with it, it's your tool :-)

test/script/annotations.t
```diff
@@ -0,0 +1 @@
+  $ owi script --no-exhaustion reference/proposals/annotations/annotations.wast
```
zapashcanon, line 1: Could you put it in `reference.t` (and `reference_opt.t` too) instead?

test/conc/plus.wat
```diff
@@ -0,0 +1,8 @@
...
+    (@contract $plus_three
+        (ensures (= result (+ $x 3)))
+    )
+    (func $plus_three
+        (param $x i32) (result i32)
+        (i32.add (i32.const 3) (local.get $x)))
```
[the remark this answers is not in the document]
Laplace-Demon: Oh yes, I mis-added this file

src/annot/spec.ml
zapashcanon: To be clear, I mean that we should allow both when parsing, but use the nicer one when printing.

src/primitives/float32.mli
Laplace-Demon: Ok, it's fixed

src/annot/spec.ml
Laplace-Demon: fixed

src/parser/text_parser.mly
Laplace-Demon: I had a look at `cmd_sym.ml` and `compile.ml`. I don't want to break into their structure too much so for now I think I'll always parse contract and generate code, and emit potential errors.

src/ast/code_generator.ml
```diff
@@ -0,0 +1,585 @@
...
+(* Written by the Owi programmers *)
+
+open Types
+open Binary
+open Spec
+open Syntax
+
+let type_env (m : modul) (func_ty : binary param_type * binary result_type)
+  (owi_funcs : (string * int) list) =
+  object
```
zapashcanon, line 12: Why did you decide to go with an object ?

Laplace-Demon: I need to pass in the types of global variables, function parameters and results, symbolic execution functions and quantified variables to generate correct code, and I don't want to complicate the type. So I think maybe an object helps me to abstract away all these?

zapashcanon: Would you like to try with modules instead of objects ? There is currently no objects in the codebase (there's object *types* only) and I would like to avoid them if possible. :sweat_smile:

zapashcanon: I also expect a record to be enough there.

src/ast/code_generator.ml
```diff
@@ -199,12 +176,12 @@ let rec term_generate tenv term : (binary expr * binary val_type) Result.t =
     let* expr2, ty2 = term_generate tenv tm2 in
     binop_generate b expr1 ty1 expr2 ty2
   | Result (Some i) -> (
-    match tenv#get_result_type i with
-    | Some t -> Ok ([ Local_get (Raw (tenv#get_param_number + i)) ], t)
+    match List.nth_opt tenv.result_types i with
```
zapashcanon, line 227: If you're using `nth{_opt}` then you should probably replace this type by an array.

src/ast/code_generator.ml
```diff
@@ -21,25 +21,67 @@ type type_env =
...
-  (owi_funcs : (string * int) list) : type_env =
+  (owi_funcs : (string * int) array) : type_env =
   let param_types = List.map snd (fst func_ty) in
-  let global_types (* would get simplified with #353 *) =
+  let global_types =
     List.map
-      (fun (x : (global, binary global_type) Runtime.t Indexed.t) ->
-        match Indexed.get x with
+      (fun (x : (global, binary global_type) Runtime.t) ->
+        match x with
```
zapashcanon, line 31: `function` should now be enough instead of `fun c -> match x with ...`

src/ast/code_generator.ml
```diff
@@ -390,23 +404,27 @@ let subst_index ?(subst_custom = false) (old_index : int) (index : int)
...
-      | None -> Ok (Fmt.str "func_%i" old_index, type_f) )
-    | Some (Imported { modul; name; assigned_name; desc }) -> (
-      match assigned_name with
-      | Some assigned_name -> Ok (assigned_name, desc)
-      | None -> Ok (Fmt.str "func_%s_%s_%i" modul name old_index, desc) )
-    | None -> Error (`Contract_unknown_func (Raw old_index))
+    if old_index >= func_num || old_index < 0 then
+      Error (`Contract_unknown_func (Raw old_index))
+    else
+      match Array.get func old_index with
```
zapashcanon, line 431: `match func.(old_index) with`

src/ast/code_generator.ml
```diff
@@ -454,111 +469,88 @@ let contracts_generate (owi_funcs : (string * int) list) (m : modul)
...
-      List.partition_map
-        (fun i ->
-          let v = Indexed.get i in
-          match v with
-          | Runtime.Imported _ -> Either.Left (Indexed.get_index i, v)
-          | Local _ -> Either.Right (Indexed.get_index i, v) )
-        m.func.values
+      let i =
+        Option.fold ~none:func_num
+          ~some:(fun x -> x)
```
zapashcanon, line 535: `~some:Fun.id`

src/ast/code_generator.ml
Laplace-Demon: In the definition of binary module fields, record field names are duplicated like:

```ocaml
type exports =
  { global : export list
  ; mem : export list
  ; table : export list
  ; func : export list
  }

type global =
  { typ : binary global_type (* TODO: init : binary+const expr*)
  ; init : binary expr
  ; id : string option
  }

...

type data =
  { id : string option
  ; init : string
  ; mode : data_mode
  }

...

type elem =
  { id : string option
  ; typ : binary ref_type (* TODO: init : binary+const expr*)
  ; init : binary expr list
  ; mode : elem_mode
  }
```

So without the type annotation OCaml cannot infer the type

src/text_to_binary/rewrite.ml
```diff
@@ -481,22 +481,31 @@ let rec rewrite_prop ~(binder_list : string option list) ~(modul : Binary.modul)
...
+    | Local { type_f; _ } -> type_f
+    | Imported { desc; _ } -> desc
+  in
+  let* func_param_list =
+    match func_bt with
+    | Bt_ind ind -> (
+      let (Raw i) = find modul.typ ind in
+      match Indexed.get_at i modul.typ.values with
+      | Some (Def_func_t (func_pt, _)) ->
+        Ok (List.map (fun (str_opt, _) -> str_opt) func_pt)
```
zapashcanon, line 505: `List.map fst func_pt`

src/text_to_binary/rewrite.ml
```diff
@@ -481,22 +481,31 @@ let rec rewrite_prop ~(binder_list : string option list) ~(modul : Binary.modul)
...
+      | _ -> Error (`Spec_invalid_indice (Int.to_string i)) )
+    | Bt_raw (_, (func_pt, _)) ->
+      Ok (List.map (fun (str_opt, _) -> str_opt) func_pt)
```
zapashcanon, line 508: same

zapashcanon:
I don't understand your `add --add-assert-to-assume option`, I believe this is something that should be done at the symbolic execution implementation level directly, you should not try to emulate it in `rac`.

src/ast/code_generator.ml
Laplace-Demon: No, I think in `List.map f l` OCaml determines the (erroneous) type of `f` first then uses that to check `l`. So it thinks `m.global` should have type of `(elem, (_ * binary heap_type)) Runtime.t array`

Laplace-Demon:
> I don't understand your `add --add-assert-to-assume option`, I believe this is something that should be done at the symbolic execution implementation level directly, you should not try to emulate it in `rac`.

Okay ! When I finished I got complaint that `owi_assume` is never used, that made me think of `--add-assert-to-assume` might be something to add. I'll just mark it with `_owi_assume` instead

src/ast/code_generator.ml
zapashcanon: OK fair enough.

zapashcanon:
> > I don't understand your `add --add-assert-to-assume option`, I believe this is something that should be done at the symbolic execution implementation level directly, you should not try to emulate it in `rac`.
> 
> Okay ! When I finished I got complaint that `owi_assume` is never used, which made me think that `--add-assert-to-assume` might be something to add. I'll just mark it with `_owi_assume` instead

OK, you can also simply remove it if it is not needed for now.

src/ast/code_generator.ml
Laplace-Demon: Here (along with 3 other `List.nth_opt`s), the index may be out of bound. Is it okay to explicitly catch the `Invalid_argument` exception when using array?

Laplace-Demon: Oh maybe check the bound first, I'll change it for an array

zapashcanon:
This needs to be rebased

src/bin/owi.ml
```diff
@@ -179,6 +179,15 @@ let opt_cmd =
   in
   Cmd.v info Term.(const Cmd_opt.cmd $ debug $ unsafe $ files)
 
+let rac_cmd =
+  let open Cmdliner in
+  let info =
+    let doc = "Perform runtime assertion checking" in
```
zapashcanon, line 185: I believe it should be something like `Generate an instrumented file with runtime assertion checking coming from Weasel specifications.` ?

zapashcanon:
So I believe we should add some stuff:

- being able to generate an instrumented version that does not depend on symbolic execution (let's call it `rac` for `runtime assertion checking`);
- being able to generate an instrumented version that does depend on symbolic execution (let's call it `srac`);
- add an option `--rac` to `owi run` that would perform the checks;
- add options `--rac` and `--srac` to `owi sym` and `owi conc`;
- add a `--symbolic` option (or maybe `--srac` to be more consistent?) to `owi rac` that would generate the one that depends on symbolic execution.


Moreover, what do you think about renaming `owi rac` into `owi instrument` ?

Laplace-Demon:
> Moreover, what do you think about renaming `owi rac` into `owi instrument` ?

For the name, `rac` is used temporarily and not the most suitable, `instrument` is better

src/ast/code_generator.ml
```diff
@@ -745,8 +678,19 @@ let add_owi_funcs (m : modul) : modul * (string * int) array =
...
-let generate (m : modul) : modul Result.t =
-  let m, owi_funcs = add_owi_funcs m in
+let generate (symbolic : bool) (m : modul) : modul Result.t =
+  let owi_funcs =
+    if symbolic then
+      [| ("i32_symbol", ([], [ Num_type I32 ]))
+       ; ("i64_symbol", ([], [ Num_type I64 ]))
+       ; ("f32_symbol", ([], [ Num_type F32 ]))
+       ; ("f64_symbol", ([], [ Num_type F64 ]))
+       ; ("assume", ([ (None, Num_type I32) ], []))
```
zapashcanon, line 688: Are all these functions actually required by the generated code?

zapashcanon: OK so, they can probably removed. We'll re-introduce them if we start generating more complicated RAC stuff.

Laplace-Demon: ok

zapashcanon:
Could you rebase so I can merge? :)

Laplace-Demon:
> Could you rebase so I can merge? :)

I'll rebase later today, to check if there are something left :)

zapashcanon:
Thanks! I think you can add something in `CHANGES.md`. Also, could you add an example as you did for E-ACSL?

zapashcanon:
Perfect! Could you add a small test with a `.wat` file containing an `assert` and call it with `owi run --rac` to simply check that it works correctly ?

zapashcanon:
One more thing I am thinking of, instead of requiring an assert function from the host in the concrete case, we could simply compile the code to `if (!cond) then unreachable`, right?

zapashcanon:
@Laplace-Demon, I wanted to merge this but it seems there's an issue with some tests from the custom-annotation proposal (see the logs in the CI). Do you want to have a look before I merge or do you prefer to fix it later?

Laplace-Demon:
> @Laplace-Demon, I wanted to merge this but it seems there's an issue with some tests from the custom-annotation proposal (see the logs in the CI). Do you want to have a look before I merge or do you prefer to fix it later?

Yes, I think we've already discussed that during my internship and all the failed tests are due to error messages different from that of the reference interpreter.

Since there are no false positives (and false negatives), I think it's okay to merge it for now. Though a fix does not amount to simply changing our error messages, but would probably require some structural change of our parser and lexer.

zapashcanon:
Oh, OK sorry, I forget about that. But I remember now. I'm merging, thanks! :)

PR #413, opened 2024-08-21
add `--emit-file` option to emit .wat files

vasucp1207:
Issue https://github.com/OCamlPro/owi/issues/247

Add --emit-flag flag to emit file.wat from file.wasm


src/cmd/cmd_wasm2wat.ml
```diff
@@ -4,13 +4,19 @@
 
 open Syntax
 
-let cmd_one file =
+(* Utility function to handle writing to a file or printing to stdout *)
```
zapashcanon, line 7: If you want this to be an item comment for the function, you should remove the empty line between the comment and the `let cmd_one ...` line.

Also, you must replace `(*` by `(**`. see the [manual](https://ocaml.org/manual/5.2/doccomments.html#ss:item-comments).

src/cmd/cmd_wasm2wat.ml
```diff
@@ -4,13 +4,19 @@
...
+  let wat_file = Fpath.set_ext "wat" file in
+
   match ext with
   | ".wasm" ->
     let* m = Parse.Binary.Module.from_file file in
     let m = Binary_to_text.modul m in
-    Ok (Fmt.pr "%a@\n" Text.pp_modul m)
+    let wat_module = Fmt.str "%a@\n" Text.pp_modul m in
+    if emitfile then Bos.OS.File.write wat_file wat_module
+    else Ok (Fmt.pr "%s" wat_module)
```
zapashcanon, line 19: It probably does not matter in practice, but it would be a little bit more efficient to avoid allocating a string:

```ocaml
let m = Binary_to_text.modul m in
if emitfile then Bos.OS.File.writef wat_file "%a@\n" Text.pp_modul m
else Fmt.str "%a@\n" Text.pp_modul m;
Ok ()
```

(It's untested so there may be a type error in the call to `writef` :sweat_smile:)

zapashcanon:
Hi!

Welcome to the project, and thanks for the PR! :)

I added some minor comments. Maybe you could also add an entry in `CHANGES.md` ?

vasucp1207:
> Hi!
> 
> Welcome to the project, and thanks for the PR! :)
> 
> I added some minor comments. Maybe you could also add an entry in `CHANGES.md` ?
> 
> I have two other questions, it's a little bit unusual for `-o` to be a simple flag, one usually expects something like `-o outputname.wat`. We could either rename `-o` to `--emit-file` or change `-o` from a flag to a parameter expecting an actual file name. WDYT ?

Because `wasm2wat` can take multiple files as arguments, I think it's not a good idea to change `-o` flag to a parameter that expects a file name, `--emit-file` is better because it can handle multiple files. Correct me, If I am missing something.

> 
> Could you add a small test in `test/wasm2wat/` ? You would have to create a file named `emit.t` for instance, which would have the following content:
> 
> ```
> test that we can emit to a file:
>   $ owi wasm2wat m.wasm --emit-file
>   $ cat m.wat
> ```
> 
> Then you would run `dune runtest` (it may take a while). You'll then see some output, if it looks correct, you should run `dune promote` and commit the changes.
> 
> You'll need to run the `dune runtest`+`dune promote` procedure anyway in order to update the manual pages that are making the CI fail. :)

zapashcanon:
> Because wasm2wat can take multiple files as arguments, I think it's not a good idea to change -o flag to a parameter that expects a file name, --emit-file is better because it can handle multiple files. Correct me, If I am missing something.

Indeed, I forgot about it being able to handle multiple files... `--emit-file` is better indeed.

zapashcanon:
Thanks, the tests look good now. I'm going to merge once the two reviews about [src/cmd/cmd_wasm2wat.ml](https://github.com/OCamlPro/owi/pull/413/files/0e7af4f664b53fd125d32580a15a1896aa4bfb17#diff-67f119f295484f31ca29c9cf18af6d241b83a5dce2ada975ccf97bf80748712f) are addressed.

src/cmd/cmd_wasm2wat.ml
vasucp1207: Doing like this give an error,
```ocaml
else Fmt.str "%a@\n" Text.pp_modul m;
Ok ()
```

This works fine,
```ocaml
else Ok (Fmt.pr "%a@\n" Text.pp_modul m)
```

zapashcanon: Aw OK, I guess `Bos.OS.File.writef` already returns a result ? That's fine with me. :)

zapashcanon: @vasucp1207, please ping me once this is done so I can merge. Otherwise I'll forget.

vasucp1207: @zapashcanon, do you want me to make any changes🤔? It works correctly.

zapashcanon: Oh sorry, for some reasons I overlooked the changes. This is good to go now, thanks! :D

zapashcanon:
Thanks!

PR #416, opened 2024-08-23
fix: add `--output` to `wat2wasm`, `wasm2wat` and `opt` cmd

vasucp1207:
Emit wasm files in same dir as wat files


src/ast/binary_encoder.ml
```diff
@@ -763,7 +763,8 @@ let write_file filename content =
   let _dir, filename = Fpath.split_base filename in
   let filename, _ext = Fpath.split_ext filename in
   let filename = Fpath.add_ext ".wasm" filename in
-  let filename = Fpath.to_string filename in
+  let fullpath = Fpath.add_seg _dir (Fpath.to_string filename) in
+  let filename = Fpath.to_string fullpath in
```
zapashcanon, line 767: I think you can do even simpler by replacing the whole:

```ocaml
  let _dir, filename = Fpath.split_base filename in
  let filename, _ext = Fpath.split_ext filename in
  let filename = Fpath.add_ext ".wasm" filename in
  let fullpath = Fpath.add_seg _dir (Fpath.to_string filename) in
  let filename = Fpath.to_string fullpath in
```

By:

```ocaml
  let filename = Fpath.set_ext ".wasm" filename in
```

zapashcanon:
Also, I'm not sure we really want this behaviour, I checked and the `wat2wasm` command from `wabt` actually outputs the file in the current folder. @filipeom, @krtab WDYT?

vasucp1207:
> Also, I'm not sure we really want this behaviour, I checked and the `wat2wasm` command from `wabt` actually outputs the file in the current folder. @filipeom, @krtab WDYT?

Yes, you are right, because the `wasm2wat` emit the files in the same folder that's why I did this.

zapashcanon:
> > Also, I'm not sure we really want this behaviour, I checked and the `wat2wasm` command from `wabt` actually outputs the file in the current folder. @filipeom, @krtab WDYT?
> 
> Yes, you are right, because the `wasm2wat` emit the files in the same folder that's why I did this.

Aw OK. Maybe we could have a flag then, something like `--emit-file-in-current-directory` for the same behavior as wat2wasm and keep the new one as default; or `--emit-file-in-same-directory` to get the new one and keep the wat2wasm one as default ?

filipeom:
> Also, I'm not sure we really want this behaviour, I checked and the wat2wasm command from wabt actually outputs the file in the current folder. @filipeom, @krtab WDYT?

I understand why this PR is needed, but I'm not entirely in favor of it. I believe both `--emit-file-in-current-directory` and `--emit-file-in-same-directory` are adequate solutions.

However, I must confess that I prefer the `wat2wasm` approach: passing in one file and exporting it to another using `-o`. When I need to process multiple files, I can simply iterate through them in a shell loop. I think accepting a list of files as input is less flexible, as there might be use cases where someone wants each file to be exported to a different destination.

I'll leave the final decision to someone with a stronger opinion.

vasucp1207:
> > Also, I'm not sure we really want this behaviour, I checked and the wat2wasm command from wabt actually outputs the file in the current folder. @filipeom, @krtab WDYT?
> 
> I understand why this PR is needed, but I'm not entirely in favor of it. I believe both `--emit-file-in-current-directory` and `--emit-file-in-same-directory` are adequate solutions.
> 
> However, I must confess that I prefer the `wat2wasm` approach: passing in one file and exporting it to another using `-o`. When I need to process multiple files, I can simply iterate through them in a shell loop. I think accepting a list of files as input is less flexible, as there might be use cases where someone wants each file to be exported to a different destination.
> 
> I'll leave the final decision to someone with a stronger opinion.

I agree that accepting a single file as an argument is a good idea for specifying the target path. let `wat2wasm` be the default approach and I suppose there is no need for these type of flags `--emit-file-in-same-directory`.

zapashcanon:
@vasucp1207, do you want to make the changes to accepting a single file, to use the current directory and to accept the `-o` option? :)

vasucp1207:
> @vasucp1207, do you want to make the changes to accepting a single file, to use the current directory and to accept the `-o` option? :)

Yup, `wabt` also accepts a single file.

zapashcanon:
Yes sure, so I propose the following changes:

- make `owi wat2wasm` and `owi wasm2wat` accept only one file;
- add a `-o FILE` option to both (we can keep `--emit-file` too for `owi wat2wasm`);
- make the default output directory be the current directory (when `-o FILE` or `--emit-file` is not present).

vasucp1207:
> * make the default output directory be the current directory (when `-o FILE` or `--emit-file` is not present).

Not include the `--emit-file` option to `wat2wasm` because it's by default emits the file anyway.

src/ast/binary_encoder.ml
```diff
@@ -759,18 +759,17 @@ let encode
...
-  let filename, _ext = Fpath.split_ext filename in
-  let filename = Fpath.add_ext ".wasm" filename in
-  let fullpath = Fpath.add_seg _dir (Fpath.to_string filename) in
-  let filename = Fpath.to_string fullpath in
+  let filename = Fpath.set_ext "wasm" filename in
+  let filename = match outfile with Some name -> Fpath.v name | None -> filename in
+  let filename = Fpath.to_string filename in
   let oc = Out_channel.open_bin filename in
   Out_channel.output_string oc content;
   Out_channel.close oc
```
zapashcanon, line 769: Could you replace:

```ocaml
  let filename = Fpath.to_string filename in
  let oc = Out_channel.open_bin filename in
  Out_channel.output_string oc content;
  Out_channel.close oc
```

By:

```ocaml
Bos.OS.File.write filename content
```

?

src/ast/binary_encoder.mli
```diff
@@ -3,7 +3,8 @@
 (* Written by the Owi programmers *)
 
 val convert :
-     Fpath.t
+     string option
```
zapashcanon, line 6: Can you replace this by an `Fpath.t option` ? This should simplify the code.

zapashcanon:
@vasucp1207, I believe we also have to do the same for `owi opt` along the way

vasucp1207:
> @vasucp1207, I believe we also have to do the same for `owi opt` along the way

Sure, but I didn't add `--emit-file` to `owi opt` because of name conflicts if the source file is in the current dir

src/ast/binary_encoder.ml
```diff
@@ -762,14 +762,15 @@ let encode
...
   let _dir, filename = Fpath.split_base filename in
   let filename = Fpath.set_ext "wasm" filename in
-  let filename = match outfile with Some name -> Fpath.v name | None -> filename in
-  let filename = Fpath.to_string filename in
-  let oc = Out_channel.open_bin filename in
-  Out_channel.output_string oc content;
-  Out_channel.close oc
+  match outfile with
+  | Some name -> Bos.OS.File.write name content
+  | None -> Bos.OS.File.write filename content
```
zapashcanon, line 767: This can be rewritten `Bos.OS.File.write (Option.value outfile ~default:filename) content` FYI :)

src/ast/binary_encoder.ml
```diff
@@ -762,14 +762,15 @@ let encode
...
-let convert (outfile: string option) (filename : Fpath.t) ~unsafe ~optimize m =
+let convert (outfile : Fpath.t option) (filename : Fpath.t) ~unsafe ~optimize m
+    =
   Log.debug0 "bin encoding ...@\n";
   let+ m = Compile.Text.until_optimize ~unsafe ~optimize m in
   let content = encode m in
-  write_file outfile filename content
+  match write_file outfile filename content with
+  | Error _ -> raise Exit
+  | Ok _ -> ()
```
zapashcanon, line 776: This code is a little bit complicated due to the `let+` and the monad.

We should not `raise Exit` like this.

Does it compile it you rewrite it like this:

```ocaml
  let* m = Compile.Text.until_optimize ~unsafe ~optimize m in
  let content = encode m in
  write_file outfile filename content
```

(note the `let+` that has been replace by a `let*` in the first line)

src/cmd/cmd_opt.ml
```diff
@@ -7,11 +7,15 @@ open Syntax
...
   if debug then Log.debug_on := true;
-  list_iter
-    (fun file ->
-      let+ m = optimize_file ~unsafe file in
-      let m = Binary_to_text.modul m in
-      Fmt.pr "%a@\n" Text.pp_modul m )
-    files
+  match print_or_emit ~unsafe file outfile with
+  | Error _ -> raise Exit
+  | Ok _ -> Ok ()
```
zapashcanon, line 21: It should work if you simply remove the match, right?

vasucp1207: No it's not, I replace this with

```ocaml
Result.bind (print_or_emit ~unsafe file outfile) (fun _ -> Ok ())
```

let me know if it's okay.

CHANGES.md
```diff
@@ -1,5 +1,6 @@
 ## unreleased
 
+- `wat2wasm` emit wasm files in same dir as wat files
```
zapashcanon, line 3: This should be updated now :)

src/bin/owi.ml
```diff
@@ -39,6 +39,21 @@ let files =
   let f = existing_non_dir_file in
   Cmdliner.Arg.(value & pos_all f [] (info [] ~doc))
 
+let sourcefile ty =
+  let doc = Fmt.str "%s source file" ty in
+  let f = existing_non_dir_file in
+  Cmdliner.Arg.(required & pos 0 (some f) None (info [] ~doc))
+
+let outfile ty =
+  let doc = Fmt.str "%s" ty in
```
zapashcanon, line 48: I believe the documentation should rather be something like `Write output to file.`

src/bin/owi.ml
```diff
@@ -39,6 +39,21 @@ let files =
   let f = existing_non_dir_file in
   Cmdliner.Arg.(value & pos_all f [] (info [] ~doc))
 
+let sourcefile ty =
+  let doc = Fmt.str "%s source file" ty in
```
zapashcanon, line 43: I believe using "Source file" or "Input file" is enough (some commands like `owi opt` accepts both kind of file).

src/cmd/cmd_opt.ml
zapashcanon: You have to replace:

```ocaml
let print_or_emit ~unsafe file outfile =
  let+ m = optimize_file ~unsafe file in
```

By:

```ocaml
let print_or_emit ~unsafe file outfile =
  let* m = optimize_file ~unsafe file in
```

Because either `writef` or your `Ok ...` are already returning a result. Basically, `let+` is like `let*` (binding in the monad) but also wraps the result in an `Ok` (map in the monad).

Once you did this, the return type of `print_or_emit` should be a proper Result (right now I believe it should be a Result wrapped in another Result...).

Then, you should be able to remove your `Result.bind` :)

zapashcanon:
We are getting there! Don't forget to run and promote the tests once the last issues have been fixed so that I can merge directly. :)

zapashcanon:
Thanks for your patience! :)

PR #417, opened 2024-08-23
Fix missing upper bound in out of bounds check

filipeom:


src/symbolic/symbolic_memory_concretizing.ml
```diff
@@ -113,11 +113,15 @@ module Backend = struct
       match Hashtbl.find m.chunks base with
       | exception Not_found -> return (Error Trap.Memory_leak_use_after_free)
       | size ->
-        let+ is_out_of_bounds = select (I32.ge_u offset size) in
+        let upper_bound = I32.add offset (const_i32 (Int32.of_int n)) in
```
zapashcanon, line 116: Calling this `upper_bound` is a little bit confusing as I was expecting it to be the bound that should not be crossed. It is actually the last element that will be written, right? What about naming it `last` (or anything better if you have an idea) ?

src/symbolic/symbolic_memory_make.ml
```diff
@@ -73,56 +73,65 @@ module Make (Backend : M) = struct
...
     | _ -> Smtml.Expr.cvtop (Ty_bitv 32) (Zero_extend 16) v
 
+  (* TODO: Smtml should do this. Make call to Expr.simplify *)
+  let try_to_remove_extract v =
+    match Smtml.Expr.view v with
+    | Extract ({ node = Concat (({ node = Ptr _; _ } as ptr), _); _ }, 8, 4) ->
+      ptr
+    | Extract ({ node = Concat (_, ({ node = Ptr _; _ } as ptr)); _ }, 4, 0) ->
+      ptr
+    | _ -> v
```
zapashcanon, line 105: I'm fine about waiting for the next Smt.ml release, so that we don't forget about this piece of code. :sweat_smile: WDYT?

src/symbolic/symbolic_memory_concretizing.ml
filipeom: Yes I agree, maybe `last_offset`?

src/symbolic/symbolic_memory_make.ml
filipeom: Fine by me, we can leave this pending for now :smiley:

zapashcanon:
I made an issue so that we don't forget.

PR #420, opened 2024-08-26
instruction-level block types declaration

Laplace-Demon:
In `binary_encoder.ml`, we acquiesce block types should not have index. These indices are removed by function `bt_some_to_raw` in the rewriting phase, which checks inline types' correspondence with the type section.

```ocaml
let write_block_type buf (typ : binary block_type option) =
  match typ with
  | None | Some (Bt_raw (None, ([], []))) -> Buffer.add_char buf '\x40'
  | Some (Bt_raw (None, ([], [ vt ]))) -> write_valtype buf vt
  | Some (Bt_raw (None, (pt, _))) -> write_paramtype buf pt
  (* TODO: memo
     will this pattern matching be enough with the use of the new modul.types field?
  *)
  | _ -> assert false (* TODO: same, new pattern matching cases ? *)
```

```ocaml
 let bt_some_to_raw : text block_type -> binary block_type Result.t = function
    | Bt_ind ind -> begin
      let+ v = get (`Unknown_type ind) modul.typ ind in
      match Indexed.get v with
      | Def_func_t t' ->
        let idx = Indexed.get_index v in
        Bt_raw (Some (Raw idx), t')
      | _ -> assert false
    end
    | Bt_raw (type_use, t) -> (
      let* t = Binary_types.convert_func_type None t in
      match type_use with
      | None -> Ok (Bt_raw (None, t))
      | Some ind ->
        (* we check that the explicit type match the type_use, we have to remove parameters names to do so *)
        let* t' =
          let+ v = get (`Unknown_type ind) modul.typ ind in
          match Indexed.get v with Def_func_t t' -> t' | _ -> assert false
        in
        let ok = Types.func_type_eq t t' in
        if not ok then Error `Inline_function_type else Ok (Bt_raw (None, t)) )
```

This would make the encoding of inline function-like block types invalid. For example :

```shell-session
$ cat block_type.wat
(module
  (func
    i32.const 1
    (block (param i32) (result i32)
      i32.const 1
      i32.add
    )
    drop
  )
  (start 0)
)
$ owi wat2wasm block_type.wat
$ owi wasm2wat block_type.wasm
     Invalid_argument("index out of bounds")
```

Here, `block (param i32) (result i32)` is encoded as `0x02` (for `block`) `0x01 0x7F` (for `param i32`, but I think it should be `0x7F` for representing the inline value type `i32`). And in reality, the first `0x01` would be parsed as type index, and the second `0x7F` parsed as the instruction `i64.div_s`, making arraying indexing out of bounds.

In contrast, function types are added in the grouping phase by `declare_func_type`, and rewrited in the rewriting phase by `rewrite_block_type`, to assign them a type index. I think we should do the same thing to inline block types.

(In the reference interpreter, block types are also added to the types section :

```shell-session
$ cat block_type.wat
(module
  (func
    i32.const 1
    (loop (param i32) (result i32)
      i32.const 1
      i32.add
    )
    drop
  )
  (start 0)
)
$ ./wasm block_type.wat -o block_type.wasm
$ ./wasm block_type.wasm -o block_type.ref.wat
$ cat block_type.ref.wat 
(module
  (type $0 (func))
  (type $1 (func (param i32) (result i32)))
  (func $0
    (type 0)
    (i32.const 1)
    (loop (type 1) (i32.const 1) (i32.add))
    (drop)
  )
  (start 0)
)
```
)

### DIFF
--- a/src/text_to_binary/grouped.ml
+++ b/src/text_to_binary/grouped.ml
@@ -106,15 +106,13 @@ let add_mem value (fields : t) (curr : curr) =
 let rec extract_block_types expr =
   let aux instr =
     match instr with
-    | Block (_str_opt, bt, expr1) ->
-      Option.to_list bt @ extract_block_types expr1
-    | Loop (_str_opt, bt, expr1) ->
+    | Block (_str_opt, bt, expr1) | Loop (_str_opt, bt, expr1) ->
       Option.to_list bt @ extract_block_types expr1
     | If_else (_str_opt, bt, expr1, expr2) ->
       Option.to_list bt @ extract_block_types expr1 @ extract_block_types expr2
-    | Return_call_indirect (_ind, bt) -> [ bt ]
-    | Return_call_ref bt -> [ bt ]
-    | Call_indirect (_ind, bt) -> [ bt ]
+    | Return_call_indirect (_, bt) | Return_call_ref bt | Call_indirect (_, bt)
+      ->
+      [ bt ]
     | _ -> []
   in
   List.concat_map aux expr
@@ -124,8 +122,7 @@ let add_func value (fields : t) (curr : curr) =
     match value with
     | Runtime.Local func ->
       let fields = declare_func_type fields func.type_f in
-      let temp = extract_block_types func.body in
-      List.fold_left declare_func_type fields temp
+      List.fold_left declare_func_type fields (extract_block_types func.body)
     | Imported func -> declare_func_type fields func.desc
   in
   let index = !(curr.func) in
--- a/src/text_to_binary/grouped.ml
+++ b/src/text_to_binary/grouped.ml
@@ -77,7 +77,7 @@ let init_curr () =
   ; data = ref 0
   }
 
-let declare_func_type type_f (fields : t) =
+let declare_func_type (fields : t) type_f =
   match type_f with
   | Bt_ind _ -> fields
   | Bt_raw (id, typ) ->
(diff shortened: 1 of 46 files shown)


src/text_to_binary/rewrite.ml
```diff
@@ -143,32 +132,50 @@ let rewrite_expr (modul : Assigned.t) (locals : binary param list)
...
       Loop (id, bt, e)
     | Block (id, bt, e) ->
-      let* bt = bt_to_raw bt in
+      let* bt =
+        match bt with
+        | Some bt ->
+          let+ bt = rewrite_block_type (typemap modul.typ) modul bt in
+          Some bt
+        | None -> Ok None
+      in
```
zapashcanon, line 163: This pattern is repeated many times, could you make a helper function to make it shorter?

src/text_to_binary/grouped.ml
```diff
@@ -103,13 +103,31 @@ let add_mem value (fields : t) (curr : curr) =
...
+    match instr with
+    | Block (_str_opt, bt, expr1) ->
+      Option.to_list bt @ extract_block_types expr1
+    | Loop (_str_opt, bt, expr1) ->
+      Option.to_list bt @ extract_block_types expr1
+    | If_else (_str_opt, bt, expr1, expr2) ->
+      Option.to_list bt @ extract_block_types expr1 @ extract_block_types expr2
+    | Return_call_indirect (_ind, bt) -> [ bt ]
+    | Return_call_ref bt -> [ bt ]
+    | Call_indirect (_ind, bt) -> [ bt ]
```
zapashcanon, line 117: Can you merge these three cases into one ? (You'll have to rename `_ind` to `_`)

zapashcanon:
Thanks a lot for fixing this! I had a quick look but it seems quite good to me. Can you add a test ? The one in your PR description looks good to me. It may be worth to add one for `loop` and other instructions ?

zapashcanon:
Thanks!

PR #421, opened 2024-08-26
Deactivate memory tracking once freed

krtab:


zapashcanon:
I'm not sure I understand this one. Could you explain when such a pointer is going to be used?

filipeom (approved):
Looks good to me!

Just cast the calls to `owi_free` to void to get rid of the import type errors. 

For example, change free in https://github.com/OCamlPro/owi/blob/aef0bd62b3dd028023a63547928f686c3438c3e5/src/libc/src/stdlib.c#L50

To 

```c
void free(void *ptr) { (void)owi_free(ptr); }
```

filipeom:
> I'm not sure I understand this one. Could you explain when such a pointer is going to be used?

The rust allocator does a bunch of offset reads with a heap chunk when deallocating stuff. So in order to avoid having a bunch of heap overflow errors when allocating things in rust we need `dealloc` to deregister the pointer from Owi's chunk table and return an untracked pointer so that the rust allocator can no nasty stuff with it. 

> Also, I'm surprised the code in src/libc/stdlib.c is not changed. Shouldn't you change realloc and free along the way ?

free and realloc can remain the same I think

zapashcanon:
I see. It would be great if we could start to test these things in Owi itself...

@krtab there is a conflict that needs to be resolved before I can merge

filipeom:
> I see. It would be great if we could start to test these things in Owi itself...

What do you suggest? I think for this case in particular it may be a bit difficult since we don't have the rust frontend atm

zapashcanon:
I'm not suggesting anything for the current PR, but yes, it would be nice to have a small Rust frontend integrated in Owi for us to test this.

zapashcanon:
Thanks!

PR #433, opened 2024-09-04
small optimisation of the lazy memory model

zapashcanon:
This should be an optimization.

I'm not sure if the compiler will properly optimize `{ data : (Int32.t, Value.int32) Hashtbl.t option ref`. It could remove one layer of indirection by "merging" the `option ref` part but I'm not sure about it (we should ask @chambart and @lthls).

The idea is the following:

- instead of creating an empty Hashtbl, we use an option;
- we create the table the first time we actually write something;
- then, when cloning, if the option is None, our children will never read our data, so we use our parent as the parent of the new children (instead of ourself, that is, the grand-parent of the children will actually be its parent)

When traversing the list of parents, this allows to switch from:

Empty -> NonEmptyA -> NonEmptyB -> Empty -> Empty -> NonEmptyC -> Empty -> NonEmptyD

To:

Empty -> NonEmptyA -> NonEmptyB -> NonEmptyC -> NonEmptyD

The invariant is that only the youngest and the oldest element of the list can be empty. Said otherwise, when looking at the whole tree, only the root node and the leaves may be empty. This should reduce allocation and allow to have less element to traverse and try in the list.

cc @filipeom 

(I would like to benchmark this on Test-Comp but also running it in B-Tree would be probably more interesting, but Idk how to run this one..)

### DIFF
--- a/src/symbolic/symbolic_memory.ml
+++ b/src/symbolic/symbolic_memory.ml
@@ -14,14 +14,11 @@ end)
 type t =
   { mutable data : Symbolic_value.int32 Map.t
   ; mutable size : Symbolic_value.int32
-  ; chunks : (Int32.t, Symbolic_value.int32) Hashtbl.t
+  ; mutable chunks : Symbolic_value.int32 Map.t
   }
 
 let create size =
-  { data = Map.empty
-  ; size = Symbolic_value.const_i32 size
-  ; chunks = Hashtbl.create 16
-  }
+  { data = Map.empty; size = Symbolic_value.const_i32 size; chunks = Map.empty }
 
 let i32 v =
   match Smtml.Expr.view v with Val (Num (I32 i)) -> i | _ -> assert false
@@ -65,8 +62,6 @@ let blit_string m str ~src ~dst ~len =
 let clone { data; chunks; size } =
   (* Caution, it is tempting not to rebuild the record here, but...
        it must be! otherwise the mutable data points to the same location *)
-  (* TODO: we can make this lazy as well *)
-  let chunks = Hashtbl.copy chunks in
   { data; chunks; size }
 
 let load_byte { data; _ } a =
@@ -184,7 +179,7 @@ let check_within_bounds m a =
   match Smtml.Expr.view a with
   | Val (Num (I32 _)) -> Ok (Symbolic_value.Bool.const false, a)
   | Ptr { base; offset } -> (
-    match Hashtbl.find_opt m.chunks base with
+    match Map.find_opt base m.chunks with
     | None -> Error Trap.Memory_leak_use_after_free
     | Some size ->
       let ptr = Int32.add base (i32 offset) in
@@ -197,11 +192,16 @@ let check_within_bounds m a =
   | _ -> assert false
 
 let free m base =
-  if not @@ Hashtbl.mem m.chunks base then
-    Fmt.failwith "Memory leak double free";
-  Hashtbl.remove m.chunks base
+  if not @@ Map.mem base m.chunks then Fmt.failwith "Memory leak double free"
+  else begin
+    let chunks = Map.remove base m.chunks in
+    m.chunks <- chunks
+  end
 
-let realloc m base size = Hashtbl.replace m.chunks base size
+let realloc m base size =
+  let chunks = Map.remove base m.chunks in
+  let chunks = Map.add base size chunks in
+  m.chunks <- chunks
 
 module ITbl = Hashtbl.Make (struct
   include Int
--- a/src/symbolic/symbolic_memory.ml
+++ b/src/symbolic/symbolic_memory.ml
@@ -2,38 +2,39 @@
 (* Copyright © 2021-2024 OCamlPro *)
 (* Written by the Owi programmers *)
 
-module Value = Symbolic_value
-module Expr = Smtml.Expr
-module Ty = Smtml.Ty
-open Expr
-
 let page_size = Symbolic_value.const_i32 65_536l
 
+module Map = Map.Make (struct
+  include Int32
+
+  (* TODO: define this in Int32 directly? *)
+  let compare i1 i2 = compare (Int32.to_int i1) (Int32.to_int i2)
+end)
+
 type t =
-  { mutable data : (Int32.t, Value.int32) Hashtbl.t option
-  ; parent : t option
-  ; mutable size : Value.int32
-  ; chunks : (Int32.t, Value.int32) Hashtbl.t
+  { mutable data : Symbolic_value.int32 Map.t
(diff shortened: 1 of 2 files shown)


lthls:
> I'm not sure if the compiler will properly optimize `{ data : (Int32.t, Value.int32) Hashtbl.t option ref`. It could remove one layer of indirection by "merging" the `option ref` part but I'm not sure about it (we should ask @chambart and @lthls).

The compiler can't optimise anything here. Is here any reason to use a `ref` rather than a mutable field ? Unless the reference is meant to be shared, the mutable field should be much better.
The indirection of the option is unfortunate; Jane Street has a proposal for `or_null` types, which provide options without indirections and could remove the last extra indirection, but it's not even implemented yet in the flambda-backend repo.

filipeom:
This looks great! 

> (I would like to benchmark this on Test-Comp but also running it in B-Tree would be probably more interesting, but Idk how to run this one..)

It's currently only available here: [wasm-se-bench](https://github.com/filipeom/wasm-se-bench). You can change [`to_run`](https://github.com/filipeom/wasm-se-bench/blob/main/runner/runner.ml#L22) to either `` `Owi `` or `` `Owi_24 `` and it should produce a `results-\today/results.csv` with the execution times. Or, I can just add the B-tree dataset to this repo if you want :)

zapashcanon:
@lthls, thanks for your answer!

> The compiler can't optimise anything here. Is here any reason to use a ref rather than a mutable field ? Unless the reference is meant to be shared, the mutable field should be much better.

It is not meant to be shared for now, but this is probably going to be required for #434. I'll switch to a mutable field for the time being. Thanks for your detailed answer! I have one more question, do you think it could be possible to implement an efficient "nullable ref" type as @¢hambart did in [nullable array](https://github.com/chambart/ocaml-nullable-array)? Or is there something fundamentally different between references and arrays that would prevent this?

> It's currently only available here: [wasm-se-bench](https://github.com/filipeom/wasm-se-bench). You can change [to_run](https://github.com/filipeom/wasm-se-bench/blob/main/runner/runner.ml#L22) to either `Owi or `Owi_24 and it should produce a results-\today/results.csv with the execution times. Or, I can just add the B-tree dataset to this repo if you want :)

What do you think about doing something like test-comp ? I.e. having the all Wasm code in a git submodule and just keep the OCaml code needed to run the benchmarks. I believe most of the OCaml code can be shared and it's nice to get an unified way to run benchmarks. I'll have a look at your `runner.ml` code and see what I can do. :) In the mean time, I'm going to run TestComp on this PR.

lthls:
> do you think it could be possible to implement an efficient "nullable ref" type as @¢hambart did in [nullable array](https://github.com/chambart/ocaml-nullable-array)? Or is there something fundamentally different between references and arrays that would prevent this?

You could literally implement it as a nullable array of size 1. You can also write a specialised version with a simple mutable record, but if you want the same guarantees as @chambart's arrays (compatibility with Marshalling) you will likely end up with something that is very similar to a nullable array of size 1.

zapashcanon:
Thanks. We don't care about Marshalling compatibility, I'll try to implement it as a record then.

filipeom:
> What do you think about doing something like test-comp ? I.e. having the all Wasm code in a git submodule and just keep the OCaml code needed to run the benchmarks. I believe most of the OCaml code can be shared and it's nice to get an unified way to run benchmarks.

I initially put the tools' comparative evaluation in the [wasm-se-bench](https://github.com/filipeom/wasm-se-bench) repo so that you wouldn’t have to deal with my spaghetti code :sweat_smile:  However, I’ve properly separated the `btree` dataset into its own [repo](https://github.com/filipeom/wasm-dataset/). I’ll create a PR to add it to the `bench` directory.

> I'll have a look at your runner.ml code and see what I can do. :) In the mean time, I'm going to run TestComp on this PR.

We can probably just reuse the current runner with a few adjustments since I based my runner on yours :sweat_smile:

zapashcanon:
Note to self: we should also try to simply use a `Map.t` instead, there would be no need to have a list anymore and it may be better in practice.

zapashcanon:
Comparison between main (tool1) and this PR (tool2):

```C
tool1 had 688 reached and tool2 had 687 reached
tool1 had 524 timeout and tool2 had 525 timeout
tool1 had 000 nothing and tool2 had 000 nothing
tool1 had 003 other   and tool2 had 003 other
tool1 had 000 killed  and tool2 had 000 killed

tools have 682 reached tasks in common

tool1 had 006 tasks tool2 did not found
tool2 had 005 tasks tool1 did not found

on      commonly  reached tasks, tool 1 took 1743.753669 sec. (mean 2.556824, median 0.472307, min 0.239539, max 29.289200) and tool 2 took 1693.221641 sec. (mean 2.482730, median 0.476174, min 0.250907, max 29.723400)
on *not commonly* reached tasks, tool 1 took 131.500210 sec. (mean 21.916702, median 27.256650, min 6.261610, max 29.337000) and tool 2 took 55.308238 sec. (mean 11.061648, median 7.734000, min 0.556148, max 28.053600)

among tasks reached only by tool 1, tool 2 replied 000 nothing, 006 timeout, 00 other and 00 killed
among tasks reached only by tool 2, tool 1 replied 000 nothing, 005 timeout, 00 other and 00 killed
tasks solved only by tool 1:
  testcomp/sv-benchmarks/c/nla-digbench-scaling/freire2_valuebound100.c
  testcomp/sv-benchmarks/c/loops/invert_string-1.c
  testcomp/sv-benchmarks/c/hardware-verification-bv/btor2c-lazyMod.gear.2.prop4-back-serstep.c
  testcomp/sv-benchmarks/c/hardware-verification-bv/btor2c-lazyMod.at.6.prop1-back-serstep.c
  testcomp/sv-benchmarks/c/hardware-verification-bv/btor2c-lazyMod.adding.6.prop1-back-serstep.c
  testcomp/sv-benchmarks/c/hardware-verification-bv/btor2c-lazyMod.adding.5.prop1-back-serstep.c
tasks solved only by tool 2:
  testcomp/sv-benchmarks/c/xcsp/AllInterval-013.c
  testcomp/sv-benchmarks/c/seq-pthread/cs_stateful-1.i
  testcomp/sv-benchmarks/c/loops/linear_search.c
  testcomp/sv-benchmarks/c/hardware-verification-bv/btor2c-lazyMod.peg_solitaire.5.prop1-back-serstep.c
  testcomp/sv-benchmarks/c/hardware-verification-bv/btor2c-lazyMod.adding.3.prop1-back-serstep.c
```

zapashcanon:
I'll run the benchmarks for data with Map tomorrow (without my latest change with for chunks).

@chambart also told me we should try with patricia trees instead of map, I'll try to have a go at it later. :)

PR #435, opened 2024-09-05
Add btree dataset submodule

filipeom:
From the discussion in https://github.com/OCamlPro/owi/pull/433#issuecomment-2331215405

I just randomly dropped the submodule in the `bench` directory. Let me know if you'd prefer it in a specific subdirectory or if you’d like it renamed to something else.


zapashcanon:
Thanks!

PR #461, opened 2024-12-04
use named arguments in all sub-commands thanks to cmdliner monadic syntax

zapashcanon:


zapashcanon:
cc @filipeom, you may be interested by this in other projects :-)

filipeom:
> cc @filipeom, you may be interested by this in other projects :-)

I was just looking at this, and after reviewing the code, I can't understand why it's better this way (other than the fact that it looks more declarative and satisfying :relaxed:).  

Could you briefly explain if there are any functional improvements to using the monadic syntax?

I'm going to do this on smtml :sweat_smile:

zapashcanon:
It allows named arguments, so you don't have to worry about the order anymore, which is nice when you have a lot of them! :)

filipeom:
> It allows named arguments, so you don't have to worry about the order anymore, which is nice when you have a lot of them! :)

Ah yes, this is so cool! Thanks!! I'm using this everywhere moving forward :sweat_smile:

PR #463, opened 2024-12-18
Feature: Enable OWI fuzzer.ml to save generated WASM modules to local dir

spellfish:
Generated modules will be saved in WebAssembly text format (.wat) in the output directory `generated_modules`.

- fuzz/param.ml contains a flag `save_modules` that is set to 'false' by default  ('true to enable)
- Each module will be named gen_do_module_N.wat where N is the incremental counter from global_count.
- Generated modules are valid WebAssembly modules since they're generated using the type-aware generators in gen.ml
- Modules can be converted to binary format using wat2wasm and run directly using `owi run`

Modules contain random, but valid :

- Functions
- Instructions
- Memory operations
- Global variables
- Type definitions

The generated modules are random but valid WebAssembly programs, so their behavior will vary. We can add additional parameters in param.ml to control aspects of generation like module complexity, instruction types, etc.

### DIFF
--- a/test/fuzz/fuzzer.ml
+++ b/test/fuzz/fuzzer.ml
@@ -4,6 +4,11 @@ let timeout_count = ref 0
 
 let global_count = ref 0
 
+let write_module filename m = 
+  let oc = open_out filename in
+  Fmt.pf (Format.formatter_of_out_channel oc) "%a@." Owi.Text.pp_modul m;
+  close_out oc
+
 let compare (module I1 : Interprets.INTERPRET)
   (module I2 : Interprets.INTERPRET) m =
   if Param.debug then begin
@@ -69,6 +74,20 @@ let compare (module I1 : Interprets.INTERPRET)
 
 let check (module I1 : Interprets.INTERPRET) (module I2 : Interprets.INTERPRET)
   m =
+  (* Save the generated module *)
+  if Param.save_modules then (
+    (* Create output directory if it doesn't exist *)
+    if not (Sys.file_exists Param.output_dir) then
+      Unix.mkdir Param.output_dir 0o755;
+
+    let filename = Printf.sprintf "%s/gen_do_module_%d.wat" 
+      Param.output_dir !global_count in
+    write_module filename m;
+
+    if Param.debug then 
+      Fmt.epr "Saved module to %s@\n" filename;
+  );
+
   compare (module I1) (module I2) m
 
 let add_test name gen (module I1 : Interprets.INTERPRET)
--- a/test/fuzz/fuzzer.ml
+++ b/test/fuzz/fuzzer.ml
@@ -4,10 +4,10 @@ let timeout_count = ref 0
 
 let global_count = ref 0
 
-let write_module filename m = 
-  let oc = open_out filename in
-  Fmt.pf (Format.formatter_of_out_channel oc) "%a@." Owi.Text.pp_modul m;
-  close_out oc
+let write_module (fpath : Fpath.t) m = 
+  match Bos.OS.File.writef fpath "%a@." Owi.Text.pp_modul m with
+  | Ok () -> Ok ()
+  | Error (`Msg err) -> Error (`Msg (Fmt.str "Failed to write module to %a: %s" Fpath.pp fpath err))
 
 let compare (module I1 : Interprets.INTERPRET)
   (module I2 : Interprets.INTERPRET) m =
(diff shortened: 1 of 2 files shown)


zapashcanon:
Hi @spellfish.

Thanks for the PR, I think this a an addition we're willing to accept. Do you plan to re-open the PR ? If yes, I'll have some minor comments about code style but nothing too complicated.

On another topic, I'm a little bit curious about how/why do you use Owi and our fuzzer, if you're willing to tell me more, you can drop me an e-mail or join our zulip at https://chat.ocaml.pro (ping me there and I'll add you to the Wasm discussion).

test/fuzz/fuzzer.ml
```diff
@@ -4,6 +4,11 @@ let timeout_count = ref 0
 
 let global_count = ref 0
 
+let write_module filename m = 
+  let oc = open_out filename in
+  Fmt.pf (Format.formatter_of_out_channel oc) "%a@." Owi.Text.pp_modul m;
+  close_out oc
```
zapashcanon, line 10: Can you use `Bos.OS.File.writef` instead ? You have an [example here](https://github.com/OCamlPro/owi/blob/8100737e7dc19982a5646875ec6a98ed849908d3/src/cmd/cmd_fmt.ml#L20C19-L20C37). It forces you to handle error properly (I just realized `open_out` is hidden by the `prelude` library in the whole code-base but not in the fuzzer...).

test/fuzz/fuzzer.ml
```diff
@@ -69,6 +74,20 @@ let compare (module I1 : Interprets.INTERPRET)
 
 let check (module I1 : Interprets.INTERPRET) (module I2 : Interprets.INTERPRET)
   m =
+  (* Save the generated module *)
+  if Param.save_modules then (
```
zapashcanon, line 78: In the case of `if` it is more idiomatic to use `begin ... end` rather than `( ... )`.

test/fuzz/fuzzer.ml
```diff
@@ -69,6 +74,20 @@ let compare (module I1 : Interprets.INTERPRET)
...
+    (* Create output directory if it doesn't exist *)
+    if not (Sys.file_exists Param.output_dir) then
```
zapashcanon, line 80: Could you use [`Bos.File.Dir.exists`](https://github.com/OCamlPro/owi/blob/8100737e7dc19982a5646875ec6a98ed849908d3/src/cmd/cmd_c.ml#L182) instead ?

test/fuzz/fuzzer.ml
```diff
@@ -69,6 +74,20 @@ let compare (module I1 : Interprets.INTERPRET)
...
+      Unix.mkdir Param.output_dir 0o755;
```
zapashcanon, line 81: Could you use [`Bos.File.Dir.create`](https://github.com/OCamlPro/owi/blob/8100737e7dc19982a5646875ec6a98ed849908d3/src/cmd/cmd_c.ml#L182) instead ? Also, it is better to not tests existence before (the function is going to check it for you).

test/fuzz/fuzzer.ml
```diff
@@ -69,6 +74,20 @@ let compare (module I1 : Interprets.INTERPRET)
...
+
+    let filename = Printf.sprintf "%s/gen_do_module_%d.wat"
```
zapashcanon, line 83: You can use `Fpath.(Param.output_dir / (Fmt.str "gen_do_module_%d !global_count))` instead to have proper directory separator on every platform.

test/fuzz/param.ml
```diff
@@ -14,3 +14,7 @@ let initial_fuel = 100
 let allow_partial_timeout = true
 
 let max_time_execution = 0.01 (* seconds *)
+
+let save_modules = true (* Set to false to disable saving modules *)
+
+let output_dir = "generated_modules" (* Directory to save modules *)
```
zapashcanon, line 20: `let output_dir = Fpath.v "generated_modules"` (this is better than using strings to represents paths)

test/fuzz/param.ml
```diff
@@ -14,3 +14,7 @@ let initial_fuel = 100
 let allow_partial_timeout = true
 
 let max_time_execution = 0.01 (* seconds *)
+
+let save_modules = true (* Set to false to disable saving modules *)
```
zapashcanon, line 18: I would set this to false by default because this is quite likely to make the fuzzing much more slower

test/fuzz/fuzzer.ml
```diff
@@ -4,10 +4,10 @@ let timeout_count = ref 0
...
 let global_count = ref 0
 
-let write_module filename m = 
-  let oc = open_out filename in
-  Fmt.pf (Format.formatter_of_out_channel oc) "%a@." Owi.Text.pp_modul m;
-  close_out oc
+let write_module (fpath : Fpath.t) m = 
+  match Bos.OS.File.writef fpath "%a@." Owi.Text.pp_modul m with
+  | Ok () -> Ok ()
+  | Error (`Msg err) -> Error (`Msg (Fmt.str "Failed to write module to %a: %s" Fpath.pp fpath err))
```
zapashcanon, line 10: I think this can be simplified as: `let write_module fpath m = Bos.OS.File.writef fpath "%a@." Owi.Text.pp_modul m`.

zapashcanon:
Thanks! The CI errors seems unrelated (a new warning in Frama-C). I'm merging.

PR #484, opened 2025-01-30
use pin stanza in dune-project

Swrup:
We can now pin packages with dune, hopefully this reduce the need for *.opam.template files for many packages.

I also added the owi.install file to .gitignore (generated by `dune build -p owi @install`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ _coverage
 owi-out
 results-*
 a.out.wasm
+owi.install
--- a/dune-project
+++ b/dune-project
@@ -25,6 +25,16 @@
 (source
  (github ocamlpro/owi))
 
+(pin
+  (package (name crowbar))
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@ _coverage
 owi-out
 results-*
 a.out.wasm
+owi.install
```
zapashcanon, line 7: `*.install` maybe? so that I can copy-paste this file later :-)

zapashcanon:
The pin disappeared from `owi.opam`, is it expected? If yes, I'm not sure we can accept this PR as is...

Swrup:
> The pin disappeared from owi.opam, is it expected?
No, this makes opam unable to install the package.

It seems dune is not generating the opam file with the package version correctly pinned (?).

Sorry I did not notice, I'm closing this PR for now.

zapashcanon:
Yes, dune is broken

PR #491, opened 2025-02-07
Design of Weasel

Laplace-Demon:


example/weasel/language design.md
```diff
@@ -0,0 +1,197 @@
...
+unknown annotation clause owi
+```
+
+## Types of Annotations
+
+In Weasel, `annotid` specifies the type of annotation. Different types of annotations follow slightly different grammatical rules and must appear in specific locations of the source code. Weasel defines two types of annotations:
+
+- **function contract** should appear anywhere a module field is allowed (i.e. as a custom section). Though it is recommended to place the contract immediately before the corresponding function definition.
+
+- **assertion** should appear anywhere an instruction is allowed.
```
zapashcanon, line 37: I don't remember such assertions being already implemented, are they?

Do you have an example of such an assertion and why it may be useful?

example/weasel/language design.md
```diff
@@ -0,0 +1,197 @@
...
+We present the grammar of Weasel annotations in BNF form. The following notations are adopted:
+
+- Terminal symbol is written between double quotations. `"terminal"`
+- Nonterminal symbol is written as normal words. `nonterminal`
+- Bracket is used to group symbols `{sym1 sym2 ...}`
+- Asterisk denotes zero, one or more occurrences. `{sym}*`
+- Plus denotes one or more occurrences. `{sym}+`
+- Question mark denotes zero or one occurrence. `{sym}?`
+
+```ocaml
```
zapashcanon, line 52: GitHub actually supports a variant of EBNF highlighting: https://github.com/github-linguist/linguist/issues/3075#issuecomment-314835300 ; we should use it and link to the specification instead of describing it here. Something like:

> We present the grammar of Weasel annotations in [EBNF-ISO form](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form#Table_of_symbols).

example/weasel/language design.md
```diff
@@ -0,0 +1,197 @@
...
+
+assertion ::= "(@assert" prop ")"
+
+annotation ::= contract
+    | assertion
+
+```
+
+TODO:
+- vector type
```
zapashcanon, line 194: I'm not sure we want to add vector type for now. I guess you'd like to have them in order to be able to have bigger numbers than `i64` and make some manipulations to check for integer overflow?

If yes, I believe it'd be better to leave vector out for now, and instead rely on an imported module providing big numbers (we can implement it separately later, either in pure Wasm, or directly in OCaml which is likely to be much easier).

example/weasel/language design.md
Laplace-Demon: No, they are not implemented, and I no longer think they are useful. I also described that in the first point I sent to you

example/weasel/language design.md
```diff
@@ -0,0 +1,197 @@
...
+- application of (pure) functions through call, call_indirect, call_ref
+- stack access in assertion
```
zapashcanon, line 196: What is the use-case you have in mind? This is probably do-able on the OCaml-side and by exposing some Wasm functions to do it, but it really goes against the design of Wasm hand I have mixed feelings about it.

example/weasel/language design.md
Laplace-Demon: Thanks, will do that

example/weasel/language design.md
```diff
@@ -0,0 +1,197 @@
...
+    | "i64"
+
+wasm_float_type ::= "f32"
+    | "f64"
+
+wasm_type ::= wasm_int_type
+    | wasm_float_type
+
+numerical_type ::= wasm_type
+    | "integer"
```
zapashcanon, line 166: Maybe "mathematical integer" is clearer here, because in this context it is easily confused with Wasm integers.

Also, do we need integers if we have real? How is it done in WhyML/ACSL/Gospel for instance?

example/weasel/language design.md
```diff
@@ -0,0 +1,197 @@
...
+
+requires_clause ::= "(" "requires" prop ")"
+
+assigns_clause ::= "(" "assigns" {"nothing" | memarg} ")"
+
+ensures_clause ::= "(" "ensures" prop ")"
+
+simple_behavior ::= {requires_clause | assigns_clause | ensures_clause}*
+
+named_behavior ::= "(" "behavior" {assumes_clause | requires_clause | assigns_clause | ensures_clause}* ")"
```
zapashcanon, line 182: As a first step, I'd leave named behaviors out for now, simply to make things simpler.

example/weasel/language design.md
```diff
@@ -0,0 +1,197 @@
...
+    | "||" prop prop
+    | "^^" prop prop
+    | "==>" prop prop
+    | "<==>" prop prop
+    | "?:" prop prop prop
+    | "let" {id}? term prop
+    | unpred prop
+    | binpred prop prop
+    | "forall" numerical_type {id}? prop
+    | "exists" numerical_type {id}? prop
```
zapashcanon, line 153: Don't we support UTF-8 symbols too?

example/weasel/language design.md
```diff
@@ -0,0 +1,197 @@
...
+    | "i32_load8_s" memarg
+    | "i32_load8_x" memarg
+    | "i32_load16_s" memarg
+    | "i32_load16_x" memarg
+    | "i64_load8_s" memarg
+    | "i64_load8_x" memarg
+    | "i64_load16_s" memarg
+    | "i64_load16_x" memarg
+    | "i64_load32_s" memarg
+    | "i64_load32_x" memarg
```
zapashcanon, line 120: I think we can leave out all the complicated variants:

```ocaml
    | "i32_load8_s" memarg
    | "i32_load8_x" memarg
    | "i32_load16_s" memarg
    | "i32_load16_x" memarg
    | "i64_load8_s" memarg
    | "i64_load8_x" memarg
    | "i64_load16_s" memarg
    | "i64_load16_x" memarg
    | "i64_load32_s" memarg
    | "i64_load32_x" memarg
```

And only keep the four basic ones:

```ocaml
    | "i32_load" memarg
    | "i64_load" memarg
    | "f32_load" memarg
    | "f64_load" memarg
```

I'm not sure we'll have a use-case for the complicated ones before a long time.

example/weasel/language design.md
```diff
@@ -0,0 +1,197 @@
...
+- Terminal symbol is written between double quotations. `"terminal"`
+- Nonterminal symbol is written as normal words. `nonterminal`
+- Bracket is used to group symbols `{sym1 sym2 ...}`
+- Asterisk denotes zero, one or more occurrences. `{sym}*`
+- Plus denotes one or more occurrences. `{sym}+`
+- Question mark denotes zero or one occurrence. `{sym}?`
+
+```ocaml
+ind ::= "$" id
+    | integer
```
zapashcanon, line 54: I don't think this is an integer, but rather a natural number, right? (No negative allowed)

example/weasel/language design.md
zapashcanon: OK great, let's just remove them now. :)

example/weasel/language design.md
Laplace-Demon: Agree with that. I'm not sure if it's worth to have named behaviors for now.

example/weasel/language design.md
Laplace-Demon: You are right. I missed UTF-8 symbols.

example/weasel/language design.md
Laplace-Demon: Sure, they are basically the same thing (in implementation).

example/weasel/grammar.ebnf
```diff
@@ -0,0 +1,154 @@
...
+    "clz"             (on wasm_int_type)
+  | "ctz"             (on wasm_int_type)
+  | "popcnt"          (on wasm_int_type)
+  | "abs"
+  | "neg"
+  | "sqrt"
+  | "ceil"
+  | "floor"
+  | "trunc"           (on wasm_float_type or real)
+  | "nearst" ;        (on wasm_float_type or real)
```
zapashcanon, line 32: near**e**st

example/weasel/grammar.ebnf
```diff
@@ -0,0 +1,154 @@
...
+  | "div" | "/"
+  | "rem" | "%"       (on wasm_int_type or integer)
+  | "min"
+  | "max"
+  | "and" | "&"       (on wasm_int_type or integer)
+  | "or" | "|"        (on wasm_int_type or integer)
+  | "xor" | "^"       (on wasm_int_type or integer)
+  | "shl" | "<<"      (on wasm_int_type or integer)
+  | "shr_s" | ">>"    (on wasm_int_type or integer)
+  | "shr_x" | ">>>"   (on wasm_int_type or integer)
```
zapashcanon, line 47: Many of these are quite unusual on mathematical integers, are there similar operators in WhyML/Gospel/ACSL?

example/weasel/language design.md
Laplace-Demon: It seems having mathematical integers is common practice, whereas mathematical real numbers are optional and not always usable.

For example, ACSL defines the notion of real numbers in principle, and deductive verification using ACSL can involve operations on real numbers. But how they are carried out is another matter (maybe through an axiomatic approach or approximation using floats, I don't know exactly).

The E-ACSL specification language also has real numbers but there is no quantification over real numbers (neither over floating point numbers). It allows tools that implement E-ACSL to specify the level of precision up to which real numbers are sound. (so tools don't have to support arbitrary-precision real numbers)

The E-ACSL plug-in of Frama-C chooses to support floating point numbers and rational numbers.

Gospel chooses a simpler approach: it only defines "int" (machine integer), "integer" (mathematical integer), "float" (machine float-point number), and that's all. There's no notion of real number.

Laplace-Demon: For the name, "mathematical integer" is a little bit lengthy to write in specifications. I added an explanation in the documentation, though.

Laplace-Demon: Based on that, we can either
- not to have real numbers, or
- claim to have real numbers, and
    1. only implement floating point numbers (through Wasm types) + rational numbers (through math libraries, as we will do for mathematical integers)
    2. define cast rules between those two types

example/weasel/grammar.ebnf
Laplace-Demon: I included mainly because they are present in WebAssembly. So these operations are free to get for floating point numbers, while extra work needs to be done for rational numbers.

example/weasel/language design.md
zapashcanon: Hum... we could start by not including real numbers. It may be simpler for now, WDYT?

example/weasel/grammar.ebnf
zapashcanon: Yeah sure, I mean, there is a typo in `nearst` it should be `near*e*st`.

example/weasel/grammar.ebnf
zapashcanon: Fine with me.

example/weasel/language design.md
Laplace-Demon: Though I haven't seen any nontrivial aspect of having real numbers, that would still save some work. So I agree.

example/weasel/grammar.ebnf
```diff
@@ -118,7 +118,7 @@ pprop =
   | binpred , prop , prop
   | ("forall" | "∀") , numerical_type , [ id ] , prop
   | ("exists" | "∃") , numerical_type , [ id ] , prop
-  | (* predicate application *) ;
+  | "call" , ind , { term } ; (the function must return an i32)
```
zapashcanon, line 138: Why should the function return an i32?

Laplace-Demon:
I added the `pure` clause in function contract which indicates the function should pass certain syntactic check to be recognized as a pure function, and is therefore in the scope of weasel function application. It has slight semantic difference with `assigns nothing` since the latter implies a runtime check via comparing the new and the old values.

In the following week I'll implement what's currently in the grammar. That involves the so-called "interval inference" (computing the smallest accommodatable numerical type), but also some type checking work, because:
- Weasel operators are polymorphic. They may be: 1. meaningful only on machine integers, 2. meaningful on machine integers and mathematical integers, 3. meaningful on machine integers, mathematical integers, and floating-point numbers, 4. meaningful only on floating-point numbers. So we need to validate Weasel expressions.
- Weasel propositions are also embedded in numerical types, say `i32`, and extra rules need to be introduced to specify the interaction between `prop` and `term` of type `i32`. My plan is to make clear distinction between them, except in logical predicates where `i32` may be implicitly cast to `prop`.

example/weasel/grammar.ebnf
Laplace-Demon: It depends on what type we use for prop and how term values get converted to prop. Here the design is to 1. use `i32` to encode prop, 2. only pure functions returning `i32` are considered as predicates.

zapashcanon:
> I added the pure clause in function contract which indicates the function should pass certain syntactic check to be recognized as a pure function, and is therefore in the scope of weasel function application. It has slight semantic difference with assigns nothing since the latter implies a runtime check via comparing the new and the old values.

Is this really needed? The way I understand this is : if a function is used in a contract, it must be pure. So we check it has the pure attribute, and in a separate phase, we check that every function with the pure attribute is actually pure.

Instead, we could simply check that every function used in a contract is pure directly, without going through an attribute (it may be useful for Owi to have this information anyway, even out of the Weasel context).

> Weasel operators are polymorphic. They may be: 1. meaningful only on machine integers, 2. meaningful on machine integers and mathematical integers, 3. meaningful on machine integers, mathematical integers, and floating-point numbers, 4. meaningful only on floating-point numbers. So we need to validate Weasel expressions.

Makes sense.

> Weasel propositions are also embedded in numerical types, say i32, and extra rules need to be introduced to specify the interaction between prop and term of type i32. My plan is to make clear distinction between them, except in logical predicates where i32 may be implicitly cast to prop.

Makes sense too. :)

example/weasel/grammar.ebnf
zapashcanon: Yes but we could instead allow any function call as long as it is part of an expression returning an `i32`, for instance `call $f = i64.const 42` ?

I'm not opposed to having only function returning an `i32`. I understand it is easier so it may be better to start like this and relax later if needed, but I just wanted to understand why you wanted such a restriction

Laplace-Demon: My mind went like this:

1. E-ACSL used `int` for `prop`
-> 2. we should use `int32` for `prop`
-> 3. predicates should have return type `i32`

But I didn't realize the subtle point : 1. and 2. are only about internal representation, whereas 3. is about external representation.

So, I think it is better (more intuitive) to simply have zero and non-zero numbers when representing `prop` externally, be it `i32` or `i64`. That is:

```suggestion
  | "call" , ind , { term } ; (the function must return wasm_int_type)
```

Laplace-Demon:
> Is this really needed? The way I understand this is : if a function is used in a contract, it must be pure. So we check it has the pure attribute, and in a separate phase, we check that every function with the pure attribute is actually pure.
> 
> Instead, we could simply check that every function used in a contract is pure directly, without going through an attribute (it may be useful for Owi to have this information anyway, even out of the Weasel context).

Well, it's simply because Gospel has `pure` annotation so I had a path dependence :D As you said, both will work. But the way I was thinking about is in another direction: `pure` first registers one function in Weasel, so that Weasel can recognize it in expression.  

My argument to defend `pure` can be: it might be desirable to have different scope in Weasel and in Wasm? Maybe we can reuse logic functions/predicates across several files.

PR #493, opened 2025-02-10
add union-find data structure

bclement-ocp:
The API might need to be adapted slightly — for instance it might make sense to have `union` take `default` value (or make `merge` take two `option`s) instead of raising `Invalid_argument` when the keys are not present, for the use cases @zapashcanon has in mind.


zapashcanon:
Thanks!

PR #494, opened 2025-02-10
path-condition slicing

zapashcanon:


src/symbolic/solver.ml
```diff
@@ -16,10 +16,14 @@ let fresh solver () =
 
 let check (S (solver_module, s)) pc =
   let module Solver = (val solver_module) in
+  (* TODO: maybe we should change the API of smtml so that Solver.check expects a set instead of a list ? (to avoid duplicates) *)
+  let pc = Smtml.Expr.Set.to_list pc in
   Solver.check s pc
 
 let model (S (solver_module, s)) ~symbols ~pc =
   let module Solver = (val solver_module) in
+  (* TODO: maybe we should change the API of smtml so that Solver.check expects a set instead of a list ? (to avoid duplicates) *)
```
zapashcanon, line 25: cc @filipeom, what do you think about this potential change in smtml?

src/symbolic/symbolic_path_condition.ml
```diff
@@ -32,7 +32,9 @@ let to_set pc =
 (* Return the set of constraints from [pc] that are relevant for [c]. *)
 let slice (pc : t) (c : Symbolic_value.vbool) : Smtml.Expr.Set.t =
   match Smtml.Expr.get_symbols [ c ] with
-  | [] -> Smtml.Expr.Set.singleton c
+  | [] ->
+    (* TODO: using Smtml.Expr.Set.empty seems to be working but it looks suspicious, it should be tested *)
+    Smtml.Expr.Set.singleton c
```
filipeom, line 37: I guess if `get_symbols` returns `[]` it must be a concrete expression that is not being simplified properly :cry:

src/symbolic/solver.ml
filipeom: We have this https://github.com/formalsec/smtml/blob/67ef31468fb2251071365cbb97800ad6ad08a963/src/solver_intf.ml#L64. But internally it unfold the set into a list. You can problably using this function and we optimize it directly in smtml if we can

src/symbolic/symbolic_path_condition.ml
zapashcanon: Hum... I believe this branch is sometimes reached. I can provide the list of expressions that are not simplified if you're interested? :)

src/symbolic/solver.ml
zapashcanon: Oh, great!

src/symbolic/symbolic_path_condition.ml
zapashcanon: I added a commit that fails on such not properly simplified expressions, see https://github.com/OCamlPro/owi/pull/494/commits/5d4b35b69c2ac8ef83fc8dfef982ba8b955db492

They all seem to be related to `++`, `extract` and `extend`.

filipeom: Ah thanks! I think there are a few of these that I can fix easily.

The ones with pointers are more tricky because I don't want to simplify pointers so much that we loose the `Ptr` type from a given expression. Then, we would loose the ability to track heap chunks. 

Do you think having something like an `unsafe_simplify`, that is more aggressive and simplifies every concrete expression to a given value helpful here?

zapashcanon: I need to think more about it to get a clear idea of what we want. I believe you know better than me so, `unsafe_simplify` is probably OK for now! :)

src/symbolic/symbolic_path_condition.mli
```diff
@@ -0,0 +1,13 @@
...
+(* Copyright © 2021-2024 OCamlPro *)
+(* Written by the Owi programmers *)
+
+type t
+
+val empty : t
+
+val add : t -> Symbolic_value.vbool -> t
+
+val slice : t -> Symbolic_value.vbool -> Smtml.Expr.Set.t
```
chambart, line 11: The invariant seems to be that slice can only be called when the vbool has already been added to the set.
It should be mentioned, and maybe asserted in the code also

zapashcanon:
I'm getting a few (3):

```shell-session
$ owi c ...
     Invalid_argument("index out of bounds")
     Raised at Stdlib__Domain.join in file "domain.ml", line 299, characters 16-24
     Called from Stdlib__Array.iter in file "array.ml", line 113, characters 31-48
     Called from Owi__Wq.read_as_seq.(fun) in file "src/data_structures/wq.ml", line 17, characters 4-16
     Called from Owi__Cmd_sym.cmd.print_and_count_failures in file "src/cmd/cmd_sym.ml", line 100, characters 10-20
     Called from Owi__Cmd_sym.cmd in file "src/cmd/cmd_sym.ml", line 131, characters 15-49
     Called from Cmdliner_term.app.(fun) in file "cmdliner_term.ml", line 24, characters 19-24
     Called from Cmdliner_eval.run_parser in file "cmdliner_eval.ml", line 35, characters 37-44
```

When running on Test-Comp (on programs 829, 843 and 1187).

zapashcanon:
TODO:

change `type t = Smtml.Expr.Set.t Union_find.t` to `type t = Smtml.Expr.Set.t Union_find.t * Smtml.Expr.Set.t` so that we don't have to convert the UF to a set all the time. It should be almost free to maintain.

src/symbolic/symbolic_path_condition.mli
zapashcanon: It is already asserted (I added a comment in the `.mli` about the invariant).

PR #497, opened 2025-02-10
Attempt to fix CI by Installing rust-wasm 

filipeom:


filipeom:
The ubuntu CI seems to be ok. Failing on linter, but this is already fixed in #494. 

The macos CI is failing on an expect test because previously the c compilar was only named `clang` and now it's `clang-15`.

zapashcanon:
Actually, I already opened https://github.com/OCamlPro/owi/pull/490 that depends on https://github.com/ocaml/opam-repository/pull/27381 ; but I'm fine about merging this now.

> The macos CI is failing on an expect test because previously the c compilar was only named clang and now it's clang-15.

Hum this is annoying... Maybe it is simply that we're not installing the latest version but clang-15 that is not the latest anymore and is thus not symlinked to clang?

filipeom:
> Actually, I already opened https://github.com/OCamlPro/owi/pull/490 that depends on https://github.com/ocaml/opam-repository/pull/27381 ; I expect to progress on both of them in a few days but I'm fine about merging this now.

Ah yes, I had seen this but I forgot :sweat_smile:. Maybe we can drop this PR then since I don't think any of us actively needs CI atm.

> Hum this is annoying... Maybe it is simply that we're not installing the latest version but clang-15 that is not the latest anymore and is thus not symlinked to clang?

That is the most likely cause I believe. But we could probably delete the line in the test to make it pass. (Something like `... | sed '/clang\(-1[5-9]*\)\?: error: linker command failed with exit code 1 (use -v to see invocation)/d'`)

zapashcanon:
Oh, I thought the executable was not found because of the new name. If it is simply displayed differently, indeed, we can suppress the line.

PR #510, opened 2025-02-23
Use `wait4` and measure memory usage of children in the runner

### LINKED ISSUE #478: mesure memory usage in benchmarks
with https://github.com/savonet/ocaml-mem_usage for instance

filipeom:
Should Close #478 

The PR is not ready yet, I still have to format the code, remove some comments, and I'm going to split this into two commits: one which focuses on introducing `wait4` and replacing the usage of `waitpid` and another which counts and displays `maxrss`. 

Also, I'm benchmarking this locally to see if the results are consistent with main.

But feel free to comment already :)

### DIFF
--- a/bench/report/rusage.ml
+++ b/bench/report/rusage.ml
@@ -2,4 +2,8 @@ type t =
   { clock : float
   ; utime : float
   ; stime : float
+  ; maxrss : int64
   }
+
+let pp fmt { clock; utime; stime; maxrss } =
+  Fmt.pf fmt "%4.2f %4.2f %4.2f %Ld" clock utime stime maxrss
--- a/bench/tool/tool.ml
+++ b/bench/tool/tool.ml
@@ -32,95 +32,80 @@ let mk_symbiotic () = Symbiotic
 
 exception Sigchld
 
-let kill_klee_descendants () =
-  let _ = Format.ksprintf Sys.command "pkill klee" in
-  ()
+let wait_pid ~pid ~timeout ~tool ~dst_stderr =
+  let did_timeout = ref false in
+  let start_time = Unix.gettimeofday () in
+  begin
+    try
+      Sys.set_signal Sys.sigchld (Signal_handle (fun (_ : int) -> raise Sigchld));
+      Unix.sleepf timeout;
+      did_timeout := true;
+      (* we kill the process group id (pgid) which should be equal to pid *)
+      Unix.kill (-pid) 9;
+      Sys.set_signal Sys.sigchld Signal_default
+    with Sigchld -> ()
+  end;
+  Sys.set_signal Sys.sigchld Signal_default;
+  let ( waited_pid
+      , status
+      , { ExtUnix.Specific.ru_utime = utime; ru_stime = stime; ru_maxrss = _ } )
+      =
+    ExtUnix.Specific.wait4 [] ~-pid
+  in
+  let end_time = Unix.gettimeofday () in
+  assert (waited_pid = pid);
 
-let wait_pid =
-  let last_utime = ref 0. in
-  let last_stime = ref 0. in
-  fun ~pid ~timeout ~tool ~dst_stderr ->
-    let did_timeout = ref false in
-    let start_time = Unix.gettimeofday () in
-    begin
-      try
-        Sys.set_signal Sys.sigchld
-          (Signal_handle (fun (_ : int) -> raise Sigchld));
-        Unix.sleepf timeout;
-        did_timeout := true;
-        (* we kill the process group id (pgid) which should be equal to pid *)
-        Unix.kill (-pid) 9;
-        Sys.set_signal Sys.sigchld Signal_default
-      with Sigchld -> ()
-    end;
-    Sys.set_signal Sys.sigchld Signal_default;
-    let waited_pid, status = Unix.waitpid [] (-pid) in
-    (* Because symbiotic is leaking klee processes *)
-    kill_klee_descendants ();
-    let end_time = Unix.gettimeofday () in
-    let { Rusage.utime; stime; _ } = Rusage.get Rusage.Children in
-    assert (waited_pid = pid);
+  let clock = end_time -. start_time in
 
-    let utime_diff = utime -. !last_utime in
-    let stime_diff = stime -. !last_stime in
-    last_utime := utime;
-    last_stime := stime;
-    let utime = utime_diff in
-    let stime = stime_diff in
-    let clock = end_time -. start_time in
+  (* Sometimes the clock goes a little bit above the allowed timeout... *)
+  let clock = min clock timeout in
+  let rusage = { Report.Rusage.clock; utime; stime } in
 
-    (* Sometimes the clock goes a little bit above the allowed timeout... *)
-    let clock = min clock timeout in
-    let rusage = { Report.Rusage.clock; utime; stime } in
-
-    if !did_timeout || Float.equal clock timeout then
-      Report.Run_result.Timeout rusage
-    else
-      match status with
-      | WEXITED code -> begin
-        match tool with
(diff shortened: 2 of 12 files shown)


bench/tool/tool.ml
```diff
@@ -43,22 +48,25 @@ let wait_pid =
       with Sigchld -> ()
     end;
     Sys.set_signal Sys.sigchld Signal_default;
-    let waited_pid, status = Unix.waitpid [] (-pid) in
+    let waited_pid, status, { Rusage.utime; stime; maxrss; _ } = wait4 [] pid in
```
zapashcanon, line 51: It should be `-pid` with a minus because we want to wait for the whole group (we use `setpgid` and everything somewhere else).

filipeom: My bad

filipeom:
I also made Marvin give us the memory statistics:

image

I know that you've told me you don't like that we're always iterating through the results list to accumulate the results we want to print/compare. Do you think we should make an issue to address it? 

We could try to use [owl](https://github.com/owlbarn/owl) to create a results data frame and then do some statistics/graphs?

filipeom:
On a side note, do you think that adding `wait4` to an existing Unix library might be more appropriate than explicitly including it here?  

I was thinking of [ocaml-rusage](https://github.com/CraigFe/ocaml-rusage) or, more appropriately, [extunix](https://github.com/ygrek/extunix), since they are already dependencies.

zapashcanon:
> On a side note, do you think that adding wait4 to an existing Unix library might be more appropriate than explicitly including it here?

Yes, I'd be very happy not to have to maintain this piece of C code haha. IIRC @chambart initially wrote some C bindings for wait4 but they were much more lightweight, they should still be in the commit history if you want to have a look

zapashcanon:
> I also made Marvin give us the memory statistics:

This look good but why is the `min` much bigger than `max` ? haha

bench/report/rusage.ml
```diff
@@ -2,4 +2,8 @@ type t =
   { clock : float
   ; utime : float
   ; stime : float
+  ; maxrss : int64
   }
+
+let pp fmt { clock; utime; stime; maxrss } =
+  Fmt.pf fmt "%g %g %g %Ld" clock utime stime maxrss
```
zapashcanon, line 9: While we're at it, do you think we could use `%4.2f` instead of `%g` ?

zapashcanon:
> We could try to use [owl](https://github.com/owlbarn/owl) to create a results data frame and then do some statistics/graphs?

I tried that at some point but I finally ended-up using the gnuplot ocaml bindings (see the code in `bench/report/time_distribution.ml` for instance)

bench/report/rusage.ml
filipeom: Ofc! Will change this

filipeom:
> > Btw, I know that you've told me you don't like that we're always iterating through the results list to accumulate the results we want to print/compare. Do you think we should make an issue to address it?
> 
> I don't understand what you're refering to 😅

I was talking the added functions in [bench/report/runs.ml](https://github.com/OCamlPro/owi/pull/510/files#diff-7aa3e9fddfb9f718c0a215febf994a3a9921cfdc24d8c99081de675ec41f3422)

> > On a side note, do you think that adding wait4 to an existing Unix library might be more appropriate than explicitly including it here?
> 
> Yes, I'd be very happy not to have to maintain this piece of C code haha. IIRC @chambart initially wrote some C bindings for wait4 but they were much more lightweight, they should still be in the commit history if you want to have a look

I will look into these and try to get a pr on `extunix` :crossed_fingers: 

> > I also made Marvin give us the memory statistics:
> 
> This looks good but why is the `min` much bigger than `max` ? haha

It's because in the `min_max_maxrss` function the neutral element for the `min` operator is `Int64.max_int` and since I ran the testcomp runner without files it returned `Int64.max_int`. But it should be fine, unless you want me to change it for this particular case?



> > We could try to use [owl](https://github.com/owlbarn/owl) to create a results data frame and then do some statistics/graphs?
> 
> I tried that at some point but I finally ended-up using the gnuplot ocaml bindings (see the code in `bench/report/time_distribution.ml` for instance)

Ah yes, I forgot we already had graphs :sweat_smile: my bad

zapashcanon:
> I was talking the added functions in [bench/report/runs.ml](https://github.com/OCamlPro/owi/pull/510/files#diff-7aa3e9fddfb9f718c0a215febf994a3a9921cfdc24d8c99081de675ec41f3422)

It seems OK! :)

> I will look into these and try to get a pr on extunix 🤞

Would be great, thanks for taking the time!

> It's because, in the min_max_maxrss function, the neutral element for the min operator is Int64.max_int, and since I ran the testcomp runner without files, it returned Int64.max_int. But it should be fine unless you want me to change it for this particular case.

Haha OK, I was just wondering if you inverted two numbers or something. No I don't mind about it haha, it's just some benchmarking code, it does not matter.

> Ah yes, I forgot we already had graphs 😅 my bad

Yes, I'm not sure how easily we can post them to Zulip. Maybe we can post them in a base64 encoding or something? To avoid having to upload them etc.

filipeom:
Made it depend on https://github.com/ygrek/extunix/pull/58 directly. So we need to wait for a new release or pin it. 

> Yes, I'm not sure how easily we can post them to Zulip. Maybe we can post them in a base64 encoding or something? To avoid having to upload them etc

This is a nice feature I might experiment with this in a separate PR :smiley:

zapashcanon:
@filipeom, it seems this should be ready now that a new `extunix` version was published on opam, right? It may simply be missing a `dev-setup` dependency on `extunix` and this should be good to go, right?

filipeom:
> @filipeom, it seems this should be ready now that a new `extunix` version was published on opam, right? It may simply be missing a `dev-setup` dependency on `extunix` and this should be good to go?

I'm testing it now, but there is something weird going on. I think the latest version of `extunix` pulls in a `ppx-lib` version which is conflicting with `sedlex` and since `scfg` depends on sedlex it is asking to remove smtml :swe

zapashcanon:
My guess is that a rebase is needed for the CI to pass

filipeom:
> My guess is that a rebase is needed for the CI to pass (simply to get the new dependencies on main right)

I rebased it. The problem is still the same, `extunix` in 0.4.4 upgraded the ppxlib dep to `ppxlib >= 0.36.0` while most of the packages we depend on have: `ppxlib <= 0.35.0`. And, it's a bit painful to go to all of these and make PRs to support the latest ppxlib. So maybe we'll just have to wait and see :sweat_smile:

zapashcanon:
I see. I'm going to have a look at the packages with `ppxlib <= 0.35.0` and open issues (I don't know ppxlib so I won't be able to do PR haha). Thanks for making the PR! Hopefully we'll be able to merge it at some point :)

filipeom:
> I see. I'm going to have a look at the packages with `ppxlib <= 0.35.0` and open issues (I don't know ppxlib so I won't be able to do PR haha). Thanks for making the PR! Hopefully we'll be able to merge it at some point :)

I think it's just an API change: https://github.com/ygrek/extunix/commit/7c189801d2dbd4178d90ba127df35880b932a915. It's probably easy to do, but doing PR + waiting for a release of the library is pretty time consuming and draining

filipeom:
> I'm thinking more and more to move the benchmarking code to its own repository. I believe this would fix the issue we are having here, right?

I think so, the runner is very self-contained in the dependencies that it includes so it might be ok as a stand-alone package. But that said, in your local switch you will still have the issue when installing both owi and the runner due to the ppxlib version incompatibility

filipeom:
closing in favour of https://github.com/OCamlPro/symbocalypse/pull/2

PR #517, opened 2025-02-25
Expose stop primitive in Choice monad

### LINKED ISSUE #514: export `stop` from the symbolic interface
so that it can be used in `abort` instead of adding `false` to the PC (which will keep executing a bit until the monad realize its unsatisfiable and stops execution)

it should be added here https://github.com/OCamlPro/owi/blob/main/src/intf/symbolic_choice_intf.ml

@krtab I assigned you because you may find this fun to do but I can do it if you don't want to 👼

krtab:
Closes #514

### DIFF
--- a/src/cmd/cmd_conc.ml
+++ b/src/cmd/cmd_conc.ml
@@ -184,6 +184,7 @@ let new_node (pc : Concolic_choice.pc) (head : Concolic_choice.pc_elt) : node =
   | Assume cond -> Assume { cond; cont = fresh_tree (Assume cond :: pc) }
   | Assert cond ->
     Assert { cond; cont = fresh_tree (Assert cond :: pc); status = Unknown }
+  | EltExplicitStop -> Explored
 
 let try_initialize (pc : Concolic_choice.pc) (node : eval_tree)
   (head : Concolic_choice.pc_elt) =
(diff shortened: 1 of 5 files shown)


krtab:
With the help of @chambart for the concolic one

src/concrete/concrete_choice.ml
```diff
@@ -15,6 +15,10 @@ let map v f =
   return (f v)
 [@@inline]
 
+exception ExplicitStop
+
+let stop = raise ExplicitStop
```
zapashcanon, line 20: This is raising the exception on module loading. You need something like: `let stop () = raise ExplicitStop` (but we should rather get rid of stop in the concrete case, see my other comment).

krtab: Oopsie

src/minimalist_symbolic/minimalist_symbolic_choice.mli
```diff
@@ -5,6 +5,7 @@
 type err = private
   | Assert_fail
   | Trap of Trap.t
+  | ExplicitStop
```
zapashcanon, line 8: This one seems completely unused now, right?

zapashcanon:
Thanks!

Issue #523, opened 2025-02-26
Primtives for pleasant display when replaying


krtab:
I think two primitives could be sufficient to offer a nice interface when replaying:

- A printing primitive, that should ideally handle UTF8 (or another encoding of Unicode)
- A simple primitive indicating whether we are in replay mode or not

krtab:
Done in #524

PR #525, opened 2025-02-26
Pretty print requested symbols during replay

krtab:
Blocked on: #523


krtab:
This can be merged once #524 is in

PR #529, opened 2025-02-27
Add basic klee api compatibility

S41d:


src/libc/src/klee.c
```diff
@@ -0,0 +1,83 @@
...
+    for (int i = 0; i < (size / sizeof(char)); i++) {
+      t[i] = owi_i8();
+    }
+    break;
+  }
+  return t;
+}
+
+void klee_define_fixed_object(void *addr, size_t nbytes) {
+  addr = make_in(nbytes);
```
zapashcanon, line 30: This is not doing what you expect. It should be `*addr = make_in(nbytes)`, to assign to the actual memory location pointed to by `addr`, otherwise, you are only modifying the value of the local `addr` variable.

src/libc/src/klee.c
```diff
@@ -0,0 +1,83 @@
...
+}
+
+void klee_make_symbolic(void *addr, size_t nbytes, const char *name) {
+  addr = make_in(nbytes);
```
zapashcanon, line 34: same issue here

zapashcanon:
Thanks! And congrats on your first PR ;)

PR #534, opened 2025-03-03
Fix interpet's use of stale memory

krtab:
Each time the memory is cloned, previous versions should be considered stale. This is especially true for https://github.com/OCamlPro/owi/pull/433 where the inner data structures are made purely functional, but is likely to cause bugs even without.

We should think of better ways to ensure this invariant is not broken, as the bugs caused are pretty nasty to track down.

### DIFF
--- a/src/interpret/interpret.ml
+++ b/src/interpret/interpret.ml
@@ -962,7 +962,6 @@ module Make (P : Interpret_intf.P) :
       let len, stack = Stack.pop_i32 stack in
       let c, stack = Stack.pop_i32 stack in
       let pos, stack = Stack.pop_i32 stack in
-      let* mem = Env.get_memory env mem_0 in
       let* c = Choice.select_i32 c in
       let c =
         let c = Int32.to_int c in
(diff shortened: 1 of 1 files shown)


src/interpret/interpret.ml
```diff
@@ -1406,6 +1419,7 @@ module Make (P : Interpret_intf.P) :
       in
       if out_of_bounds then Choice.trap Out_of_bounds_memory_access
       else begin
+        let* mem = Env.get_memory env mem_0 in
```
zapashcanon, line 1422: May not be needed as we're not cloning before.

zapashcanon:
There are three cases:

- we have to add/move the `let* mem` after cloning, this is *needed* to be correct;
- we simply move down the `let* mem` even without cloning, this is more future proof;
- we add a new `let* mem` even when there's no cloning, this is more future proof but it also makes things slower.

I'm fine about keeping things as is but I also think we could remove the last case?

I'll let you choose what you prefer and then I'll merge.

PR #542, opened 2025-03-05
add `owi zig` subcommand

### LINKED ISSUE #202: add a `zig` subcommand
That would be similar to `owi c` but for Zig.

zapashcanon:
fix #202 (exactly one year after opening it!)

This is full of shortcomings, like it can handle only one file at a time, it's missing a lot of options etc. But at least it is working on simple examples and allows experimenting more.


zapashcanon:
Thanks for fixing the remaining issues!

PR #549, opened 2025-03-10
add `--model-output-format` option

S41d:
Adds `--model-output-format` option with `json` and `scfg` as possible values, defaulting to `scfg`.

This also changes the output of concolic mode which wasn't `scfg` previously. I don't know if we're supposed to keep the old output format, but now the output format of concolic and symbolic modes are the same.

### DIFF
--- a/src/bin/owi.ml
+++ b/src/bin/owi.ml
@@ -42,7 +42,7 @@ let model_format_conv =
     match String.lowercase_ascii s with
     | "scfg" -> Ok Cmd_utils.Scfg
     | "json" -> Ok Json
-    | _ -> Error (`Msg "Expected \"json\" or \"scfg\" but got %s\n")
+    | _ -> Fmt.error_msg {|Expected "json" or "scfg" but got "%s"|} s
   in
   let pp fmt = function
     | Cmd_utils.Scfg -> Fmt.string fmt "scfg"
@@ -138,7 +138,7 @@ let solver =
     & info [ "solver"; "s" ] ~doc )
 
 let model_output_format =
-  let doc = "The format of the output model (\"json\" or \"scfg\")" in
+  let doc = {| The format of the output model ("json" or "scfg") |} in
   Arg.(value & opt model_format_conv Scfg & info [ "model-output-format" ] ~doc)
 
 let unsafe =
--- a/src/bin/owi.ml
+++ b/src/bin/owi.ml
@@ -124,6 +124,10 @@ let solver =
     & opt solver_conv Smtml.Solver_type.Z3_solver
     & info [ "solver"; "s" ] ~doc )
 
+let model_output_format =
+  let doc = "The format of the output model" in
+  Arg.(value & opt string "scfg" & info [ "model-output-format" ] ~doc)
+
 let unsafe =
   let doc = "skip typechecking pass" in
   Arg.(value & flag & info [ "unsafe"; "u" ] ~doc)
@@ -177,11 +181,12 @@ let cpp_cmd =
   and+ deterministic_result_order
   and+ fail_mode
   and+ solver
-  and+ profile in
+  and+ profile
+  and+ model_output_format in
   Cmd_cpp.cmd ~debug ~arch ~workers ~opt_lvl ~includes ~files ~profiling ~unsafe
     ~optimize ~no_stop_at_failure ~no_value
     ~no_assert_failure_expression_printing ~deterministic_result_order
-    ~fail_mode ~concolic ~solver ~profile
+    ~fail_mode ~concolic ~solver ~profile ~model_output_format
 
 (* owi c *)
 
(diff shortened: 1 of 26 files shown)


test/conc/immediately_fail.t
```diff
@@ -1,7 +1,5 @@
   $ owi conc immediately_fail.wat
   Assert failure
   Model:
-    (model
-      )
-  Reached problem!
+   modelReached problem!
```
zapashcanon, line 4: Looks like a line break is missing when the model is empty.

src/cmd/cmd_conc.ml
```diff
@@ -419,19 +419,45 @@ let run solver tree link_state modules_to_run =
...
   (* let no_stop_at_failure = deterministic_result_order || no_stop_at_failure in *)
+  let to_string =
+    match String.lowercase_ascii model_output_format with
+    | "json" -> fun m -> assignments_to_model m |> Smtml.Model.to_json_string
+    | "scfg" ->
+      fun m -> assignments_to_model m |> Smtml.Model.to_scfg_string ~no_value
+    | _ ->
+      Fmt.epr "Expected \"json\" or \"scfg\" but got %s\n" model_output_format;
+      assert false
+  in
```
zapashcanon, line 460: It would be better to put this in `bin/owi.ml` and write a cmdliner `model_format_conv` with a dedicated type (`type model_output_format = Scfg | Json`) so that it is not duplicated here and in symbolic.

zapashcanon:
> This also changes the output of concolic mode which wasn't scfg previously. I don't know if we're supposed to keep the old output format, but now the output format of concolic and symbolic modes are the same.

It is much better like this, I remember wanting to have this in the past. Thanks!

src/bin/owi.ml
```diff
@@ -37,6 +37,19 @@ let path_conv = Arg.conv (Fpath.of_string, Fpath.pp)
 
 let solver_conv = Arg.conv (Smtml.Solver_type.of_string, Smtml.Solver_type.pp)
 
+let model_format_conv =
+  let of_string s =
+    match String.lowercase_ascii s with
+    | "scfg" -> Ok Cmd_utils.Scfg
+    | "json" -> Ok Json
+    | _ -> Error (`Msg "Expected \"json\" or \"scfg\" but got %s\n")
```
zapashcanon, line 45: Two remarks, instead of `"hello"` it is possible to use the `{|` and `|}` delimiters in OCaml: `{|hello|}`, moreover, you can "customize" these separators, for instance, you could use `{html|` and `|html}` and write: `{html|<h1>My title</h1>|html}`.

Here, it would allow to write: `Msg {|Expected "json" or "scfg" but got %s\n|}`.

On a related note, there is a very useful function `Fmt.error_msg`, you can use it like this: `Fmt.error_message {|Expected "json" or "scfg" but got %s\n|}`.

(Also, I'm not sure the code is currently doing what you expect, I see a `%s` but no `Fmt.str` and no string is given either, using Fmt.error_msg will make it easy to solve).

src/bin/owi.ml
```diff
@@ -125,8 +138,8 @@ let solver =
     & info [ "solver"; "s" ] ~doc )
 
 let model_output_format =
-  let doc = "The format of the output model" in
-  Arg.(value & opt string "scfg" & info [ "model-output-format" ] ~doc)
+  let doc = "The format of the output model (\"json\" or \"scfg\")" in
```
zapashcanon, line 141: Same here, `{|` would be better

src/cmd/cmd_conc.ml
```diff
@@ -450,13 +450,11 @@ let cmd ~profiling ~debug ~unsafe ~rac ~srac ~optimize ~workers:_
   (* deterministic_result_order implies no_stop_at_failure *)
   (* let no_stop_at_failure = deterministic_result_order || no_stop_at_failure in *)
   let to_string =
-    match String.lowercase_ascii model_output_format with
-    | "json" -> fun m -> assignments_to_model m |> Smtml.Model.to_json_string
-    | "scfg" ->
+    match model_output_format with
+    | Cmd_utils.Json ->
+      fun m -> assignments_to_model m |> Smtml.Model.to_json_string
+    | Scfg ->
```
zapashcanon, line 456: Could you rewrite this:

```ocaml
let to_string m =
  let model = assignments_to_model m in
  match ...
```

It reads better

output.model
```diff
@@ -0,0 +1,5 @@
+{
+  "model": {
+    "symbol_0": { "scope": "some_label", "value": "0", "ty": "i32" }
+  }
+}
```
zapashcanon, line 5: I also think this was included by mistake.

src/bin/owi.ml
```diff
@@ -42,7 +42,7 @@ let model_format_conv =
     match String.lowercase_ascii s with
     | "scfg" -> Ok Cmd_utils.Scfg
     | "json" -> Ok Json
-    | _ -> Error (`Msg "Expected \"json\" or \"scfg\" but got %s\n")
+    | _ as s -> Fmt.error_msg {|Expected "json" or "scfg" but got "%s"|} s
```
zapashcanon, line 45: Maybe remove the `as s`, because if the users enters `YOLO`, he'll see: `Expected "json" or "scfg" but got yolo` (instead of `YOLO`), which is confusing

zapashcanon:
Thanks!

PR #559, opened 2025-03-21
add `--entry-point` option

S41d:


src/bin/owi.ml
```diff
@@ -159,6 +159,10 @@ let workspace =
   let doc = "write results to dir" in
   Arg.(value & opt path_conv (Fpath.v "owi-out") & info [ "outpt"; "o" ] ~doc)
 
+let entry_point =
+  let doc = "possible entry points of the executable" in
```
zapashcanon, line 163: This need to be updated (there's only one entry points now)

src/cmd/cmd_utils.ml
```diff
@@ -34,63 +34,79 @@ let write_testcase =
...
+    (function
+      | { Binary.name; _ } when List.mem name exported_names -> true
+      | _ -> false )
+    m.exports.func
+
+let set_entry_point entry_point (m : Binary.modul) =
   (* We are checking if there's a start function *)
   if Option.is_some m.start then Ok m
   else
     (* If there is none, we look for a function exported with the name `main` *)
```
zapashcanon, line 48: This comment should be updated (and the one below "We found a main function" too)

test/sym/entry_points/export.t
```diff
@@ -0,0 +1,7 @@
+export:
+  $ owi sym export.wat --entry-point=a
+  All OK
+
+  $ owi sym export.wat --entry-point=f
+  Entry point f not found
+  [26]
```
zapashcanon, line 7: Maybe add a small third test that actually fails after creating a symbol?

test/rust/entry_point.rs
```diff
@@ -0,0 +1,12 @@
+#![no_main]
+
+use owi_sym::Symbolic;
+
+#[no_mangle]
+pub extern "C" fn fun () -> i32 {
```
zapashcanon, line 6: Same here.

zapashcanon: cc @krtab, do you know if there is a way without changing the sources to tell the rust compiler not to change a function name (when it is exported in Wasm)?

S41d: After testing I found out that we don't actually need the `extern "c"` here and `#[no_mangle]` is enough. Is that good enough?

zapashcanon:
Thanks!

PR #577, opened 2025-03-28
change `-o` and add `--workspace` + set default working dir to a tmp dir

### LINKED ISSUE #563: rework the various commands to avoid outputting files everywhere
we should use a temporary directory by default and eventually add a CLI parameter allowing the user to specify the output directory

S41d:
Fixes #563

### DIFF
--- a/example/cpp/README.md
+++ b/example/cpp/README.md
@@ -180,6 +180,9 @@ OPTIONS
        --no-value
            do not display a value for each symbol
 
+       -o FILE, --output=FILE
+           Write output to a file.
+
        -O VAL (absent=3)
            specify which optimization level to use
 
@@ -201,6 +204,9 @@ OPTIONS
        -w VAL, --workers=VAL (absent=n)
            number of workers for symbolic execution. Defaults to the number
            of physical cores.
+
+       --workspace=DIR
+           write results and intermediate compilation artifacts to dir
 
 COMMON OPTIONS
        --help[=FMT] (default=auto)
--- a/example/cpp/README.md
+++ b/example/cpp/README.md
@@ -181,7 +181,7 @@ OPTIONS
            do not display a value for each symbol
 
        -o FILE, --output=FILE
-           Write output to a file.
+           Output the generated .wasm or .wat to FILE.
 
        -O VAL (absent=3)
            specify which optimization level to use
--- a/src/cmd/cmd_c.ml
+++ b/src/cmd/cmd_c.ml
@@ -168,10 +168,12 @@ let cmd ~debug ~arch ~property ~testcomp:_ ~workspace ~workers ~opt_lvl
   ~concolic ~eacsl ~solver ~profile ~model_output_format
   ~(entry_point : string option) ~invoke_with_symbols ~out_file : unit Result.t
     =
-  let workspace =
-    Option.value ~default:(Cmd_utils.tmp_dir "owi_c_%s") workspace
+  let* workspace =
+    match workspace with
+    | Some path -> Ok path
+    | None -> Bos.OS.Dir.tmp "owi_c_%s"
   in
-  let* _ = OS.Dir.create Fpath.(workspace / "test-suite") in
+  let* _did_create : bool = OS.Dir.create Fpath.(workspace / "test-suite") in
   let entry_point = Option.value entry_point ~default:"main" in
 
   let includes = Cmd_utils.c_files_location @ includes in
@@ -181,8 +183,8 @@ let cmd ~debug ~arch ~property ~testcomp:_ ~workspace ~workers ~opt_lvl
   in
   let* () = metadata ~workspace arch property files in
   let files = [ modul ] in
-  let entry_point = Some entry_point
-  and workspace = Some workspace in
+  let entry_point = Some entry_point in
+  let workspace = Some workspace in
   (if concolic then Cmd_conc.cmd else Cmd_sym.cmd)
     ~profiling ~debug ~unsafe ~rac:false ~srac:false ~optimize ~workers
     ~no_stop_at_failure ~no_value ~no_assert_failure_expression_printing
--- a/src/cmd/cmd_c.ml
+++ b/src/cmd/cmd_c.ml
@@ -69,8 +69,8 @@ let eacsl_instrument eacsl debug ~includes (files : Fpath.t list) :
     outs
   else Ok files
 
-let compile ~entry_point ~includes ~opt_lvl debug (files : Fpath.t list) :
-  Fpath.t Result.t =
+let compile ~workspace ~entry_point ~includes ~opt_lvl ~out_file debug
+  (files : Fpath.t list) : Fpath.t Result.t =
   let flags =
     let stack_size = 8 * 1024 * 1024 |> string_of_int in
     let includes = Cmd.of_list ~slip:"-I" (List.map Fpath.to_string includes) in
@@ -95,8 +95,7 @@ let compile ~entry_point ~includes ~opt_lvl debug (files : Fpath.t list) :
 
   let* clang_bin = OS.Cmd.resolve @@ Cmd.v "clang" in
 
-  let out = Fpath.(v "a.out.wasm") in
-
+  let out = Option.value ~default:Fpath.(workspace / "a.out.wasm") out_file in
   let* libc = Cmd_utils.find_installed_c_file (Fpath.v "libc.wasm") in
 
   let files = Cmd.of_list (List.map Fpath.to_string (libc :: files)) in
@@ -160,7 +159,6 @@ let metadata ~workspace arch property files : unit Result.t =
     Ok ()
   in
   let fpath = Fpath.(workspace / "test-suite" / "metadata.xml") in
-  let* (_exists : bool) = OS.Dir.create ~path:true (Fpath.parent fpath) in
   let* res = OS.File.with_oc fpath out_metadata { arch; property; files } in
   res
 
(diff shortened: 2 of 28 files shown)


example/cpp/README.md
```diff
@@ -180,6 +180,9 @@ OPTIONS
        --no-value
            do not display a value for each symbol
 
+       -o FILE, --output=FILE
+           Write output to a file.
```
zapashcanon, line 184: Which output is this? The model? The Wasm file generated by compiling the C++ programs?

src/cmd/cmd_c.ml
```diff
@@ -160,24 +159,30 @@ let metadata ~workspace arch property files : unit Result.t =
...
   ~includes ~files ~profiling ~unsafe ~optimize ~no_stop_at_failure ~no_value
   ~no_assert_failure_expression_printing ~deterministic_result_order ~fail_mode
   ~concolic ~eacsl ~solver ~profile ~model_output_format
-  ~(entry_point : string option) ~invoke_with_symbols : unit Result.t =
+  ~(entry_point : string option) ~invoke_with_symbols ~out_file : unit Result.t
+    =
+  let workspace =
+    Option.value ~default:(Cmd_utils.tmp_dir "owi_c_%s") workspace
+  in
+  let* _ = OS.Dir.create Fpath.(workspace / "test-suite") in
```
zapashcanon, line 174: Can you add a type annotation so that we don't get a partial-application without noticing? (You could also name the result so it is easier to understand).

src/cmd/cmd_c.ml
```diff
@@ -160,24 +159,30 @@ let metadata ~workspace arch property files : unit Result.t =
...
-  let* modul = compile ~entry_point ~includes ~opt_lvl debug files in
+  let* modul =
+    compile ~workspace ~entry_point ~includes ~opt_lvl ~out_file debug files
+  in
   let* () = metadata ~workspace arch property files in
-  let workspace = Fpath.(workspace / "test-suite") in
   let files = [ modul ] in
-  let entry_point = Some entry_point in
+  let entry_point = Some entry_point
+  and workspace = Some workspace in
```
zapashcanon, line 185: There's no need for `and` here

src/cmd/cmd_cpp.ml
```diff
@@ -5,34 +5,42 @@
...
-  in
-
   let err = if debug then OS.Cmd.err_run_out else OS.Cmd.err_null in
   let* () =
-    match OS.Cmd.run ~err clang_cmd with
-    | Ok _ as v -> v
-    | Error (`Msg e) ->
-      Fmt.error_msg "clang++ failed: %s"
-        (if debug then e else "run with --debug to get the full error message")
+    let rec compile_files = function
```
zapashcanon, line 17: IIUC, you had to introduce this recursive function in order to be able to use `-o` on each file, rather than compiling all of them at once, right?

src/cmd/cmd_cpp.ml
```diff
@@ -81,15 +93,23 @@ let compile ~entry_point ~includes ~opt_lvl debug (files : Fpath.t list) :
...
   let entry_point = Option.value entry_point ~default:"main" in
   let includes = Cmd_utils.c_files_location @ includes in
-  let* modul = compile ~entry_point ~includes ~opt_lvl debug files in
+  let* modul =
+    compile ~workspace ~entry_point ~includes ~opt_lvl ~out_file debug files
+  in
   let files = [ modul ] in
-  let entry_point = Some entry_point in
+  let entry_point = Some entry_point
+  and workspace = Some workspace in
```
zapashcanon, line 110: No need for `and` here

src/cmd/cmd_rust.ml
```diff
@@ -44,14 +45,20 @@ let compile ~entry_point ~includes:_ ~opt_lvl:_ debug (files : Fpath.t list) :
 let cmd ~debug ~arch:_ ~workers ~opt_lvl ~includes ~files ~profiling ~unsafe
   ~optimize ~no_stop_at_failure ~no_value ~no_assert_failure_expression_printing
   ~deterministic_result_order ~fail_mode ~concolic ~solver ~profile
-  ~model_output_format ~entry_point ~invoke_with_symbols : unit Result.t =
+  ~model_output_format ~entry_point ~invoke_with_symbols ~out_file
+  ~(workspace : Fpath.t option) : unit Result.t =
+  let workspace =
+    Option.value workspace ~default:(Cmd_utils.tmp_dir "owi_rust_%s")
+  in
+  let* _ = OS.Dir.create workspace in
```
zapashcanon, line 53: Same

src/cmd/cmd_sym.ml
```diff
@@ -51,12 +51,16 @@ let cmd ~profiling ~debug ~unsafe ~rac ~srac ~optimize ~workers
   ~no_stop_at_failure ~no_value ~no_assert_failure_expression_printing
   ~deterministic_result_order ~fail_mode ~workspace ~solver ~files ~profile
   ~model_output_format ~entry_point ~invoke_with_symbols =
+  let workspace =
+    Option.value workspace ~default:(Cmd_utils.tmp_dir "owi_sym_%s")
+  in
+  let* _ = Bos.OS.Dir.create Fpath.(workspace / "test-suite") in
```
zapashcanon, line 57: Same

src/cmd/cmd_sym.ml
```diff
@@ -120,7 +124,9 @@ let cmd ~profiling ~debug ~unsafe ~rac ~srac ~optimize ~workers
       let* () =
         if not no_value then
           let testcase = Smtml.Model.get_bindings model |> List.map snd in
-          Cmd_utils.write_testcase ~dir:workspace testcase
+          Cmd_utils.write_testcase
+            ~dir:Fpath.(workspace / "test-suite")
```
zapashcanon, line 128: Is this equivalent to the previous path? (I'm just wondering if we'll have to adjust something for testcomp or not)

src/cmd/cmd_utils.ml
```diff
@@ -38,6 +38,12 @@ let write_testcase =
     in
     res
 
+let tmp_dir pat =
+  match Bos.OS.Dir.tmp pat with
+  | Ok path -> path
+  | Error e ->
+    Fmt.failwith "Failed to create tmp directory: %a" Rresult.R.pp_msg e
```
zapashcanon, line 45: Can you avoid using `Rresult`? Also, I believe you could simply use:

```ocaml
let* path = Bos.OS.Dir tmp pat
```

Everywhere instead of declaring this one ?

src/cmd/cmd_zig.ml
```diff
@@ -18,12 +18,15 @@ let compile ~entry_point ~includes debug (files : Fpath.t list) :
...
-    let out = Fpath.set_ext ".wasm" file |> Fpath.filename |> Fpath.v in
+    let default_out = Fpath.set_ext ".wasm" file |> Fpath.filename |> Fpath.v in
+    let out = Option.value ~default:Fpath.(workspace // default_out) out_file in
     let entry = Fmt.str "-fentry=%s" entry_point in
     let zig : Cmd.t =
       Cmd.(
-        zig_bin % "build-exe" % "-target" % "wasm32-freestanding" % entry
-        %% includes % p file % p libzig )
+        zig_bin % "build-exe" % "-target" % "wasm32-freestanding"
+        % "--cache-dir" % p workspace
```
zapashcanon, line 27: Are you sure `--cache-dir` is the right option? It seems suspicious to me

src/cmd/cmd_zig.ml
```diff
@@ -45,13 +48,21 @@ let compile ~entry_point ~includes debug (files : Fpath.t list) :
...
   ~no_stop_at_failure ~no_value ~no_assert_failure_expression_printing
   ~deterministic_result_order ~fail_mode ~concolic ~solver ~profile
-  ~model_output_format ~entry_point ~invoke_with_symbols : unit Result.t =
-  let includes = Cmd_utils.zig_files_location @ includes in
+  ~model_output_format ~entry_point ~invoke_with_symbols ~out_file ~workspace :
+  unit Result.t =
+  let workspace =
+    Option.value workspace ~default:(Cmd_utils.tmp_dir "owi_zig_%s")
+  in
+  let* _ = OS.Dir.create workspace in
```
zapashcanon, line 56: same

src/cmd/cmd_cpp.ml
S41d: Yes exactly

src/cmd/cmd_sym.ml
S41d: no, it's `previous path/test-suite`

example/cpp/README.md
zapashcanon: OK, you may also change the name of the option? Removing `-o` and only having `--wasm-output` for instance?

src/cmd/cmd_cpp.ml
zapashcanon: OK fine by me. Could you add a comment saying something like:

```ocaml
(* TODO: we use this recursive function in order to be able to use `-o` on each file. We could get rid of this if we managed to call the C++ compiler and the linker in the same step as it is done for C. *)
```

(so that I eventually improve this)

src/cmd/cmd_sym.ml
zapashcanon: Fine. Can you open an issue saying "check if the test-suite directory is still working for Test-Comp"?

src/cmd/cmd_utils.ml
zapashcanon: Using it in `~default:` will still create it (~default is not lazy)

src/cmd/cmd_cpp.ml
S41d: done

PR #589, opened 2025-04-03
Fix memarg binary encoding

### LINKED ISSUE #587: memarg fields may be swapped in binary encoder
The binary [encoder](https://github.com/OCamlPro/owi/blob/bf2622c8aec476d635cdd9dee70b959f031857db/src/ast/binary_encoder.ml#L145) seems to have the fields emitted in the wrong order for `memargs`. It emits `offset; align` instead of `align; offset` which is what the [parser](https://github.com/OCamlPro/owi/blob/bf2622c8aec476d635cdd9dee70b959f031857db/src/parser/binary_parser.ml#L272) expects.

(Unless I'm reading the pattern matching wrong in the encoder, I'm new to OCaml)

zapashcanon:
Fix #587.

@simon-camp, this should fix the issue you reported.

The first commit exhibit the bug: `stack : [ i32.const -16777216 ]` become `stack        : [ i32.const -520159232 ]` after going through the binary encoder.

And you can see the fix in the second one with the test now looking better.

### DIFF
--- a/src/ast/binary_encoder.ml
+++ b/src/ast/binary_encoder.ml
@@ -143,8 +143,8 @@ let write_limits buf (limits : limits) =
     write_u32_of_int buf max
 
 let write_memarg buf ({ offset; align } : memarg) =
-  write_u32 buf offset;
-  write_u32 buf align
+  write_u32 buf align;
+  write_u32 buf offset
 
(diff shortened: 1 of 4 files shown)


simon-camp:
Cool, thanks.

PR #590, opened 2025-04-03
Improvements and fixes to symbolic value

filipeom:
- Implement printer for refs in symbolic value 

- Correctly raise a trap when an `Integer_overflow` or `Conversion_to_integer` is raised in a symbolic `trunc_f` operation

- Add missing `trunc_sat_f` operators in symbolic value

This depends on ocaml/opam-repository#27705


zapashcanon:
Looks good, thanks for taking care of it! :)

PR #594, opened 2025-04-07
add a bump allocator

### LINKED ISSUE #528: complete `owi replay`
Moreover, the `alloc`, `free` etc. functions are not available in it.

EDIT: as owi replay only works on Wasm files, the `-I` actually makes no sense (and -O neither). :-)

S41d:
closes #528 

For now, I get `Invalid_argument("index out of bounds")` on the following example: 
[example.zip](https://github.com/user-attachments/files/19633129/example.zip)

### DIFF
--- a/src/cmd/cmd_replay.ml
+++ b/src/cmd/cmd_replay.ml
@@ -12,6 +12,7 @@ let run_file ~unsafe ~optimize ~entry_point ~invoke_with_symbols filename model
       incr next;
       !next
   in
+  let brk = ref @@ Int32.of_int 0 in
 
   let module M :
     Wasm_ffi_intf.S0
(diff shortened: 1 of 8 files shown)


src/cmd/cmd_replay.ml
```diff
@@ -48,11 +49,14 @@ let run_file ~unsafe ~optimize ~entry_point ~invoke_with_symbols filename model
 
     let symbol_bool = symbol_char
 
-    let abort () = assert false
+    let abort () = ()
```
zapashcanon, line 52: I'm not sure about this because the execution will keep going on which would not be the case in the symbolic case.

We should probably display an error when reaching such a case.

zapashcanon:
Thanks, looks good. Could you add basic tests to check the call to each function?

zapashcanon:
Also, something looks wrong with the old tests.

test/replay/alloc.scfg
```diff
@@ -0,0 +1,3 @@
+model {
+  symbol symbol_0 i32 9
+}
```
zapashcanon, line 3: It may be better not to store this file and to use something like `owi c alloc.c -o alloc.wasm > alloc.scfg` in the cram test?

zapashcanon:
The CI was failing because the download link for Zig changed on their website. I fixed it on main. Can you rebase? I'll rerun the CI afterward.

zapashcanon:
Thanks!

PR #601, opened 2025-04-11
add --model-out-file

S41d:


src/cmd/cmd_conc.ml
```diff
@@ -490,14 +490,22 @@ let cmd ~profiling ~debug ~unsafe ~rac ~srac ~optimize ~workers:_
     else Ok ()
   in
   if print_paths then Fmt.pr "Completed paths: %d@." !count_path;
+  let to_file path assignements =
+    match Bos.OS.File.write path (to_string assignements) with
+    | Ok _ -> ()
+    | Error (`Msg msg) ->
+      Fmt.epr "Failed to write model to the given path: %s@." msg
```
zapashcanon, line 497: I think it is better to let owi fail when it can not write to the file. (meaning, this function should return a Result)

src/cmd/cmd_sym.ml
```diff
@@ -41,27 +41,40 @@ let run_file ~entry_point ~unsafe ~rac ~srac ~optimize ~invoke_with_symbols pc
...
     | Scfg -> Smtml.Model.to_scfg_string ~no_value
   in
+  let to_file path model =
+    let path =
+      if no_stop_at_failure then Fpath.(path + "_" + string_of_int id) else path
+    in
+    match Bos.OS.File.write path (to_string model) with
+    | Ok _ -> ()
+    | Error (`Msg msg) ->
+      Fmt.epr "Failed to write model to the given path: %s@." msg
```
zapashcanon, line 58: Same here, we should fail

src/cmd/cmd_sym.ml
```diff
@@ -41,27 +41,40 @@ let run_file ~entry_point ~unsafe ~rac ~srac ~optimize ~invoke_with_symbols pc
...
+  no_assert_failure_expression_printing =
   let to_string =
-    match model_output_format with
+    match model_format with
     | Cmd_utils.Json -> Smtml.Model.to_json_string
     | Scfg -> Smtml.Model.to_scfg_string ~no_value
   in
+  let to_file path model =
+    let path =
+      if no_stop_at_failure then Fpath.(path + "_" + string_of_int id) else path
```
zapashcanon, line 53: It should be mentioned in the manpage that the path given on CLI will be used as a prefix when there are many test cases

zapashcanon:
Thanks!

PR #608, opened 2025-04-15
add `owi_label` primitive and `--with-breadcrumbs` option

S41d:


src/symbolic/symbolic_wasm_ffi.ml
```diff
@@ -69,6 +69,32 @@ module M :
...
       Choice.return ()
     | _ -> assert false
+
+  let label m (id : Value.int32) (str_ptr : Value.int32) =
+    let open Choice in
+    let rec make_str accu i =
+      let* p = Memory.load_8_u m (Expr.value (Num (I32 i))) in
+      match Smtml.Expr.view p with
+      | Val (Num (I32 c)) ->
+        let ch = char_of_int @@ Int32.to_int c in
```
zapashcanon, line 79: `char_of_int` can fail:

> Raises Invalid_argument if the argument is outside the range 0--255.

I have not been able to hide it in `Prelude` because there is no safe equivalent in the stdlib (and I don't want to add new functions in prelude, only hide existing stuff).

It would be better to check the value of `Int32.to_int c` before calling it (we can simply `assert false` for now if it is not valid).

zapashcanon:
I added a nitpick comment.

This is missing a test for `--with-breadcrumbs` and then this should be ready to be merged.

zapashcanon:
Thanks!

PR #613, opened 2025-04-17
Print the path condition when --debug

felixL-K:
Printing the path condition when using owi with the --debug flag.

### DIFF
--- a/src/cmd/cmd_c.ml
+++ b/src/cmd/cmd_c.ml
@@ -162,12 +162,12 @@ let metadata ~workspace arch property files : unit Result.t =
   let* res = OS.File.with_oc fpath out_metadata { arch; property; files } in
   res
 
-let cmd ~debug ~arch ~property ~testcomp:_ ~workspace ~workers ~opt_lvl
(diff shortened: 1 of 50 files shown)


src/cmd/cmd_c.ml
```diff
@@ -185,7 +185,7 @@ let cmd ~debug ~arch ~property ~testcomp:_ ~workspace ~workers ~opt_lvl
   let files = [ modul ] in
   let entry_point = Some entry_point in
   let workspace = Some workspace in
-  (if concolic then Cmd_conc.cmd else Cmd_sym.cmd)
+  (if concolic then Cmd_conc.cmd else Cmd_sym.cmd ~print_pc)
```
zapashcanon, line 188: Could you add `--print-pc` to `Cmd_cond.cmd` too?

zapashcanon:
Looks good, could you add a small cram test in `test/sym/`? For instance, printing the PC on the small program that computes the wrong mean function would be nice!

zapashcanon:
Thanks and congratulations on your first PR! :)

PR #614, opened 2025-04-22
fix `model-out-file` when set with `--no-stop-at-failure`

S41d:
~fixes an error in `test\fuzz\interprets.ml` as well~

### DIFF
--- a/src/cmd/cmd_sym.ml
+++ b/src/cmd/cmd_sym.ml
@@ -84,8 +84,23 @@ let print_bug ~model_format ~labels ~model_out_file ~id ~no_value
       Fmt.str "%a" Scfg.Pp.directive { model with children }
   in
   let to_file path model =
-    let path =
-      if no_stop_at_failure then Fpath.(path + "_" + string_of_int id) else path
+    let model_ext = match model_format with Json -> "json" | Scfg -> "scfg" in
+    let contains_ext =
+      Fpath.has_ext ".json" path || Fpath.has_ext ".scfg" path
+    in
+    let* path =
+      let* () =
+        if contains_ext && (not @@ Fpath.has_ext model_ext path) then
+          Fmt.error_msg
+            "Given model file extension is not compatible with the current \
+             model format"
+        else Ok ()
+      in
+      let path = Fpath.to_string @@ Fpath.rem_ext path in
+      let base =
+        Fpath.v (if no_stop_at_failure then Fmt.str "%s_%d" path id else path)
+      in
+      Ok (Fpath.add_ext model_ext base)
     in
     Bos.OS.File.write path (to_string model labels)
   in
--- a/test/c/dune
+++ b/test/c/dune
@@ -14,4 +14,5 @@
   main.c
   malloc_aligned.c
   malloc_zero.c
+  multi_model.c
   range.c))
--- a/test/c/multi_model.c
+++ b/test/c/multi_model.c
@@ -1,0 +1,11 @@
+#include <owi.h>
+
+int main() {
+    int nondet = owi_bool();
+    if (nondet) {
+        owi_assert(0);
+    } else {
+        owi_assert(0);
+    }
+    return 0;
+}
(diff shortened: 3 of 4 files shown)


src/cmd/cmd_sym.ml
```diff
@@ -84,8 +84,17 @@ let print_bug ~model_format ~labels ~model_out_file ~id ~no_value
       Fmt.str "%a" Scfg.Pp.directive { model with children }
   in
   let to_file path model =
+    let model_ext = match model_format with Json -> "json" | Scfg -> "scfg" in
+    let contains_ext =
+      Fpath.has_ext ".json" path || Fpath.has_ext ".scfg" path
+    in
     let path =
-      if no_stop_at_failure then Fpath.(path + "_" + string_of_int id) else path
+      if contains_ext && (not @@ Fpath.has_ext model_ext path) then assert false;
```
zapashcanon, line 92: I don't understand what you are testing here ?

src/cmd/cmd_sym.ml
```diff
@@ -84,8 +84,17 @@ let print_bug ~model_format ~labels ~model_out_file ~id ~no_value
...
+      let path = if contains_ext then Fpath.rem_ext path else path in
```
zapashcanon, line 93: You don't have to test `contains_ext` before calling `rem_ext` :)

src/cmd/cmd_sym.ml
```diff
@@ -84,8 +84,17 @@ let print_bug ~model_format ~labels ~model_out_file ~id ~no_value
...
+      let base = Fpath.to_string path in
+      Fpath.v
+        ( if no_stop_at_failure then Fmt.str "%s_%d.%s" base id model_ext
+          else Fmt.str "%s.%s" base model_ext )
```
zapashcanon, line 97: All of this is a little complicated :sweat_smile: 

What about something like:

```ocaml
let base =
  if no_stop_at_failure then Fpath.(path + "_" + string_of_int id)
  else path
in
Fpath.add_ext model_ext base
```

(untested but the idea is there)

zapashcanon:
Could you add a small test (simply outputting two files and calling `ls` to check they're out there)

src/cmd/cmd_sym.ml
S41d: I'm testing if the user has given a wrong file extension, ex: `--model-out-file=out.json --model-format=scfg`, but I should probably add a better error message

zapashcanon: Ah, makes sense, indeed, having a clear error message would help here.

src/cmd/cmd_sym.ml
S41d: I tried that first but I got something like this: `model._.3.scfg`. The + operator in Fpath adds an extension so that's why I convert to a string, add the suffix, then reconvert to Fpath.t

src/cmd/cmd_sym.ml
```diff
@@ -84,8 +84,21 @@ let print_bug ~model_format ~labels ~model_out_file ~id ~no_value
...
+      if contains_ext && (not @@ Fpath.has_ext model_ext path) then
+        Fmt.error_msg
+          "Given model file extension is not compatible with the current model \
+           format"
+      else
+        let path = Fpath.to_string @@ Fpath.rem_ext path in
+        let base =
+          Fpath.v (if no_stop_at_failure then Fmt.str "%s_%d" path id else path)
+        in
+        Ok (Fpath.add_ext model_ext base)
```
zapashcanon, line 101:
```suggestion
    let+ () =
      if contains_ext && (not @@ Fpath.has_ext model_ext path) then
        Fmt.error_msg
          "Given model file extension is not compatible with the current model \
           format"
      else Ok ()
    in
    let path = Fpath.to_string @@ Fpath.rem_ext path in
    let base =
      Fpath.v (if no_stop_at_failure then Fmt.str "%s_%d" path id else path)
    in
    Fpath.add_ext model_ext base
```

zapashcanon:
Thanks! The CI failures are unrelated, I'm merging. :)

PR #619, opened 2025-04-23
prepare for smtml switch to bitvectors

zapashcanon:
cc @filipeom

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -76,7 +76,7 @@ An OCaml library is available to embed WebAssembly modules and import OCaml func
   (sedlex
     (>= 3.3))
   (smtml
-    (>= 0.6.3))
+    (>= 0.7.0))
   (uutf
     (>= 1.0.3))
   (xmlm (>= 1.4.0))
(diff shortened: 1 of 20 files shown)


filipeom:
@zapashcanon Can you rebase this? I can't solve the conflicts because I don't want understand what was changed recently, sorry :sweat_smile:

zapashcanon:
@filipeom done! :) (you can remove the commit that promote broken tests and simply regenerate it if it becomes too annoying)

filipeom:
I'll review this once more to check if everything is ok. But probably good to go once https://github.com/ocaml/opam-repository/pull/27861 is accepted.

filipeom:
Ah, so I think be blocker now is https://github.com/ocaml/opam-repository/pull/27852. So we have to wait a little longer

filipeom:
I think this is good to go, I can't quite understand why the debian CI is not passing tho

zapashcanon:
Great! Thanks for all the work. I'll investigate the CI issue later, I don't want to let this one becoming stale.

PR #626, opened 2025-04-25
optimize away an SMT call on each conditional when the first branch is

### LINKED ISSUE #625: do not check conditional negation when the conditional is unsat
When we do:

```c
if e then x else y
```

We check the satisfiability of `e` before going into `x`, and the satisfiability of its negation before going into `y`.

A possible optimization is the following: if `e` is UNSAT, then, its negation has to be SAT.

This is true only under the assumption that the current PC is always SAT (eager pruning), and we should ensure it is.

zapashcanon:
UNSAT

Fix #625.

@krtab, I'm not sure if this is related to what was done in #544, we should discuss it


krtab:
Yes it's very close to what I did in #544 we can discuss the differences soon :)

PR #675, opened 2025-04-30
remove simplifications that have been moved to smtml

### LINKED ISSUE #422: remove simplification once it has been moved to smtml
see https://github.com/OCamlPro/owi/pull/417#discussion_r1729177992

zapashcanon:
Fix #422

### DIFF
--- a/src/symbolic/symbolic_memory_make.ml
+++ b/src/symbolic/symbolic_memory_make.ml
@@ -95,19 +95,11 @@ module Make (Backend : M) = struct
     | Val (Num (I32 _)) -> v
     | _ -> Smtml.Expr.cvtop (Ty_bitv 32) (Zero_extend 16) v
 
-  (* TODO: Smtml should do this. Make call to Expr.simplify *)
-  let try_to_remove_extract v =
-    match Smtml.Expr.view v with
-    | Extract ({ node = Concat (({ node = Ptr _; _ } as ptr), _); _ }, 8, 4) ->
-      ptr
-    | Extract ({ node = Concat (_, ({ node = Ptr _; _ } as ptr)); _ }, 4, 0) ->
-      ptr
-    | _ -> v
-
   let load_32 m a =
     let open Symbolic_choice_without_memory in
     let+ a = must_be_valid_address m.data a 4 in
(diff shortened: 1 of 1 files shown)


src/symbolic/symbolic_memory.ml
```diff
@@ -60,46 +60,13 @@ let load_byte a { data; _ } =
...
-  | _ -> Smtml.Expr.make (Concat (msb, lsb))
-
 let loadn m a n =
   let rec loop addr size i acc =
     if i = size then acc
     else
       let addr' = Int32.(add addr (of_int i)) in
       let byte = load_byte addr' m in
-      loop addr size (i + 1) (concat i ~msb:byte ~lsb:acc)
+      loop addr size (i + 1) (Smtml.Expr.concat byte acc)
```
filipeom, line 69: I think for the current version of smtml we would need to use `Smtml.Expr.concat3`

src/symbolic/symbolic_memory.ml
```diff
@@ -133,26 +100,11 @@ let load_32 m a = loadn m (i32 a) 4
...
-      , ({ node = Symbol { ty = Ty_bitv 8; _ }; _ } as sym) ) ->
-    sym
-  | _ -> Smtml.Expr.make (Extract (v, pos + 1, pos))
-
 let storen m ~addr v n =
   let a0 = i32 addr in
   for i = 0 to n - 1 do
     let addr' = Int32.add a0 (Int32.of_int i) in
-    let v' = extract v i in
+    let v' = Smtml.Expr.extract v ~low:i ~high:(i + 1) in
```
filipeom, line 107:
```suggestion
    let v' = Smtml.Expr.extract2 v i in
```

Likewise, here it would be `extract2`

filipeom (commented):
I think the current version of these functions is not ready to replace the code we had here. We should use `concat3` and `extract2` instead. 

But probably to avoid breaking things atm, we should wait for https://github.com/OCamlPro/owi/pull/619

filipeom:
There might be some infinite loop in the old code for the CI to timeout :sweat_smile:

zapashcanon:
That's indeed what it looks like :sweat_smile:

PR #676, opened 2025-05-02
remove concat simplification that has been moved to smtml

zapashcanon:
Not needed anymore since #619


zapashcanon:
Not needed anymore since #619

PR #677, opened 2025-05-02
remove extract simplification that has been moved to smtml

zapashcanon:


zapashcanon:
@filipeom, I splitted #675 in three, and looks like this one is the culprit for the infinite loop

PR #679, opened 2025-05-05
add scopes

S41d:


src/cmd/cmd_replay.ml
```diff
@@ -130,6 +131,19 @@ let run_file ~unsafe ~optimize ~entry_point ~invoke_with_symbols filename model
...
+
+    let open_scope m strptr =
+      let* chars = make_str m [] strptr in
+      let str = String.init (Array.length chars) (Array.get chars) in
+      opened_scopes := str :: !opened_scopes;
+      Concrete_choice.return ()
+
+    let end_scope () =
+      match !opened_scopes with
+      | [] -> assert false
```
zapashcanon, line 143: Maybe use a proper error here?

src/cmd/cmd_sym.ml
```diff
@@ -99,30 +95,26 @@ let print_bug ~model_format ~labels ~model_out_file ~id ~no_value
...
-  | `ETrap (tr, model, _, _) -> (
-    Logs.err (fun m -> m "Trap: %s" (Result.err_to_string tr));
+  let output model labels breadcrumbs scoped_symbols =
     match model_out_file with
-    | Some path -> to_file path model
+    | Some path -> to_file path model labels breadcrumbs scoped_symbols
     | None -> begin
-      Logs.app (fun m -> m "%s" (to_string model labels));
+      Logs.app (fun m ->
+        m "%s" (to_string model labels breadcrumbs scoped_symbols) );
```
zapashcanon, line 105: It's not really ideal that we're using `to_string` here actually. We should instead define a `pp` function that works on `Format.formatter` and use it directly rather than allocating a string.

src/symbolic/scoped_symbol.ml
```diff
@@ -0,0 +1,107 @@
...
+    | Open_scope name :: rest ->
+      let children, remaining = process_scope rest in
+      let scope = { name = "scope"; params = [ name ]; children } in
+      process_tokens remaining (scope :: acc)
+    | Symbol s :: rest -> process_tokens rest (symbol_to_scfg s :: acc)
+    | Close_scope :: _ -> assert false
+  in
+  Log.debug (fun m -> m "Current scope tokens: %a" pp_list scope_tokens);
+  let children = process_tokens (List.rev scope_tokens) [] in
+  [ { name = "model"; params = []; children } ]
```
zapashcanon, line 67: I'm leaving an idea here so that I don't forget, but we should discuss it later. Maybe, when a scope (say "my_scope") is made of a single symbol, we could display only the symbol with the name coming from the scope (something like `symbol my_scope i32 42` with no surrounding scope)

src/symbolic/scoped_symbol.ml
```diff
@@ -0,0 +1,107 @@
+type scope_token =
```
zapashcanon, line 1: Maybe rename the whole module from `scoped_symbol` to `symbol_scope` (because we let the user work with scopes now, and not only the symbol as it was done in the beginning).

Then this type could be simply `token`

src/symbolic/scoped_symbol.mli
```diff
@@ -0,0 +1,19 @@
+type scope_token =
+  | Open_scope of string
+  | Close_scope
+  | Symbol of Smtml.Symbol.t
```
zapashcanon, line 4: Do we really need to expose this type? Could it could be made opaque?

src/symbolic/scoped_symbol.mli
```diff
@@ -0,0 +1,19 @@
...
+
```
zapashcanon, line 5: I'm thinking, we could have an opaque `Scoped_symbol.t` (which would be implemented as a `scope_token list`) and move a lot of the logic in this module, right?

src/symbolic/scoped_symbol.ml
```diff
@@ -0,0 +1,107 @@
...
+  | Close_scope -> Fmt.pf fmt "close"
+  | Symbol _ -> Fmt.pf fmt "symbol"
+
+let pp_list fmt tokens =
+  Fmt.list ~sep:(fun fmt () -> Fmt.pf fmt " ") pp fmt tokens
+
+let symbol s = Symbol s
+
+let only_symbols tokens =
+  List.filter_map (function Symbol s -> Some s | _ -> None) tokens
```
zapashcanon, line 18: `| _ -> None` should be replaced by `| Open_scope _ | Close_scope -> None` so that we don't get a bug if we ever add a new case here later (sorry, I did not think about it when we wrote that together)

src/symbolic/solver.ml
```diff
@@ -18,8 +18,9 @@ let check (S (solver_module, s)) pc =
   let module Solver = (val solver_module) in
   Solver.check_set s pc
 
-let model (S (solver_module, s)) ~symbols ~pc =
+let model (S (solver_module, s)) ~scoped_symbols ~pc =
   let module Solver = (val solver_module) in
+  let symbols = Some (Scoped_symbol.only_symbols scoped_symbols) in
```
zapashcanon, line 23: Something looks suspicious here: why did we use an option before and always have `Some` now? We should either remove this `Some` (and fix the following code) or change something elsewhere.

src/cmd/cmd_replay.ml
S41d: Is adding
```ocaml
        Logs.err (fun m -> m "Trying to close a non existing scope");
```
enough or should I create a new error for this case?

zapashcanon: It's fine. You can also use ~~~`Msg "blablabla"~~~ if you don't want to create a new kind of error

src/cmd/cmd_replay.ml
```diff
@@ -14,6 +14,7 @@ let run_file ~unsafe ~optimize ~entry_point ~invoke_with_symbols filename model
   in
   let brk = ref @@ Int32.of_int 0 in
   let covered_labels = Hashtbl.create 16 in
+  let opened_scopes = ref [] in
```
zapashcanon, line 17: We could use `Symbolic_scope.empty` here, right?

src/libc/include/owi.h
```diff
@@ -41,6 +41,12 @@ owi_bool(void);
 #ifdef __OWI_INTERNALS
 __attribute__((import_module("symbolic"))) __attribute__((import_name("invisible_bool_symbol"))) int
 owi_invisible_bool();
+
+__attribute__((import_module("symbolic"))) __attribute__((import_name("open_scope"))) void
+owi_open_scope(const char *name);
+
+__attribute__((import_module("symbolic"))) __attribute__((import_name("close_scope"))) void
+owi_close_scope();
```
zapashcanon, line 49:
```suggestion
owi_close_scope(void);
```

src/symbolic/symbol_scope.mli
```diff
@@ -0,0 +1,35 @@
+type token
+
+type t
+
+val symbol : Smtml.Symbol.t -> token
```
zapashcanon, line 5: I think we should make this have type:

```ocaml
val symbol: Smtml.Symbol.t -> t -> t
```

So that we can completely hide the `token` type.

src/symbolic/symbol_scope.mli
```diff
@@ -0,0 +1,35 @@
...
+
+val open_scope : string -> token
```
zapashcanon, line 7:
```ocaml
val open_scope : string -> t -> t
```

src/symbolic/symbol_scope.mli
```diff
@@ -0,0 +1,35 @@
...
+
+val close_scope : token
```
zapashcanon, line 9:
```ocaml
val close_scope : t -> t
```

src/symbolic/symbol_scope.mli
```diff
@@ -0,0 +1,35 @@
...
+
+val empty : t
+
+val push : token -> t -> t
```
zapashcanon, line 13: This can probably be removed once we do the API change I'm proposing.

src/symbolic/symbol_scope.mli
```diff
@@ -0,0 +1,35 @@
...
+
+val only_symbols : t -> Smtml.Symbol.t list
+
+val of_expressions : Smtml.Expr.Set.t -> t
+
+val of_symbol : Smtml.Symbol.t -> t
+
+val pp_token : Format.formatter -> token -> unit
```
zapashcanon, line 21: This can probably be removed once we do the API change I'm proposing.

src/cmd/cmd_replay.ml
S41d: We could, but I don't know if it adds anything to use Symbolic_scope in replay. We only keep track of the last opened scope in replay anyways. Do you still want me to use `Symbolic_scope.empty` ?

src/symbolic/symbol_scope.mli
```diff
@@ -0,0 +1,35 @@
...
+
+val to_scfg : no_value:bool -> Smtml.Model.t -> t -> Scfg.Types.directive list
+
+val to_json :
+     no_value:bool
+  -> Smtml.Model.t
+  -> t
+  -> [> `Assoc of (string * Yojson.Basic.t) list ]
+
+val model_of_json : string -> (Smtml.Symbol.t, Smtml.Value.t) Hashtbl.t Result.t
```
zapashcanon, line 33: > (Smtml.Symbol.t, Smtml.Value.t) Hashtbl.t

I think there's a type for that in Smtml ?

src/symbolic/symbol_scope.mli
```diff
@@ -0,0 +1,35 @@
...
+
+val model_of_scfg : string -> (Smtml.Symbol.t, Smtml.Value.t) Hashtbl.t Result.t
```
zapashcanon, line 35: (same)

src/symbolic/symbol_scope.mli
```diff
@@ -0,0 +1,35 @@
...
+
+val of_expressions : Smtml.Expr.Set.t -> t
+
+val of_symbol : Smtml.Symbol.t -> t
+
+val pp_token : Format.formatter -> token -> unit
+
+val pp : Format.formatter -> t -> unit
+
+val to_scfg : no_value:bool -> Smtml.Model.t -> t -> Scfg.Types.directive list
```
zapashcanon, line 25:
```suggestion
val to_scfg : no_value:bool -> Smtml.Model.t -> t -> Scfg.Types.config
```

src/symbolic/symbol_scope.ml
```diff
@@ -0,0 +1,242 @@
...
+        Logs.err (fun m -> m "Reached end without closing scope: %s" name);
+        assert false
+      | Close_scope :: rest -> (List.rev acc, rest)
+      | Open_scope name :: rest ->
+        let content, remaining = process_scope name rest in
+        let scope =
+          match content with
+          | sym :: [] ->
+            { name = sym.name; params = sym.params @ [ name ]; children = [] }
+          | _ as children -> { name = "scope"; params = [ name ]; children }
```
zapashcanon, line 70:
```suggestion
          | children -> { name = "scope"; params = [ name ]; children }
```

src/symbolic/symbol_scope.ml
```diff
@@ -0,0 +1,242 @@
...
+  let result = process_tokens rev_scope_tokens [] in
+  `Assoc [ ("model", `List result) ]
+
+let model_of_json json_str =
+  let open Yojson.Basic in
+  let open Syntax in
+  let tbl = Hashtbl.create 16 in
+
+  let json =
+    try Ok (from_string json_str)
```
zapashcanon, line 170: Is there a non-raising version of this function?

src/symbolic/symbol_scope.ml
```diff
@@ -0,0 +1,242 @@
...
+      Hashtbl.add tbl key value;
+      process_json rest
+    | `Assoc obj :: rest when List.mem_assoc "scope" obj ->
+      let content = Util.member "content" (`Assoc obj) |> Util.to_list in
+      let* () = process_json content in
+      process_json rest
+    | _ :: _ -> assert false
+    | [] -> Ok ()
+  in
+  let* parsed_json = json in
```
zapashcanon, line 200: Why not put the `let*` before when actually parsing the json ?

src/symbolic/symbol_scope.ml
S41d: I can't seem to find any, they all raise Json_error

zapashcanon:
OK so I have one higher-level question: what happens if the user creates many scopes with the same name?

S41d:
Do you have any cases in mind where it'd cause a bug?
Here's a json example of with two scopes with an empty string as scope name:
```json
{
  "labels": [],
  "model": [
    {
      "scope": "",
      "content": [
        { "symbol": "symbol_0", "type": "i32", "value": "0" },
        { "symbol": "symbol_1", "type": "i32", "value": "0" }
      ]
    },
    {
      "scope": "",
      "content": [
        { "symbol": "symbol_2", "type": "i32", "value": "0" },
        { "symbol": "symbol_3", "type": "i32", "value": "0" }
      ]
    }
  ]
}
```
same example in scfg:
```scfg
model {
  scope "" {
    symbol symbol_0 i32 0
    symbol symbol_1 i32 0
  }
  scope "" {
    symbol symbol_2 i32 0
    symbol symbol_3 i32 0
  }
}
```
both cases reach the assertion failure

zapashcanon:
[I think this is a bug](https://github.com/OCamlPro/owi/pull/679/files#diff-1cfd2fc62de7cde755be6173cce9c5a5b4db6dfe1ce100f0cc837d496a908804R227]) (and there's the same in JSON), it'll probably appear only when using replay.

You should be able to trigger it with the following:

```c
void f(void) {
    int x = symbol_i32("aaa");
    int y = symbol_i32("aaa");
    if (x == 42 && y != 42) {
      assert(false);
    }
}
```

Running `owi c` will find a bug, but then, `replay` will not find it.

zapashcanon:
Thanks!

PR #687, opened 2025-05-15
fix src/dune

gasche:
The `integers` dependency was removed in 4b37195b07faddda599eaf7f8c8054597855f7c8 .


zapashcanon:
Thanks! The CI error looks like some solver suddenly deciding to pick another integer so it's fine.

PR #694, opened 2025-05-21
split c files

### LINKED ISSUE #533: change c files structure
instead of having `src/libc/*` we should rather have `src/lang_c/{libowi,libklee,libc}/*`

zapashcanon:
Fix #533


src/lang_c/libowi/src/test-comp.c
```diff
@@ -0,0 +1,66 @@
...
+
+__int128 __VERIFIER_nondet_int128(void) {
+  // TODO: looks wrong!
+  return owi_i32();
+}
+
+unsigned __int128 __VERIFIER_nondet_uint128(void) {
+  // TODO: looks wrong!
+  return (unsigned __int128)owi_i32();
+}
```
filipeom, line 41: I think I was responsible for this :sweat_smile: Do the type casts even do anything in the compiled wasm? 

Should we just instead create the appropriate `owi_{i|u}{8|16|32|64|128}` for the respective types? I could do this I think

filipeom: > Yes don't worry, I'm going to take care of this! :D
Ah okok

> I think the cast actually have an effect on the Wasm code generated by clang. The thing is that, some types, like `long` for instance, can be either 32 bits or 64bits depending on the platform and I have to think about which behavior we want.

We can probably do something with macros to resolve it at compile time:

```c
#if LONG_MAX == 2147483647L
    #define LONG_IS_32_BIT
#elif LONG_MAX == 9223372036854775807L
    #define LONG_IS_64_BIT
#endif


long __VERIFIER_nondet_long(void) {
#if defined(LONG_IS_32_BIT)
    return (long)owi_i32();
#elif defined(LONG_IS_64_BIT)
    return (long)owi_i64();
#endif
}
```
Probably not using `limits.h` constants because we probably hardcode those constants, but maybe using `sizeof` would work

zapashcanon: Yes, I expect the C compiler to be smart enough to optimize away the branching in:

```c
  if (size == 32) {
    return (long) owi_int32();
  } else if (size == 64) {
    return (long) owi_int64();
  } else {
    // error unsupported platform
  }
```

PR #696, opened 2025-05-22
Start handling v128

chambart:
This is just the basis for adding the v128 type and the strict minimal to get the `reference/simd_const.wast` test is now passing.

There are still a few `assert false` TODOs.

This is not sufficient to test GHC generated wasm files (it's missing `v128.load`) but that should not be too much.
It does not have the corresponding wat2wasm tests for what is handled


src/primitives/v128.mli
```diff
@@ -0,0 +1,40 @@
...
+  -> int
+  -> int
+  -> int
+  -> int
+  -> int
+  -> int
+  -> int
+  -> int
+  -> int
+  -> t
```
zapashcanon, line 22: :scream:

zapashcanon:
Thanks!

PR #698, opened 2025-05-27
V128: proper symbolic and concolic handling

chambart:
Still TODO from previous PR #696 are `I8x16` and `I16x8` additions


zapashcanon:
I think you need to add a `v128_symbol` in `src/cmd/cmd_replay.ml` to fix the tests. This is because your are now adding one such import in `owi iso` for which there is a test that is replayed later.

zapashcanon:
Thanks!

PR #699, opened 2025-05-30
Model with entry point

S41d:


doc/src/manpages/owi-c.md
```diff
@@ -56,6 +56,9 @@ OPTIONS
            given this is used as a prefix and the ouputed files would have
            PREFIX_%d.
 
+       --model-with-entry-point
+           Add the entry point in the generated model for easier replay.
```
redianthus: Is there a case where we wouldn't want to add it?

src/cmd/cmd_replay.ml
```diff
@@ -298,8 +298,13 @@ let cmd ~unsafe ~optimize ~replay_file ~source_file ~entry_point
...
           bindings
       in
-      Array.of_list model
+      (entry_point, Array.of_list model)
   in
+  let entry_point =
+    if Option.is_none entry_point then model_entry_point else entry_point
+  in
+  Logs.debug (fun m -> m "%s" (Option.value ~default:"na" entry_point));
+  Logs.debug (fun m -> m "%s" (Option.value ~default:"na" model_entry_point));
```
redianthus: Maybe add more information about what you're printing? Something like `m "entry point: %s" ...`

src/cmd/cmd_sym.ml
```diff
@@ -193,10 +207,16 @@ let handle_result ~workers ~no_stop_at_failure ~no_value
       Array.iter Domain.join join_handles )
   in
   let results = sort_results deterministic_result_order results in
+  let* entry_point =
+    if model_with_entry_point then
+      match entry_point with Some ep -> Ok ep | None -> Fmt.error_msg ""
+    else Ok ""
```
redianthus: This looks suspicious

doc/src/manpages/owi-c.md
S41d: I was thinking about cases where we would want to start replay with other entry points than the one we specified for symbolic execution, but we have the `--entry-point` option already so I guess that case is already taken care of. I don't think there's another case

src/cmd/cmd_replay.ml
S41d: These were meant just for debugging a problem I had 😅 . They could be useful so I'll add what you suggested

src/cmd/cmd_sym.ml
S41d: I'll add a better error message. Other than that, I didn't really find a better way to do this. I just wanted to extract the string but also return an error if `--model-with-entry-point` was set and we couldn't find an entry point. This will be simplified a bit when I remove the option though

doc/src/manpages/owi-c.md
redianthus: Well, if you want to replay with a given model but on a different entry point, I think things are not going to go well haha : if you have a different number of symbols or different types of symbols it'll crash, if by chance they match, it'll probably not find anything interesting. I'm for removing this option completely

src/cmd/cmd_replay.ml
```diff
@@ -298,8 +298,16 @@ let cmd ~unsafe ~optimize ~replay_file ~source_file ~entry_point
                    (Fmt.str "unexpected value type: %a" Smtml.Value.pp v) ) )
           bindings
       in
-      Array.of_list model
+      (entry_point, Array.of_list model)
   in
+  let entry_point =
+    if Option.is_none entry_point then model_entry_point else entry_point
```
redianthus, line 304: Does it make sense to keep the `--entry-point` argument for `owi replay`? Couldn't we remove it now? And if not, what happen when they are both defined?

src/symbolic/symbol_scope.mli
```diff
@@ -24,6 +24,6 @@ val to_json :
   -> t
   -> [> `Assoc of (string * Yojson.Basic.t) list ]
 
-val model_of_json : string -> Smtml.Model.t Result.t
+val model_of_json : string -> (string option * Smtml.Model.t) Result.t
 
-val model_of_scfg : string -> Smtml.Model.t Result.t
+val model_of_scfg : string -> (string option * Smtml.Model.t) Result.t
```
redianthus, line 29: This could be a `string * Smtml.Model.t` now that the entry point is mandatory, right?

src/cmd/cmd_sym.ml
```diff
@@ -71,10 +72,23 @@ let print_bug ~model_format ~model_out_file ~id ~no_value ~no_stop_at_failure
...
+          let ep =
+            [ { Scfg.Types.name = "entry_point"
+              ; params = [ entry_point ]
+              ; children = []
+              }
+            ]
+          in
+          { model with children = model.children @ ep }
+        | None -> model
+      in
```
redianthus, line 90: Do you think we could put the entry point first in the output model (followed by the symbols)? I find it easier to read this way

S41d: sure, i'll do that

src/cmd/cmd_replay.ml
S41d: I guess not, I'll remove the option

PR #701, opened 2025-06-05
Log SMT queries per test

felixL-K:
This issue starts logging all SMT queries sent to the solver during each test in bench/. Logs are written to a queries_log.jsonl file per test.

What works
- Each query is logged as JSON (hash + assumptions).

Known issues
- No timing or performance info is logged yet.

Next steps
- Add query timing, memory, and CPU metrics.

How to use
```
git clone https://github.com/felixL-K/smtml -b trace-smt-queries
cd smtml && dune b @install && dune install

git clone https://github.com/felixL-K/owi -b trace-queries2
cd owi && dune exec -- testcomp/testcomp.exe <your-test-id>
```

### DIFF
--- a/bench/tool/tool.ml
+++ b/bench/tool/tool.ml
@@ -49,6 +49,8 @@ let wait_pid =
         Unix.sleepf timeout;
         did_timeout := true;
         (* we kill the process group id (pgid) which should be equal to pid *)
+        Unix.kill (-pid) 15;
+        Unix.sleepf 1.;
         Unix.kill (-pid) 9;
         Sys.set_signal Sys.sigchld Signal_default
       with Sigchld -> ()


bench/testcomp/testcomp.ml
```diff
@@ -4,7 +4,7 @@ module Unix = struct
 end
 
 let tool =
-  Tool.mk_owi ~concolic:false ~workers:56 ~optimisation_level:3
```
redianthus: Could you revert this?

src/cmd/cmd_sym.ml
```diff
@@ -215,7 +215,8 @@ let cmd ~unsafe ~rac ~srac ~optimize ~workers ~no_stop_at_failure ~no_value
     | Some path -> Ok path
     | None -> OS.Dir.tmp "owi_sym_%s"
   in
-
+  let log_path = String.concat "" [Fpath.(to_string @@ parent workspace); "queries_log.jsonl"] in
```
redianthus:
```suggestion
  let log_path = Fpath.(parent workspace / v "queries.jsonl") |> Fpath.to_string in
```

(untested but much more readable)

bench/testcomp/testcomp.ml
felixL-K: I don't know how multiple domains are going to interact with Smtml.Tmp_log_path.
It seems to be doing ok.

filipeom: If you want to be extra safe maybe add a mutex?

bench/testcomp/testcomp.ml
```diff
@@ -5,7 +5,7 @@ end
 
 let tool =
   Tool.mk_owi ~concolic:false ~workers:56 ~optimisation_level:3
-    ~solver:Smtml.Solver_type.Z3_solver
```
redianthus: Can you revert this?

src/cmd/cmd_sym.ml
```diff
@@ -228,7 +228,10 @@ let cmd ~unsafe ~rac ~srac ~optimize ~workers ~no_stop_at_failure ~no_value
     | Some path -> Ok path
     | None -> OS.Dir.tmp "owi_sym_%s"
   in
-
+  (* setting the file where to log all queries sent to solvers *)
+  let log_path = Fpath.(parent workspace / "queries_log.txt") |> Fpath.to_string in
```
redianthus: Can you remove the `Fpath.to_string` part here?

src/cmd/cmd_sym.ml
```diff
@@ -228,7 +228,10 @@ let cmd ~unsafe ~rac ~srac ~optimize ~workers ~no_stop_at_failure ~no_value
...
+  Smtml.Tmp_log_path.init log_path (Fmt.to_to_string Smtml.Solver_type.pp solver);
```
redianthus: You can remove the `solver` part

src/cmd/cmd_sym.ml
```diff
@@ -228,7 +228,10 @@ let cmd ~unsafe ~rac ~srac ~optimize ~workers ~no_stop_at_failure ~no_value
...
+  Stdlib__Domain.at_exit (fun () -> Smtml.Tmp_log_path.close ());
```
redianthus: Why don't you simply call this function at the end of the `owi sym` command? Moreover, I believe all of this code can be moved to the bench code directly, it does not need to be part of Owi's core code.

felixL-K: The logger in smtml/tmp_log_path uses thread-local state, so init must be called from within the thread (or domain) running the solver. Since bench uses Unix.execvp to launch Owi, any state initialized before is lost. That's why unfortunately logging must remain inside Owi’s logic I think.

redianthus: OK, makes sense. What about using an environment variable?

felixL-K: That's actually what I had implemented initially using an environment variable, but after discussing with Arthur, we felt it wasn't the cleanest solution. That said, I just reimplemented it that way, it turns out to be quite practical in this case.

bench/tool/tool.ml
```diff
@@ -49,6 +49,8 @@ let wait_pid =
         Unix.sleepf timeout;
         did_timeout := true;
         (* we kill the process group id (pgid) which should be equal to pid *)
+        Unix.kill (-pid) 15;
+        Unix.sleepf 1.;
```
redianthus, line 53: We could do something like:

```ocaml
if Env.is_set "LOG_QUERY_PATH" then begin
   Unix.kill (-pid 15);
   Unix.sleepf 1.
end
```

So that it runs normally with the correct timeout when we don't log queries ?

zapashcanon:
Could you make these changes on https://github.com/OCamlPro/symbocalypse instead?

Issue #691, opened 2025-05-16
design better combinators for the FFI


zapashcanon:
It should be possible to use something like:

```ocaml
int32 *-> int32 *-> unit
```

rather than the current syntax with constructor that is quite heavy

PR #704, opened 2025-06-12
add (^->) syntax for FFI

swrup:
issue #691

### DIFF
--- a/src/symbolic/symbolic_wasm_ffi.ml
+++ b/src/symbolic/symbolic_wasm_ffi.ml
@@ -149,77 +149,42 @@ end
 type extern_func = Symbolic.Extern_func.extern_func
 
 open M
+open Symbolic.Extern_func
+open Symbolic.Extern_func.Syntax
 
 let symbolic_extern_module =
   let functions =
-    [ ( "i8_symbol"
-      , Symbolic.Extern_func.Extern_func (Func (UArg Res, R1 I32), symbol_i8) )
-    ; ( "i16_symbol"
-      , Symbolic.Extern_func.Extern_func (Func (UArg Res, R1 I32), symbol_i16)
-      )
-    ; ( "i32_symbol"
-      , Symbolic.Extern_func.Extern_func (Func (UArg Res, R1 I32), symbol_i32)
-      )
-    ; ( "i32_symbol_constant"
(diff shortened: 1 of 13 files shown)


src/symbolic/symbolic_wasm_ffi.ml
```diff
@@ -208,16 +208,13 @@ let symbolic_extern_module =
...
-    ; ( "dealloc"
-      , Symbolic.Extern_func.Extern_func
-          (Func (Mem (Arg (I32, Res)), R1 I32), free) )
-    ; ("abort", Symbolic.Extern_func.Extern_func (Func (UArg Res, R0), abort))
-    ; ( "exit"
-      , Symbolic.Extern_func.Extern_func (Func (Arg (I32, Res), R0), exit) )
+    [ ("alloc", Extern_func (memory ^-> i32 ^-> i32 ^->. i32, alloc))
+    ; ("dealloc", Extern_func (memory ^-> i32 ^->. i32, free))
+    ; ("abort", Extern_func (unit ^->. unit, abort))
+    ; ("exit", Extern_func (i32 ^->. unit, exit))
```
redianthus, line 187: Wow! This is really nice! Thanks!

zapashcanon:
Thanks a lot! Is there something left to do here?

Do you want to port all the various FFIs to it in this PR? If not, I'll do it myself later. You can find all of them easily, as they all should be in a `X_wasm_ffi.ml` with X={concrete,symbolic,concolic,minimalist_symbolic}.

Also, I believe we should update the doc to use this as this is much easier for users, but I can do it too.

Just let me know if I should merge or let you more time to work on this. :)

zapashcanon:
> I'm not sure about the ^->./^->.... thing, maybe there is a better way?

Oh, OK! They are only for results, makes sense. Then it is even much easier than what I thought! I think this is fine, we can still improve it latter if we find a better way.

> I think it looks good and is better to write and read, but it adds a layer of complexity too

I think it's fine, the code required GADTs expertise before, anyone able to understand it before would also understand it now. But it makes things much better for users of the library, so I think it's clearly a positive change.

zapashcanon:
You can simply add a unit argument to this function and it should be good enough! :-)

zapashcanon:
Thanks a lot!

PR #708, opened 2025-06-17
make model formats consistent

S41d:
This pr separates labels and breadcrumbs into their own separate fields for the Scfg format to make it more consistent with the Json format


zapashcanon:
Thanks!

PR #712, opened 2025-06-18
add LIFO, FIFO, and Random exploration strategies

Julielanor:


src/bin/owi.ml
```diff
@@ -96,6 +96,18 @@ let fail_mode =
...
+  let fifo_doc = "explore the tree in fifo order" in
+  let lifo_doc = "explore the tree in lifo order" in
+  let random_doc = "explore the tree randomly" in
+  Arg.(
+    value
+    & vflag Cmd_sym.Random
+        [ (FIFO, info [ "fifo" ] ~doc:fifo_doc)
+        ; (LIFO, info [ "lifo" ] ~doc:lifo_doc)
+        ; (Random, info [ "random" ] ~doc:random_doc)
+        ] )
```
redianthus: Thanks! Could it be changed so that they are not split in the manpage? I.e. we have `owi .. --exploration=fifo`. A similar thing is done for the solver choice, you can have a look at it to see how you can do it.

src/symbolic/symbolic_choice.ml
```diff
@@ -13,9 +13,9 @@ module Prio = struct
   (*
       Currently there is no real notion of priority. Future extensions adding it will ho here.
     *)
-  type t = Default
+  type t = Random
 
-  let default = Default
+  let random = Random  let to_int = function Random -> Random.int 10000000
```
redianthus: Something is weird with the formatting. Could you try running `dune fmt` and make a new commit?

src/symbolic/symbolic_choice.ml
```diff
@@ -71,31 +71,32 @@ module CoreImpl = struct
...
+    let add_init_task sched task =
+      Work_datastructure.push task 0 sched.work_queue
 
     let work wls sched callback =
       let rec handle_status (t : _ Schedulable.status) write_back =
         match t with
         | Stop -> ()
         | Now x -> callback x
-        | Yield (_prio, f) -> write_back f
+        | Yield (prio, f) -> write_back (Prio.to_int prio, f)
```
redianthus: I think it would be better if this `Prio.to_int` was made inside `Wq,Wpq,Ws` instead.

src/intf/work_ds_intf.ml
```diff
@@ -0,0 +1,33 @@
...
+(* Written by the Owi programmers *)
+
+module type S = sig
+  type !'a t
+
+  (** Create a new queue *)
+  val make : unit -> 'a t
+
+  (** Add a new element to the queue *)
+  val push : 'a -> int -> 'a t -> unit
```
redianthus: Could you replace `int` by `Prio.t`?

src/bin/owi.ml
```diff
@@ -96,6 +96,18 @@ let fail_mode =
         ; (Assertion_only, info [ "fail-on-assertion-only" ] ~doc:assert_doc)
         ] )
 
+let exploration_strategy =
+  let fifo_doc = "explore the tree in fifo order" in
+  let lifo_doc = "explore the tree in lifo order" in
+  let random_doc = "explore the tree randomly" in
+  Arg.(
+    value
+    & vflag Cmd_sym.Random
```
redianthus: I don't think we want the `Random` strategy by default! I would rather use `LIFO` which is likely to be the best one. Or `FIFO` which is similar to what we had before, and thus should not change the output of too many tests for now (and we can change the default later).

zapashcanon:
Also, you can add an entry in `CHANGES.md`

src/bin/owi.ml
```diff
@@ -97,16 +112,8 @@ let fail_mode =
...
-  let lifo_doc = "explore the tree in lifo order" in
-  let random_doc = "explore the tree randomly" in
-  Arg.(
-    value
-    & vflag Cmd_sym.Random
-        [ (FIFO, info [ "fifo" ] ~doc:fifo_doc)
-        ; (LIFO, info [ "lifo" ] ~doc:lifo_doc)
-        ; (Random, info [ "random" ] ~doc:random_doc)
-        ] )
+  let doc = "exploration strategy to use (\"fifo\", \"lifo\" or \"random\")" in
```
redianthus: To avoid escaping, you can use `{|` and `|}` as delimiters

zapashcanon:
Thanks!

PR #714, opened 2025-06-19
refactor symbolic parameters so that they are shared accross any command

zapashcanon:
that eventually calls the symbolic engine, thanks to @gasche for complaining enough on this one to get me working on it


gasche:
I agree that the code looks nicer, and also the documentation changes suggest that the interface became more consistent across the various tools (some extra options have been added for some tools, hopefully this is an improvement).

PR #716, opened 2025-06-19
some out of bounds checks improvements

zapashcanon:

PR #717, opened 2025-06-20
oob improvements, round 2

zapashcanon:
Changes from #716 that probably requires some changes in Smt.ml

### DIFF
--- a/src/concolic/concolic_wasm_ffi.ml
+++ b/src/concolic/concolic_wasm_ffi.ml
@@ -162,7 +162,10 @@ module M :
 
   let ptr ((_c, s) : Value.int32) : int32 Concolic_choice.t =
     match view s with
-    | Ptr { base; _ } -> Concolic_choice.return base
+    | Ptr { base; _ } ->
+      (* TODO: why don't we use offset here ? *)
+      let base = Smtml.Bitvector.to_int32 base in
+      Concolic_choice.return base
     | _ ->
       Logs.err (fun m ->
         m {|free: cannot fetch pointer base of "%a"|} Expr.pp s );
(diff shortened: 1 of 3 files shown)


filipeom:
Can you check if formalsec/smtml#376 fixes the simplification issues for you?

filipeom:
> Which looks mostly good to me! I am not sure if we can also try to simplify (i32.wrap_i64 (i64.extend_i32_u symbol_0)) ?

Yes, I think it's safe to return symbol_0 right? extend_i32_u just pads the int with zeros so my thinking is that we can return `symbol_0` just fine.

> Regarding the collections-c errors, I am not even sure they were correct before.. Aren't the files in normal supposed to be bug-free?

Yes it was originally intended to be bug-free, but this was the bug that wasp and owi detected in collection-c. One 4-byte read out-of-bounds. Now what's weird is that it's saying that it is an `out of bounds memory access` instead of an `memory heap buffer overflow`. I'd have to look at the diff more closely to try to figure this one out

zapashcanon:
> Yes, I think it's safe to return symbol_0 right? extend_i32_u just pads the int with zeros so my thinking is that we can return symbol_0 just fine.

I agree.

> Yes it was originally intended to be bug-free, but this was the bug that wasp and owi detected in collection-c. One 4-byte read out-of-bounds. Now what's weird is that it's saying that it is an out of bounds memory access instead of an memory heap buffer overflow. I'd have to look at the diff more closely to try to figure this one out

It may simply be that it is both at the same time. If the pointer is at the end of the memory, accessing ptr+x can be both out of the memory and out of the chunk bounds?

zapashcanon:
After discussing with @filipeom this is good to go so I am merging.

PR #718, opened 2025-06-20
add symbolic memory.fill and memory_blit

zapashcanon:


filipeom:
It looks good to me :) I think we don't expect these to be called with symbolic arguments do we? If so this will be ok. Otherwise it will branch a lot :sweat_smile: 

Plus I think we could not do this differently atm.

PR #721, opened 2025-06-25
Call graph

Julielanor:
Command cg creates the call graph of a Wasm program.

### DIFF
--- a/src/cmd/cmd_call_graph.ml
+++ b/src/cmd/cmd_call_graph.ml
@@ -11,16 +11,6 @@ type mode =
   | Complete
   | Sound
 
-let rec find_tables acc (e : binary instr Annotated.t) =
-  match e.raw with
-  | Table_set (Raw i) | Table_fill (Raw i) | Table_copy (Raw i, _) -> i :: acc
-  | Block (_, _, exp) | Loop (_, _, exp) ->
-    List.fold_left find_tables acc exp.raw
-  | If_else (_, _, e1, e2) ->
-    let acc = List.fold_left find_tables acc e1.raw in
-    List.fold_left find_tables acc e2.raw
-  | _ -> acc
-
 let find_functions_with_func_type func_type (acc, i)
   (f : (binary func, binary block_type) Runtime.t) =
   let (Bt_raw (_, ft)) =
@@ -78,20 +68,6 @@ let build_graph mode tables funcs (g, i) (f : (binary func, 'a) Runtime.t) =
     ((i, x.id, l) :: g, i + 1)
   | Runtime.Imported x -> ((i, x.assigned_name, []) :: g, i + 1)
 
-let find_entry_points (m : Binary.Module.t) =
-  let l = Option.to_list m.start in
-  List.fold_left (fun acc (x : Binary.export) -> x.id :: acc) l m.exports.func
-
-let rec remove_tables (l1 : (int * Binary.elem) list) l2 acc =
-  match (l1, l2) with
-  | [], [] -> acc
-  | [], _ -> acc
-  | h :: t, [] -> remove_tables t [] (h :: acc)
-  | (n, e) :: t1, h2 :: t2 ->
-    if n = h2 then remove_tables t1 t2 acc
-    else if n < h2 then remove_tables t1 l2 ((n, e) :: acc)
-    else remove_tables l1 t2 acc
-
 let eval_ibinop stack nn (op : ibinop) =
   match nn with
   | S32 ->
@@ -142,6 +118,15 @@ let eval_const env exp =
   | [ Concrete_value.I32 i ] -> Ok i
   | [ _ ] -> Error (`Type_mismatch "expected int32")
 
+let eval_tables tables env =
+  let t =
+    List.map
+      (fun (n, elem) ->
+        (n, Array.of_list (List.map (eval_const env) elem.Binary.init)) )
+      tables
+  in
+  M.of_list t
+
 let build_env (env, n) (global : (Binary.global, 'a) Runtime.t) =
   match global with
   | Runtime.Local x -> (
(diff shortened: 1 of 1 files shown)


src/cmd/cmd_call_graph.ml
```diff
@@ -0,0 +1,86 @@
+(* SPDX-License-Identifier: AGPL-3.0-or-later *)
+(* Copyright © 2021-2024 OCamlPro *)
+(* Written by the Owi programmers *)
+
+open Syntax
+open Types
+open Runtime
```
redianthus: I think this `open Runtime` is not necessary as it is used only in one place. Could you remove it and write `Runtime.Local` were needed instead?

src/cmd/cmd_call_graph.ml
```diff
@@ -0,0 +1,86 @@
...
+module M = Map.Make (Int)
+
+let rec find_functions acc e =
+  match e with
+  | Call i ->
+    if List.exists (fun x -> compare_indice i x = 0) acc then acc else i :: acc
```
redianthus: The complexity here is not very good (you are potentially calling `List.exist` for each instruction of an expression).

There are many ways this can be improved.

- do not use `List.exists`, always add the element to the list, and at the end, call `List.sort_uniq` to remove duplicates;
- use a `Set` and only add to it;
- use a local Hashtbl in the following way (untested):

```ocaml
let find_functions e =
  let tbl = Hashtbl.create 512 in
  let rec loop = function
  | Call i -> Hashtbl.replace tbl i
  | ...
  in
  loop e;
  tbl (* or convert to whatever data structure you need *)
```

src/cmd/cmd_call_graph.ml
```diff
@@ -0,0 +1,86 @@
...
+      (fun (cg, acc) i ->
+        let cg', acc', s' =
+          match i with
+          | Raw i ->
+            let cg', acc' = build_graph (cg, acc) i in
+            (cg', acc', string_of_int i)
+          | _ -> (cg, acc, "")
+        in
+        (cg', String.concat "" [ acc'; string_of_int k; "->"; s'; ";" ]) )
+      (call_graph, acc) l
```
redianthus: A comment for both `build_map` and `build_graph`: can you split the printing and the data-structure building? You should really have two distinct function: one for building the data-structure, one for printing it. Otherwise, it is much harder to understand, reason about and to maintain.

src/cmd/cmd_call_graph.ml
```diff
@@ -0,0 +1,86 @@
...
+
+let pp_call_graph fmt (call_graph, entries) =
+  let _, s = List.fold_left build_graph (call_graph, "") entries in
+  Fmt.pf fmt "digraph call_graph {%s}" s
+
+let find_entry_points (m : Binary.Module.t) =
+  let l =
+    match m.start with
+    | Some i ->
+      Logs.app (fun log -> log "start : %a" Fmt.int i);
```
redianthus: Is this some debug code or is it actually needed for producing DOT or something? Anyway, if this is for debug, it should be removed (or at least us `Logs.debug`), if this is for producing DOT, then, you should have a dedicated function to output the DOT that is not mixed with code building the actual call graph.

src/cmd/cmd_call_graph.ml
```diff
@@ -0,0 +1,86 @@
...
+    Compile.File.until_validate ~unsafe:false ~rac:false ~srac:false file
+  in
+  let funcs = m.func in
+
+  let call_graph, _ = Array.fold_left build_map (M.empty, 0) funcs in
+    let entries =
+    Option.value (Option.bind (Option.bind entry_point (
+      fun x ->
+        Array.find_index (fun f -> match f with 
+                                | Local y -> String.equal x (Option.value y.id ~default:"")
```
redianthus: This code looks suspicious (using `""`). I guess you could rewrite this like:


```ocaml
Option.compare String.compare (Some x) y.id
```

src/bin/owi.ml
```diff
@@ -285,6 +285,19 @@ let c_cmd =
...
 
+(* owi call_graph *)
+
+let cg_info =
+  let doc = "Build a call graph" in
+  let man = [] @ shared_man in
+  Cmd.info "cg" ~version ~doc ~sdocs ~man
+
+let cg_cmd =
+  let+ files
```
redianthus: I think we should only take one file on input. It'll be much easier for now. (I think there's a `source_file` for this already)

src/data_structures/graph.ml
```diff
@@ -0,0 +1,44 @@
+type node =
+  { ind : int
+  ; info : string option
+  ; children : int list
+  ; mutable visited : bool
```
redianthus: This should not be part of the graph structure, but stored separately when you print the graph (for instance in a local hashtbl that is created on each new call to pp_dot)

src/data_structures/graph.ml
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+type t =
+  { nodes : node list
+  ; entry_points : int list
+  }
+
+let empty = { nodes = []; entry_points = [] }
+
+let set_entry_points graph e = { nodes = graph.nodes; entry_points = e }
```
redianthus:
```ocaml
let set_entry_points graph entry_points = { graph with entry_points }
```

Also, I believe we should add a node for each entry point even if it has no edge, right?

src/data_structures/graph.ml
```diff
@@ -0,0 +1,44 @@
...
+
+let add_edges graph i s l =
+  let n = { ind = i; info = s; children = l; visited = false } in
+  { nodes = n :: graph.nodes; entry_points = graph.entry_points }
```
redianthus:
```ocaml
{ graph with nodes = n :: graph.nodes }
```

src/cmd/cmd_call_graph.ml
```diff
@@ -4,56 +4,181 @@
...
   | _ -> acc
 
-let build_graph (g, i) f =
+let rec find_children tables acc (l : binary instr list) =
+  match l with
+  | [] -> acc
+  | Call (Raw i) :: l | Return_call (Raw i) :: l ->
+    find_children tables (i :: acc) l
+  | I32_const x :: Call_indirect (Raw i, _) :: l
+  | I32_const x :: Return_call_indirect (Raw i, _) :: l -> (
```
redianthus:
```ocaml
| I32_const x :: (Return_call_indirect (Raw i, _) | Call_indirect (Raw i, _)) :: l -> (
```

src/bin/owi.ml
```diff
@@ -7,6 +7,19 @@ open Cmdliner
 
 (* Helpers *)
 
+let call_graph_conv =
+  let of_string s =
+    match String.lowercase_ascii s with
+    | "complete" -> Ok Cmd_call_graph.Complete
+    | "sound" -> Ok Cmd_call_graph.Sound
+    | _ -> Fmt.error_msg {|Expected "complete" or "spund" but got "%s"|} s
```
redianthus: s**o**und

PR #723, opened 2025-06-27
Display available solver for `--solver` argument in the manual

filipeom:
I did something similar for smtml, so I thought it might look cool here too. The code looks kinda messy but I like how it looks in the manual:

image


zapashcanon:
I think this is much better to list the solver available indeed!

But should we list all of them, regardless of what is installed on the user's switch? I'd like to do this for three reasons:
- the user can discover more solver than what is installed (if he tries one that is not installed, he'll get the nice error message from smtml that explains how to install it, so it's fine);
- everything will be displayed on the public documentation (in the man pages);
- the tests are going to fail locally when you don't have the exact same set of solvers than the CI.

Moreover, I think it should fix the "The code looks kinda messy" haha ?

The only downside is that in the future, we may forget to add new solvers added to smtml. 

What do you think?

filipeom:
> I think this is much better to list the solver available indeed!
> 
> But should we list all of them, regardless of what is installed on the user's switch? I'd like to do this for three reasons:
> 
> * the user can discover more solver than what is installed (if he tries one that is not installed, he'll get the nice error message from smtml that explains how to install it, so it's fine);
> * everything will be displayed on the public documentation (in the man pages);
> * the tests are going to fail locally when you don't have the exact same set of solvers than the CI.

Ah yes, I agree! All very good reasons :smiley: 

> Moreover, I think it should fix the "The code looks kinda messy" haha ?
> 
> The only downside is that in the future, we may forget to add new solvers added to smtml.
> 
> What do you think?

Yeah, I think so. Regarding this:

> The only downside is that in the future, we may forget to add new solvers added to smtml.

Maybe I can have a list `Smtml.Solver_dispatcher.supported_solvers` which contains all supported solvers by smtml, this way it avoids having to maintain a list of available solvers here as well.

zapashcanon:
> Maybe I can have a list Smtml.Solver_dispatcher.supported_solvers which contains all supported solvers by smtml, this way it avoids having to maintain a list of available solvers here as well.

Indeed, I think it would be the cleanest option!

zapashcanon:
@filipeom, I think this can be changed to use the new smtml function now?

filipeom:
> @filipeom, I think this can be changed to use the new smtml function now?

Updated the PR with these changes. What do you think of having a subcommand like:

```sh
$ owi sym list-solver
``` 

To check currently installed solvers? Is it a helpful thing, or not so much?

filipeom:
> I think `owi version` gives you the list already?

Ah indeed it does! Nice, I didn't notice the change to the version command! Very cool :smiley:

PR #734, opened 2025-07-10
instruction counter

Julielanor:


src/interpret/interpret.ml
```diff
@@ -83,13 +83,13 @@ module Make (P : Interpret_intf.P) :
 
   let page_size = const_i64 65_536L
 
-  let pop_choice stack =
+  let pop_choice stack prio_t prio_f =
```
redianthus: Can you use named arguments here? (`~prio_t` and `prio_f`)

src/interpret/interpret.ml
```diff
@@ -807,10 +807,15 @@ module Make (P : Interpret_intf.P) :
...
+
+    Atomic.incr instr.Annotated.nb_iter;
     let st stack = Choice.return (State.Continue { state with stack }) in
     Logs.info (fun m -> m "stack         : [ %a ]" Stack.pp stack);
     Logs.info (fun m ->
-      m "running instr : %a" (Types.pp_instr ~short:true) instr.Annotated.raw );
+      m "running instr : %a (%a)"
+        (Types.pp_instr ~short:true)
+        instr.Annotated.raw Fmt.int
+        (Atomic.get instr.Annotated.nb_iter) );
```
redianthus: We'll have to remove this before merging

src/interpret/interpret.ml
```diff
@@ -936,7 +936,12 @@ module Make (P : Interpret_intf.P) :
     | Br (Raw i) -> State.branch state i
     | Br_if (Raw i) ->
       let* b, stack =
-        pop_choice stack ~counter_next_true:0 ~counter_next_false:0
+        let counter_next_false =
+          match state.pc.raw with
+          | [] -> Int.max_int
```
redianthus: So, this is going to run with a very low priority, right?

src/cmd/cmd_call_graph.ml
```diff
@@ -24,20 +24,27 @@ let rec find_children mode tables funcs acc (l : binary instr Annotated.t list)
...
+        } as instr )
       :: l
     , Complete ) ->
-    let acc, _ =
-      Array.fold_left (find_functions_with_func_type ft) (acc, 0) funcs
+    let children, _ =
+      Array.fold_left (find_functions_with_func_type ft) ([], 0) funcs
     in
+    Annotated.update_functions_called instr children;
+    let acc = children @ acc in
```
redianthus: Using an other data structure here would reduce the complexity (`Set` maybe?)

src/cmd/cmd_call_graph.ml
```diff
@@ -220,15 +230,34 @@ let build_call_graph_from_text_module call_graph_mode modul entry_point =
   | Ok m -> build_call_graph call_graph_mode m entry_point
   | _ -> assert false
 
-let cmd ~call_graph_mode ~source_file ~entry_point =
+let compute_distances m entry_point =
+  let call_graph = build_call_graph Sound m entry_point in
+  let _ = Graph.compute_distance_to_unreachable_cg call_graph in
```
redianthus: You should never write `let _ = f ... in` because if you suddenly add a new argument to `f`, then you will partially apply it and the type-checker won't help you notice it, which can lead to quite subtle and hard to spot bugs.

The usual solution is to add a type annotation like:

```ocaml
let _ : whatever = f ... in
```

This way, if you add an argument to `f`, the type will become `something -> whatever` and the type-checker will complain.

src/cmd/cmd_call_graph.ml
```diff
@@ -220,15 +230,34 @@ let build_call_graph_from_text_module call_graph_mode modul entry_point =
...
+
+    Ok ()
+  else
+    let* () =
+      Bos.OS.File.writef
+        (Fpath.set_ext ".dot" source_file)
+        "%a" Graph.pp_cg call_graph
+    in
+
+    Ok ()
```
redianthus:
```ocaml
  if scc then
    let scc = Graph.build_scc_graph call_graph in
      Bos.OS.File.writef
        (Fpath.set_ext ".dot" source_file)
        "%a" Graph.pp_scc_graph (scc, Graph.Cg)
  else
      Bos.OS.File.writef
        (Fpath.set_ext ".dot" source_file)
        "%a" Graph.pp_cg call_graph
```

Should be the same?

src/cmd/cmd_cfg.ml
```diff
@@ -77,65 +94,91 @@ let rec build_graph (l : binary expr) nodes n node edges
...
-  Fmt.pf fmt "digraph cfg {\n rankdir=LR;\n node [shape=record];\n %a;\n %a}"
-    pp_nodes nodes pp_edges edges
+let build_cfg_from_text_module modul =
+  let m =
+    Compile.Text.until_validate ~unsafe:false ~rac:false ~srac:false modul
+  in
+  match m with
+  | Ok m ->
+    let f =
+      match Array.get m.func 0 with Runtime.Local f -> f | _ -> assert false
```
redianthus: Why are you using `0` here?

src/cmd/cmd_cfg.ml
```diff
@@ -77,65 +94,91 @@ let rec build_graph (l : binary expr) nodes n node edges
...
       (Option.bind entry_point (fun x ->
-         Array.find_index
-           (fun f ->
-             match f with
-             | Runtime.Local y ->
-               Option.compare String.compare (Some x) y.id = 0
-             | _ -> false )
-           m.func ) )
+         match int_of_string_opt x with
+         | Some x -> Some x
```
redianthus:
```ocaml
| (Some _) as x -> x
```

is better as it avoids an allocation. Without the parenthesis it can be written:

```ocaml
| Some _ as x -> x
```

src/cmd/cmd_cfg.ml
```diff
@@ -77,65 +94,91 @@ let rec build_graph (l : binary expr) nodes n node edges
...
+         | None ->
+           Array.find_index
+             (fun f ->
+               match f with
```
redianthus:
```suggestion
             (function
```

src/cmd/cmd_cfg.ml
```diff
@@ -77,65 +94,91 @@ let rec build_graph (l : binary expr) nodes n node edges
...
+        "%a" Graph.pp_scc_graph (scc, Graph.Cfg)
+    in
+    Ok ()
+  else
+    let* () =
+      Bos.OS.File.writef
+        (Fpath.set_ext ".dot" source_file)
+        "%a" Graph.pp_cfg graph
+    in
+    Ok ()
```
redianthus: Same here, I think you can remove the `let* ()` and the `Ok ()` in both branches

src/cmd/cmd_cfg.mli
```diff
@@ -1 +1,8 @@
-val cmd : source_file:Fpath.t -> entry_point:string option -> unit Result.t
+val build_cfg_from_text_module :
+  Text.modul -> Types.binary Types.instr Annotated.t list Graph.t
```
redianthus: Isn't this:

```ocaml
Text.modul -> Types.binary Types.expr Graph.t
```

?

(same below)

src/cmd/cmd_sym.ml
```diff
@@ -39,6 +40,7 @@ type parameters =
 let run_file ~parameters ~source_file =
   let { unsafe; rac; srac; entry_point; invoke_with_symbols; _ } = parameters in
   let* m = Compile.File.until_validate ~unsafe ~rac ~srac source_file in
+  Cmd_call_graph.compute_distances m entry_point;
```
redianthus: We should do this *only* if the `Smart` strategy was given, otherwise it'll makes other strategies slow for no good reason.

src/data_structures/annotated.ml
```diff
@@ -1,15 +1,53 @@
-type 'a t = { raw : 'a }
+type 'a t =
+  { raw : 'a
+  ; instr_counter : int Atomic.t
+  ; mutable functions_called : int list
+  ; mutable distances : int array
```
redianthus: Instead of using a `mutable` + `array` (which are both mutable), I would suggest to use either a `Map.t ref`, or simpler, a `Dynarray` (without the `mutable` field). (Same for the other fields)

src/data_structures/annotated.ml
```diff
@@ -1,15 +1,53 @@
...
+  ; mutable d_true : int array
+  ; mutable d_false : int array
```
redianthus: What are the purposes of these three fields? My understanding is that `d_true` and `d_false` will only be used to help computing `distances`, right? If this is the case, it may be better to try to put them somewhere else.

src/data_structures/annotated.ml
```diff
@@ -1,15 +1,53 @@
...
+  }
 
-let dummy raw = { raw }
+let dummy raw =
+  { raw
+  ; instr_counter = Atomic.make 0
+  ; functions_called = []
+  ; distances = Array.make 0 0
```
redianthus: You should avoid using `Array.make` because if you use a value that is not a small scalar (`int`, `bool`...) but something that is mutable and boxed (e.g. `ref 0`) then, all elements of the array will point to the same memory location, and modifying one of them will modify all the others. Instead, if it recommended to use `Array.init`. Here you don't need any of them as the empty array should be enough `[||]`.

src/data_structures/annotated.ml
```diff
@@ -1,15 +1,53 @@
...
-let iter f { raw } = f raw
+let iter f { raw; _ } = f raw
+
+let update_functions_called annot l = annot.functions_called <- l
+
+let init_distances annot a = annot.distances <- a
+
+let init_d_true annot a = annot.d_true <- a
+
+let init_d_false annot a = annot.d_false <- a
```
redianthus: I find the name `init` confusing, maybe `set` instead (or `update`) ?

src/data_structures/graph.ml
```diff
@@ -1,61 +1,118 @@
...
+let init_cfg nodes edges =
+  let children_map, parents_map =
+    List.fold_left
+      (fun (m_children, m_parents) (n1, n2, s) ->
+        (M.add_to_list n1 (n2, s) m_children, M.add_to_list n2 n1 m_parents) )
+      (M.empty, M.empty) edges
+  in
+  let l =
+    List.rev
+      (List.map
```
redianthus: You can use the more efficient `List.rev_map` here

src/data_structures/graph.ml
```diff
@@ -1,61 +1,118 @@
...
-let pp_dot fmt g =
-  let entry_points =
-    List.concat_map
-      (fun x ->
-        Option.to_list
-          (if x < Array.length g.nodes then Some (Array.get g.nodes x) else None) )
-      g.entry_points
+let pp_node_cg fmt (n : 'a node) =
+  if snd n.info then
+    Fmt.pf fmt "%a ;\n %a" Fmt.int n.ind pp_edges (n.ind, n.children)
```
redianthus: Instead of `"%a" Fmt.int x` you can use `"%d" x` directly

src/data_structures/graph.ml
```diff
@@ -1,61 +1,118 @@
...
+
+let pp_nodes_cg fmt n =
+  Fmt.array ~sep:(fun fmt () -> Fmt.pf fmt "") pp_node_cg fmt n
+
+let pp_cg_graph fmt g = Fmt.pf fmt "%a" pp_nodes_cg g.nodes
+
+let pp_cg fmt g = Fmt.pf fmt "digraph call_graph {\n %a}" pp_cg_graph g
```
redianthus: cc @S41d

You could use some boxes here to get nice indenting

src/data_structures/graph.ml
```diff
@@ -1,61 +1,118 @@
...
+         nodes )
   in
-  let l, _ = List.fold_left (print_graph g.nodes) ([], S.empty) entry_points in
-  Fmt.pf fmt "digraph call_graph {\n%a;\n%a}" pp_entry_points g.entry_points
-    pp_edges l
+  { nodes = Array.of_list l; entry_points = [] }
+
+let pp_inst fmt i = Fmt.pf fmt "%a" (Types.pp_instr ~short:true) i.Annotated.raw
+
+let pp_exp fmt l = Fmt.list ~sep:(fun fmt () -> Fmt.pf fmt " | ") pp_inst fmt l
```
redianthus:
```suggestion
let pp_exp fmt l = Fmt.list ~sep:(fun fmt () -> Fmt.string fmt " | ") pp_inst fmt l
```

Can be used when you don't use any of `%a`, `%d` etc. in your formatting string and it should be more efficient (this is nitpicking, feel free to ignore)

src/data_structures/graph.ml
```diff
@@ -1,61 +1,118 @@
...
+
+let pp_node_cfg fmt (n : 'a node) =
+  Fmt.pf fmt {|%a [label="%a"]; %a|} Fmt.int n.ind pp_exp (List.rev n.info)
+    pp_edges (n.ind, n.children)
+
+let pp_nodes_cfg fmt g =
+  Fmt.array ~sep:(fun fmt () -> Fmt.pf fmt "") pp_node_cfg fmt g
```
redianthus:
```suggestion
  Fmt.array ~sep:Fmt.nop pp_node_cfg fmt g
```

src/data_structures/graph.ml
```diff
@@ -79,4 +136,367 @@ let is_subgraph graph subgraph =
   compare_nodes (Array.to_list graph.nodes) (Array.to_list subgraph.nodes)
   && res
 
-let get_info graph = (Array.get graph.nodes 1).info (* just to use info *)
+module IntPair = struct
```
redianthus, line 139: Could you add a comment saying this should be replaced by the brand new [`Pair`](https://github.com/ocaml/ocaml/blob/trunk/stdlib/pair.mli#L18) module in 5.4?

src/data_structures/graph.ml
```diff
@@ -79,4 +136,367 @@ let is_subgraph graph subgraph =
...
+            l
+        | None -> () )
+      node.parents
+  | None, _ -> assert false
+
+let compute_distance_to_unreachable_cg graph =
+  let unreachables, calls = find_unreachables_cg graph in
+  let nb_unr = List.length unreachables in
+  let distances =
+    Array.init (Array.length graph.nodes) (fun i ->
```
redianthus, line 252: `Array.map(i)` ?

src/data_structures/graph.ml
```diff
@@ -79,4 +136,367 @@ let is_subgraph graph subgraph =
...
+        match M.find_opt w visited with
+        | None ->
+          let num, stack, partition, visited = explore graph acc w in
+          let w =
+            match M.find_opt w visited with Some w -> w | None -> assert false
+          in
+          let visited =
+            M.update node
+              (fun v_opt ->
+                Option.bind v_opt (fun v ->
```
redianthus:
```suggestion
              (function None -> None
                | Some v ->
```

src/data_structures/graph.ml
```diff
@@ -79,4 +136,367 @@ let is_subgraph graph subgraph =
...
+                    } ) )
+              visited
+          in
+          (num, stack, partition, visited)
+        | Some w ->
+          if w.in_stack then
+            let visited =
+              M.update node
+                (fun v_opt ->
+                  Option.bind v_opt (fun v ->
```
redianthus: same here

src/data_structures/graph.mli
```diff
@@ -1,9 +1,30 @@
...
+val init_cg :
+  (int * 'a option * int list) list -> int list -> ('a option * bool) t
 
-val is_subgraph : t -> t -> bool
+val pp_cg : Format.formatter -> ('a option * bool) t -> unit
 
-val get_info : t -> string option
+val init_cfg : (int * 'a) list -> (int * int * string option) list -> 'a t
+
+val pp_cfg : Format.formatter -> 'a Types.instr Annotated.t list t -> unit
```
redianthus: `Types.expr` ?

src/data_structures/graph.mli
```diff
@@ -1,9 +1,30 @@
-type t
+type 'a t
 
-val init : (int * string option * int list) list -> int list -> t
+type _ g =
+  | Cg : ('a option * bool) t g
+  | Cfg : Types.binary Types.instr Annotated.t list t g
```
redianthus: `Types.expr` ?

test/analyze/control-flow-graph/br_table.wat
```diff
@@ -0,0 +1,18 @@
...
+    (local.get $value)
+    (global.set $A)
+    (local.get $value)
+    (i32.const 42)
+    (i32.eq)
+    (br_if 0)
+    (return)) 
+  (local.get $value)
+  (global.set $B))
+)
```
redianthus: missing line at the end (I think you should configure your editor to automatically add it, they usually all offer this option: https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files )

src/data_structures/annotated.ml
Julielanor: It's the opposite. I removed `distances` because it was not used but `d_true` and `d_false` are used when select is called two differentiate the true and false branches.

src/interpret/interpret.ml
```diff
@@ -918,7 +946,20 @@ module Make (P : Interpret_intf.P) :
     end
     | Br (Raw i) -> State.branch state i
     | Br_if (Raw i) ->
-      let* b, stack = pop_choice stack in
+      let* b, stack =
+        let counter_next_false, counter_next_true =
+          match state.pc.raw with
+          | [] -> Int.max_int, Atomic.get instr.Annotated.instr_counter -1
```
redianthus, line 952: I think we should make this an option rather than using `Int.max_int` and choose much later which value we should use for the `None` case. By much latter, I mean it should be done in the `Prio` module

Issue #566, opened 2025-03-25
assert false reached


zapashcanon:
With the following:

```c
#include <owi.h>

void f(int n, long int m, float x, double y) {
    owi_assume(n > m);
    owi_assume(m > x);
    owi_assume(x > y);
    owi_assume(y > n);
    owi_assert(0);
}
```

And running: `owi c --entry-point=f --invoke-with-symbols ./invoke_with_symbols.c --debug`

filipeom:
related to https://github.com/OCamlPro/owi/issues/519#issuecomment-2682983796 I think

krtab:
I found a similar variation of this bug

```rust
use owi_sym::{Symbolic};

fn main() {
    let x : u32 = Symbolic::symbol();
    // Unsat assume
    owi_sym::assume(if x % 10 == 0 {x % 5 != 0} else {x % 20 == 0});
    owi_sym::assert(false);
}
```

PC slicing makes it kinda worse because it kinda assumes that the path condition is always sat (the unsat parts may be sliced out before being sent to the solver when checking).

I think the solution is to reuse things I did around PC freshness in 9266b9b566ee24bc216db9f160efe49a1a64158f (from PR #544)

PR #737, opened 2025-07-10
Fixes around pc freshness and reachability

krtab:
1. Ensure that we are reachable before adding and slicing (2nd commit)
2. Ensure we are reachable before failing an assert

Fix: #566 (both reproducing examples)

### DIFF
--- a/src/symbolic/symbolic_choice.ml
+++ b/src/symbolic/symbolic_choice.ml
@@ -382,7 +382,7 @@ module Make (Thread : Thread_intf.S) = struct
     let* () = yield in
     let* solver in
     match Solver.check solver pc with
-    | `Sat -> return ()
+    | `Sat -> modify_thread Thread.mark_pc_fresh
     | `Unsat -> stop
     | `Unknown -> assert false
 
@@ -413,6 +413,7 @@ module Make (Thread : Thread_intf.S) = struct
     | `Unsat -> stop
     | `Sat -> begin
       let symbol_scopes = Symbol_scope.of_symbol symbol in
+      let* () = modify_thread Thread.mark_pc_fresh in
       (* TODO: we are doing the check two times here, because Solver.model is also doing it, we should remove it! *)
       let model = Solver.model solver ~symbol_scopes ~pc in
       match Smtml.Model.evaluate model symbol with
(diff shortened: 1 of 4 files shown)


src/symbolic/symbolic_choice.ml
```diff
@@ -402,6 +411,7 @@ module Make (Thread : Thread_intf.S) = struct
     | `Unsat -> stop
     | `Sat -> begin
       let symbol_scopes = Symbol_scope.of_symbol symbol in
+      let* () = modify_thread Thread.mark_pending_pc_done in
```
redianthus, line 414: This should be done before calling the solver, no? Because otherwise, if the things newly added to the PC are unsat, we'll crash instead of stopping.

src/symbolic/symbolic_choice.ml
```diff
@@ -378,20 +378,29 @@ module Make (Thread : Thread_intf.S) = struct
...
-    match Solver.check solver sliced_pc with
-    | `Sat -> return ()
-    | `Unsat -> stop
-    | `Unknown -> assert false
+    match Thread.pending_pc thread with
+    | [] -> return ()
+    | hd :: tl -> begin
+      let pc_to_check = List.fold_left Symbolic_value.Bool.and_ hd tl in
+      let pc_to_check =
+        Symbolic_path_condition.slice (Thread.pc thread) pc_to_check
```
redianthus, line 400: A call to `slice` can only be done if there was a call to `Symbolic_path_condition.add` before.

src/symbolic/symbolic_choice.ml
```diff
@@ -420,18 +430,19 @@ module Make (Thread : Thread_intf.S) = struct
...
       let false_branch =
         let neg_v = Symbolic_value.Bool.not v in
-        let* () = add_pc neg_v in
         let* () = add_breadcrumb 0 in
-        let+ () = check_reachability neg_v in
+        let* () = add_pc neg_v in
+        let+ () = check_reachability in
         false
       in
+      let* () = check_reachability in
```
redianthus, line 445: I don't understand the purpose one this check

src/symbolic/symbolic_choice.ml
```diff
@@ -484,18 +495,22 @@ module Make (Thread : Thread_intf.S) = struct
         in
         let this_val_branch =
           let* () = add_breadcrumb (Int32.to_int i) in
-          let+ () = add_pc this_value_cond in
+          let* () = add_pc this_value_cond in
+          let+ () = modify_thread Thread.mark_pending_pc_done in
```
redianthus, line 499: So this is adding something to the pending PC and then emptying the pending PC immediately?

PR #769, opened 2025-09-22
add priority

redianthus:
TODO:

- [x] use ocamlgraph
- [ ] use `Pqueue` from OCaml 5.4
- [ ] parameterize the target (`unreachable` for `--fail-on-trap-only`, `call $owi_assert` for `--fail-on-assertion-only`, both by default)
- [ ] remove the use of `Prio.default` wherever it is possible
- [ ] reimplement the distance to target computation
- [ ] rewrite the CG/CFG calculation so that it is cleaner
- [ ] add other simpler strategies like --least-used, or even `--least-used-then-{fifo,lifo,random}`


redianthus:
I'm merging to stop the rebase hell here. I'll finish the work in subsequent PRs.

PR #794, opened 2025-10-23
Time

### LINKED ISSUE #762: allows to measure execution time of all steps
1. Compiling C/Rust/... to Wasm
2. Parsing Wasm
3. Analyzing Wasm (typechecking etc.)
4. Execution

S41d:
Some things I'm_ not sure about are:
- the placement of the timing function call for type checking. Right now we see the type checking time appear twice.
- where does this whole thing come from : 
```
val module_to_run : '_weak1 Link.module_to_run =
  {Owi.Link.id = Some "quickstart"; env = <abstr>;
   to_run =
    [{Owi.Annotated.raw =
       [{Owi.Annotated.raw = Owi.Types.Call (Owi.Types.Raw 0)}]}]}
val link_state : '_weak1 Link.state =
  {Owi.Link.by_name = <abstr>; by_id = <abstr>;
   last =
    Some
     ({Owi.Link.globals = <abstr>; memories = <abstr>; tables = <abstr>;
       functions = <abstr>; defined_names = <abstr>},
      <abstr>);
   collection = <abstr>; envs = <abstr>}
```

Closes #762


doc/src/ocaml-api/overview.md
```diff
@@ -26,8 +26,21 @@ mdx_gen.bc.exe: [INFO] parsing      ...
...
+    [{Owi.Annotated.raw =
+       [{Owi.Annotated.raw = Owi.Types.Call (Owi.Types.Raw 0)}]}]}
+val link_state : '_weak1 Link.state =
+  {Owi.Link.by_name = <abstr>; by_id = <abstr>;
+   last =
+    Some
+     ({Owi.Link.globals = <abstr>; memories = <abstr>; tables = <abstr>;
+       functions = <abstr>; defined_names = <abstr>},
+      <abstr>);
+   collection = <abstr>; envs = <abstr>}
```
redianthus: I think you can simply replace all of this by `...` and the test should pass again. You promoted once to add the typecheking time and it promoted everything instead of keeping the "...", this is a common issue.

Otherwise you could simply disable the benchmarking stuff by default so that it does not get printed, which is probably something we should do.

src/utils/log.ml
```diff
@@ -0,0 +1,22 @@
+let main_src = Logs.Src.create "owi" ~doc:"Owi's main logs"
+
+let bench_src = Logs.Src.create "owi-bench" ~doc:"Owi's benchmark logs"
+
+module Main_log = (val Logs.src_log main_src : Logs.LOG)
+
+include Main_log
```
redianthus: I think you can put this include directly when defining `module Main_log = ...`

src/dune
```diff
@@ -121,7 +121,8 @@
   wpq
   wq
   ws
-  wutf8)
+  wutf8
+  log)
```
redianthus: can you keep it sorted? (it is easier to find/delete stuff)

src/utils/log.ml
```diff
@@ -0,0 +1,22 @@
...
+
+module Bench_log = (val Logs.src_log bench_src : Logs.LOG)
+
+let bench_fn str fn =
+  let c = Mtime_clock.counter () in
+  let r = fn () in
+  Bench_log.info (fun m -> m "%s : %a" str Mtime.Span.pp (Mtime_clock.count c));
+  r
```
redianthus: OK so IIUC, here, `fn` is something that must be called regardless of if we are in benchmarking mode or not, right?

If this is the case, I am a little bit worried of calling `Mtime_clock.counter` each time. It seems you are only using it on "big" functions so this is probably fine currently, but I am wondering if we could make this a little more explicit by changing the name to `bench_slow_fn` or whatever.

PR #796, opened 2025-10-27
Bump the smtml version to 0.13.0 and print path condition in the smtml format

### LINKED ISSUE #631: print the smtlib request sent to SMT in debug mode

hra687261:
I initially wanted to print the query right before the call to the solver in `Solver.check` and `Solver.model`, but with parallelism, the debugging messages become unreadable.

Closes #631

(It can wait for https://github.com/ocaml/opam-repository/pull/28782 and https://github.com/NixOS/nixpkgs/pull/456420 to not have to pin the dep on smtml)

### DIFF
--- a/src/interpret/interpret.ml
+++ b/src/interpret/interpret.ml
@@ -815,12 +815,10 @@ module Make (P : Interpret_intf.P) :
       match Logs.Src.level Log.main_src with
       | Some Logs.Debug ->
         let+ pc = Choice.get_pc () in
-        Log.debug (fun m ->
-          m "path condition smt query:%a"
-            (fun fmt -> function
-              | [] -> Fmt.pf fmt " empty"
-              | l -> Fmt.pf fmt "@\n @[<v>%a@]" Smtml.Expr.pp_smtml l )
-            (Smtml.Expr.Set.to_list pc) )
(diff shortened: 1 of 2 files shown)


redianthus:
> But I don't mind putting it under the debug flag (one less flag and one less argument for multiple functions), you'd be ok with simply replacing m "path condition: [ %a ]" Smtml.Expr.pp_list with m "path condition smt query:@\n @[<v>%a@]" Smtml.Expr.pp_smtml?

Yes please go ahead, this looks good to me. It is true that most of the time I'm only interested in parts of the PC that are being sent to the solver. And it will be much shorter. :)

src/interpret/interpret.ml
```diff
@@ -816,7 +816,10 @@ module Make (P : Interpret_intf.P) :
       | Some Logs.Debug ->
         let+ pc = Choice.get_pc () in
         Log.debug (fun m ->
-          m "path condition: [ %a ]" Smtml.Expr.pp_list
+          m "path condition smt query:%a"
+            (fun fmt -> function
+              | [] -> Fmt.pf fmt " empty"
```
redianthus: What about not printing anything when it is empty?

hra687261: Done!

PR #797, opened 2025-10-27
Compute and print total of the statistics of the used solver when the `--bench` flag is set

hra687261:
Prints the sum of the stats of the used solver in debug mode. 
Part of the work needed for https://github.com/OCamlPro/owi/issues/184 

Edit: Depends on #798

### DIFF
--- a/src/cmd/cmd_iso.ml
+++ b/src/cmd/cmd_iso.ml
@@ -371,14 +371,14 @@ let cmd ~deterministic_result_order ~fail_mode ~exploration_strategy ~files
   let funcexports1 =
     module1.exports.func
     |> List.map (fun { Binary.name; id } ->
-      let typ = Binary.Module.get_func_type id module1 in
-      (name, typ) )
+         let typ = Binary.Module.get_func_type id module1 in
+         (name, typ) )
   in
   let funcexports2 =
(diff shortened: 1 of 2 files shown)


src/cmd/cmd_conc.ml
```diff
@@ -305,10 +310,14 @@ let find_node_to_run tree =
 let pc_model solver pc =
   let pc = Concolic_choice.pc_to_exprs pc in
   match Solver.check solver pc with
-  | `Unsat | `Unknown -> None
+  | `Unsat | `Unknown ->
+    (* TODO: fetch stats in this case as well? *)
+    None
```
redianthus: FYI the concolic mode is broken. We don't really care about it so it does not really matter. :)

hra687261: Yes, what I meant by the comment is that we should probably also fetch and count the stats when the solver answer SAT (or just UNSAT).
After checking with z3 (which is the only solvers that provides stats for now), its sufficient to get the stats at the end because the stats are visibly not back-tracked, so the solver itself accumulates the stats of all the calls it receives.
So the current version, which gets the stats after the get_model call, is sufficient for now.

PR #802, opened 2025-10-29
show solver time in bench mode

S41d:


src/cmd/cmd_sym.ml
```diff
@@ -251,6 +251,11 @@ let handle_result ~exploration_strategy ~workers ~no_stop_at_failure ~no_value
       ~count_acc:0 ~results ~with_breadcrumbs
   in
   Log.info (fun m -> m "Completed paths: %d" (Atomic.get path_count));
+
+  let bench_stats = Thread_with_memory.bench_stats thread in
```
redianthus: Can you move this inside the second `Log.bench` so that it is not executed when not in benchmarking mode?

src/cmd/cmd_sym.ml
```diff
@@ -251,6 +251,11 @@ let handle_result ~exploration_strategy ~workers ~no_stop_at_failure ~no_value
...
+  Log.bench (fun m -> m "Benchmark : ");
+  Log.bench (fun m ->
```
redianthus, line 256: Can you merge the two?

```ocaml
  Log.bench (fun m ->
    m "Benchmark : ";
    m "solver time: %a" Mtime.Span.pp (Atomic.get bench_stats.solver_time) );

Issue #611, opened 2025-04-15
benchmark `use_ite_for_select`


redianthus:
in interpret.ml, it is set to true for now, if both prove useful, a CLI flag could be offered

redianthus:
Yes, I think adding the flag is worth it.

To get a better idea of the results, could you run [this tool](https://github.com/OCamlPro/symbocalypse/tree/main/diff) on the two results file you got? In particular, I am interested in seeing the execution time variation only on goals solved by both tools. For now I'd say we should set the option to true by default, but I may change my mind after seeing the diff results.

I think there are two interesting questions (to which I don't have an answer :-)):

1. Could we write some heuristics to decide if `use_ite_for_select` is interesting, either on a given program or on a given `select` instruction?
2. Could we use an `ite` in other places? (This one is probably easier to answer).

hra687261:
Ok! Here's the output of the diff tool:
```
tool1 had 672 reached and tool2 had 671 reached
tool1 had 501 timeout and tool2 had 503 timeout
tool1 had 001 nothing and tool2 had 001 nothing
tool1 had 037 other   and tool2 had 036 other
tool1 had 000 killed  and tool2 had 000 killed

tools have 665 reached tasks in common

tool1 had 007 tasks tool2 did not found
tool2 had 006 tasks tool1 did not found

on      commonly  reached tasks, tool 1 took 1595.700267 sec. (mean 2.399549, median 0.548621, min 0.306081, max 27.683100) and tool 2 took 1575.438922 sec. (mean 2.369081, median 0.542383, min 0.321276, max 29.703400)
on *not commonly* reached tasks, tool 1 took 131.851030 sec. (mean 18.835861, median 20.340000, min 6.878930, max 28.258600) and tool 2 took 40.464773 sec. (mean 6.744129, median 9.951390, min 0.592530, max 17.833700)

among tasks reached only by tool 1, tool 2 replied 000 nothing, 007 timeout, 00 other and 00 killed
among tasks reached only by tool 2, tool 1 replied 000 nothing, 003 timeout, 03 other and 00 killed
tasks solved only by tool 1:
  testcomp/sv-benchmarks/c/xcsp/CostasArray-12.c
  testcomp/sv-benchmarks/c/xcsp/AllInterval-016.c
  testcomp/sv-benchmarks/c/seq-mthreaded/pals_lcr.5_overflow.ufo.UNBOUNDED.pals.c
[...]

PR #812, opened 2025-11-03
Add flag to unset use-ite-for-select

hra687261:
See #611.
I am open to suggestion for the flag name and where to place the global variable.

### DIFF
--- a/src/owi.mli
+++ b/src/owi.mli
@@ -987,6 +987,8 @@ module Binary_validate : sig
 end
 
 module Interpret : sig
+  val unset_use_ite_for_select : unit -> unit
+
   module Concrete : sig
     val modul :
          timeout:float option


redianthus:
@hra687261 we'll have to try to find the bug when you come back. Also this is conflicting with #801 but it'll be easy to fix (rebase and on the conflicting commit, simply regenerate the doc with `dune runtest && dune promote`).

hra687261:
The culprit is `dobble.c`, simply running htop reveals it since it seems to take over 5 minutes or loop forever, the slowdown is reproducible with:
```
dune exec -- owi c -O1 doc/src/sap/dobble.c -w1 --no-value --no-assert-failure-expression-printing 
```
I will try to understand what is going on, but if you have the time you can take a look also @redianthus

redianthus:
Thanks!

PR #816, opened 2025-11-06
count the interpreter time in bench mode

### LINKED ISSUE #779: benchmarks time spent in solver vs. time spent in interpreter (without the solver time)

S41d:
I think we can close #779 with this pr

### DIFF
--- a/test/sym/bench.t
+++ b/test/sym/bench.t
@@ -1,4 +1,4 @@
-  $ owi sym div_i32.wat --no-value --deterministic-result-order -w1 --bench 2>&1 | sed 's/\(solver\|interpreter\|parsing\|execution\|typechecking\) \(time\).*/\1 \2/'
+  $ owi sym div_i32.wat --no-value --deterministic-result-order -w1 --bench 2>&1 | sed 's/\(solver\|interpreter\|parsing\|execution\|typechecking\) \(time\).*/\1 \2/' | sed 's/\(stat \("[^"]*"\|\w*\)\).*/\1/'
   owi: [INFO] parsing time
   owi: [INFO] typechecking time
   owi: [INFO] execution time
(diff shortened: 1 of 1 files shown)


src/cmd/cmd_replay.ml
```diff
@@ -221,7 +221,20 @@ let run_file ~unsafe ~entry_point ~invoke_with_symbols filename model =
   let* m, link_state =
     Compile.Binary.until_link ~unsafe link_state ~name:None m
   in
-  Interpret.Concrete.modul ~timeout:None ~timeout_instr:None link_state.envs m
+  if Log.is_bench_enabled () then
+    let before_time = (Unix.times ()).tms_utime in
```
redianthus: We should discuss about if we want to use Unix or rather try `rusage` as it is done here https://github.com/OCamlPro/symbocalypse/blob/main/tool/tool.ml#L61-L74

src/cmd/cmd_replay.ml
```diff
@@ -221,7 +221,20 @@ let run_file ~unsafe ~entry_point ~invoke_with_symbols filename model =
...
+    let+ () =
+      Interpret.Concrete.modul ~timeout:None ~timeout_instr:None link_state.envs
+        m
+    in
+    let after_time = (Unix.times ()).tms_utime in
+    Some (after_time -. before_time)
```
redianthus: Is it reached if we find a bug?

test/sym/bench.t
```diff
@@ -0,0 +1,51 @@
+  $ owi sym div_i32.wat --no-value --deterministic-result-order -w1 --bench 2>&1 | sed 's/\(solver\|interpreter\|Parsing\|execution\|Typechecking\) \(time\).*/\1 \2/'
```
redianthus:
```
$ owi sym div_i32.wat --no-value --deterministic-result-order -w1 --bench 2>&1 | sed 's/\(solver\|interpreter\|Parsing\|execution\|Typechecking\) \(time\).*/\1 \2/' | grep -v "stat "
```

To remove the solver stats that are likely not going to be deterministic

test/sym/bench.t
```diff
@@ -0,0 +1,51 @@
...
+  owi: [INFO] Parsing time
```
redianthus: I think you have to run the tests again and promote them because you changed "Parsing" to "parsing" in another commit

PR #829, opened 2025-11-19
show number of paths in bench mode

### LINKED ISSUE #777: benchmarks number of paths

S41d:
closes #777

### DIFF
--- a/src/symbolic/symbolic_choice.ml
+++ b/src/symbolic/symbolic_choice.ml
@@ -466,6 +466,8 @@ module Make (Thread : Thread_intf.S) = struct
 
       let true_branch = branch v true prio_true in
       let false_branch = branch (Symbolic_value.Bool.not v) false prio_false in
+      let* thread in
+      Thread.incr_path_count thread;
       if explore_first then choose true_branch false_branch
       else choose false_branch true_branch
   [@@inline]
@@ -525,6 +527,8 @@ module Make (Thread : Thread_intf.S) = struct
           let* () = add_pc not_this_value_cond in
           generator ()
         in
+        let* thread in
+        Thread.incr_path_count thread;
         choose this_val_branch not_this_val_branch
       in
       generator ()
--- a/src/symbolic/thread.ml
+++ b/src/symbolic/thread.ml
@@ -46,7 +46,9 @@ struct
     let breadcrumbs = [] in
     let labels = [] in
     let bench_stats : Benchmark.stats =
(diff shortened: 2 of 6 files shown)


src/symbolic/thread.ml
```diff
@@ -110,7 +112,9 @@ struct
     let tables = Symbolic_table.clone tables in
     let globals = Symbolic_global.clone globals in
     let bench_stats : Benchmark.stats =
-      { solver_time = bench_stats.solver_time }
+      { solver_time = bench_stats.solver_time
+      ; path_count = bench_stats.path_count
+      }
```
redianthus: Is it actually necessary to redefine `bench_stats`? I don't think there is anything to "clone" here, right?

src/cmd/cmd_sym.ml
```diff
@@ -304,10 +303,17 @@ let handle_result ~exploration_strategy ~workers ~no_stop_at_failure ~no_value
...
+       @[<v>solver time: %a@;\
+       interpreter time: %fms@;\
+       execution time: %a@;\
+       path count: %i@]"
       Mtime.Span.pp
       (Atomic.get bench_stats.solver_time)
-      ((interpreter_time +. run_time) *. 1000.) );
+      ((interpreter_time +. run_time) *. 1000.)
+      Mtime.Span.pp execution_time
+      (Atomic.get bench_stats.path_count) );
```
redianthus: I'd like to reduce the size of `cmd_sym.ml` because it is already quite hard to follow. Could you move the printing function to `benchmark.ml`?

Something like:

```ocaml
let pp thread ~run_time ~interpreter_time ~execution_time m =
  let bench_stats = ... in
  let runtime = ... in
  m "Benchmarks:@\n..." ...
```

And then, in `cmd_sym` you'd do:

```ocaml
Log.bench (Benchmarks.pp thread ~run_time ~interpreter_time ~execution_time);
```

src/symbolic/symbolic_choice.ml
```diff
@@ -507,6 +509,8 @@ module Make (Thread : Thread_intf.S) = struct
           let* () = add_pc not_this_value_cond in
           generator ()
         in
+        let* thread in
+        Atomic.incr (Thread.bench_stats thread).path_count;
```
redianthus: Could you define a function `Thread.incr_path_count` instead?

src/symbolic/thread.ml
S41d: yeah that's right, i'll fix it

PR #831, opened 2025-11-20
Fix the fetching of solver stats

### LINKED ISSUE #184: add a flag to print the solver statistics

hra687261:
This PR attempts to fix the fetching of statistics that was added #797 by making sure that the statistics of all the used solver are fetched.
That is done by keeping in a global list every solver instance that was created and iterating over that list to get all their stats when `--bench` is enabled.
In some cases this causes a segmentation fault (when using multiple workers), I am not yet sure why but I suspect Z3 does not like being asked for statistics while it is still running (this can happen we found a bug and we try to fetch stats from a domain that is still running, and what is running in that domain in Z3),

Closes #184

### DIFF
--- a/src/symbolic/solver.ml
+++ b/src/symbolic/solver.ml
@@ -8,6 +8,11 @@ type t = S : ('a solver_module * 'a) -> t [@@unboxed]
 
 let instances = Atomic.make []
 
+let rec add_solver solver =
+  let l = Atomic.get instances in
+  let success = Atomic.compare_and_set instances l (solver :: l) in
+  if not success then add_solver solver
+
 let fresh solver_ty () =
   let module Mapping = (val Smtml.Solver_dispatcher.mappings_of_solver solver_ty)
   in
@@ -15,10 +20,7 @@ let fresh solver_ty () =
   let module Batch = Smtml.Solver.Cached (Mapping) in
   let solver_inst = Batch.create ~logic:QF_BVFP () in
   let solver = S ((module Batch), solver_inst) in
-  ( if Log.is_bench_enabled () then
-      let l = Atomic.get instances in
-      let l = solver :: l in
-      Atomic.set instances l );
+  if Log.is_bench_enabled () then add_solver solver;
   solver
 
 let check (S (solver_module, s)) pc =


hra687261:
After rebasing the CI does not pass since the version of smtml changed :/

src/symbolic/solver.ml
```diff
@@ -6,13 +6,20 @@ type 'a solver_module = (module Smtml.Solver_intf.S with type t = 'a)
...
   let module Mapping = Mapping.Fresh.Make () in
   let module Batch = Smtml.Solver.Cached (Mapping) in
-  let solver = Batch.create ~logic:QF_BVFP () in
-  S ((module Batch), solver)
+  let solver_inst = Batch.create ~logic:QF_BVFP () in
+  let solver = S ((module Batch), solver_inst) in
+  ( if Log.is_bench_enabled () then
+      let l = Atomic.get instances in
+      let l = solver :: l in
+      Atomic.set instances l );
```
redianthus: I think this is actually not right. You should either use `Atomic.compare_and_set` or a `Mutex.t`, otherwise, this could be racy and loose some solvers along the way

hra687261: Indeed, fixed!

PR #837, opened 2025-11-24
Bump smtml to 0.15.0

hra687261:
To be merged after https://github.com/ocaml/opam-repository/pull/28957 (and when the nix package is available :))

### DIFF
--- a/src/symbolic/symbolic_value.ml
+++ b/src/symbolic/symbolic_value.ml
@@ -213,28 +213,32 @@ module I32 = struct
     | _ -> Smtml.Expr.cvtop ty ToBool e
 
   let trunc_f32_s x =
-    try Ok (cvtop ty TruncSF32 x) with
-    | Smtml.Eval.Eval_error `Integer_overflow -> Error `Integer_overflow
-    | Smtml.Eval.Eval_error `Conversion_to_integer ->
-      Error `Invalid_conversion_to_integer
+    try Ok (cvtop ty TruncSF32 x)
+    with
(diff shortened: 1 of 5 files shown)


src/symbolic/symbolic_value.ml
```diff
@@ -214,23 +214,27 @@ module I32 = struct
 
   let trunc_f32_s x =
     try Ok (cvtop ty TruncSF32 x) with
-    | Smtml.Eval.Integer_overflow -> Error `Integer_overflow
-    | Smtml.Eval.Conversion_to_integer -> Error `Invalid_conversion_to_integer
+    | Smtml.Eval.Eval_error `Integer_overflow -> Error `Integer_overflow
+    | Smtml.Eval.Eval_error `Conversion_to_integer ->
+      Error `Invalid_conversion_to_integer
```
redianthus: I think we could rename our `Invalid_conversion_to_integer` to `Conversion_to_integer` and then simply do:

```ocaml
    | Smtml.Eval.Eval_error e -> Error e
```

WDYT?

hra687261: Agreed! I will do that!

hra687261: Hm the type checker does not like `Smtml.Eval.Eval_error e -> Error e`, saying that the error of type `Smtml.Eval.error_kind` is different from the one of type `Owi__.Result.err`. 
So it will be this instead: 
```ocaml
| Smtml.Eval.Eval_error ((`Integer_overflow | `Conversion_to_integer) as e)
    ->
      Error e
```

PR #851, opened 2025-12-01
Add support for the multi-memory proposal

hra687261:


doc/src/wasm-toolkit/overview.md
```diff
@@ -69,7 +69,7 @@ $ owi fmt horrible.wat
     i32.const 4
     local.get $n
     i32.mul
-    i32.load align=1
+    i32.load 0 align=1
```
redianthus: Maybe we could update the printing so that it makes a special case when the id is 0 and does not print it?

src/compile/rewrite.ml
```diff
@@ -210,20 +211,49 @@ let rewrite_expr (assigned : Assigned.t) (locals : Text.param list)
...
-    | F_load (nn, memarg) -> Ok (Binary.F_load (nn, memarg))
-    | F_store (nn, memarg) -> Ok (Binary.F_store (nn, memarg))
-    | I_store (nn, memarg) -> Ok (Binary.I_store (nn, memarg))
-    | Memory_copy -> Ok Binary.Memory_copy
-    | Memory_size -> Ok Binary.Memory_size
-    | Memory_fill -> Ok Binary.Memory_fill
-    | Memory_grow -> Ok Binary.Memory_grow
+    | I_load8 (id, nn, sx, memarg) ->
+      let* id = Assigned.find_memory assigned id in
+      Ok (Binary.I_load8 (id, nn, sx, memarg))
```
redianthus: I usually write:

```ocaml
      let+ id = Assigned.find_memory assigned id in
      Binary.I_load8 (id, nn, sx, memarg)
```

instead to get less parenthesis, but it's not very important.

src/interpret/interpret.ml
```diff
@@ -998,35 +995,39 @@ struct
...
-          (I64.gt_u I64.(add dst len) size)
+          (I64.gt_u I64.(add src len) size1)
+          (I64.gt_u I64.(add dst len) size2)
       in
       if out_of_bounds then Choice.trap `Out_of_bounds_memory_access
       else begin
-        let* mem = Env.get_memory env mem_0 in
-        let* () = Memory.blit mem ~src ~dst ~len in
+        (* Is this necessary? *)
+        let* mem1 = Env.get_memory env memid1 in
```
redianthus: It is! Otherwise you'd read the memory coming from the "thread" before the `let>` and it may not be the one you should use, as `let>` is forking the execution. This is a bug we've had many times and for which I have not implemented a fool-proof solution yet...

src/interpret/interpret.ml
```diff
@@ -516,7 +512,7 @@ struct
     | Num_type V128 -> V128 V128.zero
     | Ref_type (_null, rt) -> ref_null rt
 
-  (* TODO move to module Env *)
+  (* TODO: remove *)
```
redianthus: Is it still used somewhere?

src/intf/interpret_intf.ml
```diff
@@ -2,132 +2,137 @@
...
+    val grow : t -> Value.int32 -> unit
 
-  val size : t -> int
+    val fill : t -> pos:Value.int32 -> len:Value.int32 -> char -> unit Choice.t
 
-  val typ : t -> Text.ref_type
+    val blit :
+         t
+      -> src:Value.int32
+      -> t
```
redianthus: Maybe we should name them now that we've got two of them?

src/ir/binary_encoder.ml
```diff
@@ -141,8 +141,11 @@ let write_limits buf (limits : Text.limits) =
     write_u32_of_int buf min;
     write_u32_of_int buf max
 
-let write_memarg buf ({ offset; align } : Text.memarg) =
-  write_u32 buf align;
+let write_memarg buf idx ({ offset; align } : Text.memarg) =
+  (* Set the 6th bit *)
+  write_u32 buf (Int32.logor align 0x40l);
+  (* Should this set the max_align bit instead of always setting the 6th? *)
```
redianthus: I know near to nothing to the binary format so I can not really help here. We can discuss it in person.

src/link/link.ml
redianthus: I am not sure this should still be there.

src/parser/text_parser.mly
```diff
@@ -450,34 +458,39 @@ let plain_instr :=
...
+  | I32_STORE8; id = memidx; memarg = memarg; { I_store8 (id, S32, memarg) }
+  | I64_STORE8; id = memidx; memarg = memarg; { I_store8 (id, S64, memarg) }
+  | I32_STORE16; id = memidx; memarg = memarg; { I_store16 (id, S32, memarg) }
+  | I64_STORE16; id = memidx; memarg = memarg; { I_store16 (id, S64, memarg) }
+  | I64_STORE32; id = memidx; memarg = memarg; { I64_store32 (id, memarg) }
+  | MEMORY_SIZE; id = memidx; { Memory_size id }
+  | MEMORY_GROW; id = memidx; { Memory_grow id }
+  | MEMORY_FILL; id = memidx; { Memory_fill id }
+  | MEMORY_COPY; id1 = memidx; id2 = memidx; { Memory_copy (id1, id2) }
+  | MEMORY_INIT; args = list(indice); {
```
redianthus, line 491: You can use `nonempty_list(indice)` I guess?

redianthus: Hm, maybe not, it would make the error message worst...

src/compile/rewrite.ml
hra687261: Agreed! Done!

src/intf/interpret_intf.ml
hra687261: Agreed! What do you suggest?
Something like:
- `srcmem:t -> src:int32 -> dstmem:t -> dst:int32 -> len:int32 -> unit Result.t`
- `src:t -> srcaddr:int32 -> dst:t -> dstaddr:int32 -> len:int32 -> unit Result.t`
... ?

src/interpret/interpret.ml
hra687261: Yes, I still haven't updated `apply` in `exec_extern_func`, I'll need to add the memory id to `Mem`

src/intf/interpret_intf.ml
redianthus: What about `src:t -> src_idx:int32 -> dst:t -> dst_idx:int32 -> len:int32` ?

EDIT: `addr` is fine too!

src/link/link.ml
```diff
@@ -258,15 +258,13 @@ let active_elem_expr ~offset ~length ~table ~elem =
   ]
 
 let active_data_expr ~offset ~length ~mem ~data =
-  if mem <> 0 then Error (`Unknown_memory (Text.Raw mem))
```
redianthus, line 261: But, shouldn't we still check something? I guess we'll see once the tests from the reference test suites are checked

src/intf/interpret_intf.ml
hra687261: > What about `src:t -> src_idx:int32 -> dst:t -> dst_idx:int32 -> len:int32` ?

Done!

src/link/link.ml
hra687261: I don't know tbh, I am not sure I understand this code, maybe something like `IMap.mem id env.memories`? But the Ids that are added to `env.memories` are seem to be arbitrary, so I don't know. I'll have to take a closer look at this code (It's also possible that this PR is missing something and that the Ids should be chosen some other way)

redianthus: Yes, I wasn't familiar with this code either until two weeks ago :sweat_smile:

I can explain it to you soon (along with stuff related to active/passive data segments).

doc/src/wasm-toolkit/overview.md
hra687261: Done! I am thinking of doing this also with the binary parsing/encoding, but it will need to be done once its fixed.

src/ir/text.ml
```diff
@@ -20,6 +20,11 @@ let pp_id_opt fmt = function None -> () | Some i -> pf fmt " %a" pp_id i
 
 let pp_indice fmt = function Raw u -> int fmt u | Text i -> pp_id fmt i
 
+let pp_indice_not0 fmt = function
+  | Text i -> pp_id fmt i
+  | Raw u when u = 0 -> ()
```
redianthus:
```suggestion
  | Raw 0 -> ()
```

src/interpret/interpret.ml
hra687261: That actually might be more complicated than expected, I opened a separate PR for discussion in https://github.com/hra687261/owi/pull/1

src/parser/text_parser.mly
```diff
@@ -450,34 +458,39 @@ let plain_instr :=
...
+  | I64_LOAD32_U; id = memidx; memarg = memarg; { I64_load32 (id, U, memarg) }
+  | I32_STORE8; id = memidx; memarg = memarg; { I_store8 (id, S32, memarg) }
+  | I64_STORE8; id = memidx; memarg = memarg; { I_store8 (id, S64, memarg) }
+  | I32_STORE16; id = memidx; memarg = memarg; { I_store16 (id, S32, memarg) }
+  | I64_STORE16; id = memidx; memarg = memarg; { I_store16 (id, S64, memarg) }
+  | I64_STORE32; id = memidx; memarg = memarg; { I64_store32 (id, memarg) }
+  | MEMORY_SIZE; id = memidx; { Memory_size id }
+  | MEMORY_GROW; id = memidx; { Memory_grow id }
+  | MEMORY_FILL; id = memidx; { Memory_fill id }
+  | MEMORY_COPY; id1 = memidx; id2 = memidx; { Memory_copy (id1, id2) }
```
redianthus: It seems there is a conflict now, and I suspect that memory.copy may be the culprit! I can take care of it but I'm writing this so I don't forget about it

redianthus: Looking at the reference interpreter, it seems that you should either put none of them, either have the two of them, but it is not possible for both to be independently optional.

See here: https://github.com/WebAssembly/spec/blob/4d81d3b085fe4d1e7ad1b9852a8288bd1caad505/interpreter/text/parser.mly#L610C17-L610C28
And there: https://github.com/WebAssembly/spec/blob/4d81d3b085fe4d1e7ad1b9852a8288bd1caad505/interpreter/text/parser.mly#L500

It is actually the same for `table.copy` which we handle like this in Owi:

```mly
  | TABLE_COPY; {
    Table_copy (Raw 0, Raw 0)
  }
  | TABLE_COPY; src = indice; dst = indice; { Table_copy (src, dst) }

```

hra687261: Indeed, it should be fixed now.

test/script/reference.t
```diff
@@ -176,7 +176,8 @@
   $ owi script --no-exhaustion reference/proposals/multi-memory/memory_trap1.wast
   $ owi script --no-exhaustion reference/proposals/multi-memory/memory-multi.wast
   $ owi script --no-exhaustion reference/proposals/multi-memory/memory.wast
-  $ owi script --no-exhaustion reference/proposals/multi-memory/simd_memory-multi.wast
+;  $ owi script --no-exhaustion reference/proposals/multi-memory/simd_memory-multi.wast
+; simd instructions are not yet fully supported
```
redianthus:
```suggestion
; TODO: simd instructions are not yet fully supported
```

I usually grep "TODO", so I'm simply adding it to be able to find it so it does not go forgotten for ever

src/interpret/interpret.ml
```diff
@@ -556,8 +549,8 @@ struct
       Stack.t -> (f, r) Extern_func.atype -> f -> r Choice.t =
      fun stack ty f ->
       match ty with
-      | Mem args ->
-        let* mem = Env.get_memory env mem_0 in
+      | Mem (memid, args) ->
+        let* mem = Env.get_memory env (Int32.to_int memid) in
```
redianthus: I think we can safely change this to be an `int` to avoid the conversion. And it'll be closer to what is done at the instruction level where in `Memory_size memid` for instance, `memid` is an `int`.

test/script/reference.t
```diff
@@ -54,6 +54,8 @@
   24
   24
   13
+  owi: [ERROR] expected multiple memories but got (unknown module )
+  [54]
```
redianthus: This is a little bit annoying because it prevents the remaining tests to run. I believe we should either:

1. update the submodule to a latest version where it may have been fixed
2. stop running this tests if the assumption is that an engine supporting the multi-memory proposal should only run the multi-memory tests and not the old ones (does the multi-memory proposal directory includes a copy of the old ones without the tests failing on multiple memories?)

Have you tried 1. ?

src/interpret/interpret.ml
hra687261: I agree, done!

test/script/reference.t
hra687261: I agree, the issue with updating the submodule is that its structure significantly changed, so simply updating the repository leads to many tests (all the multi-memory ones for example as well others) becoming "missing" files.
But I will take a look, maybe I can find an intermediary commit that fixes the issues and for which the structure of the files has not completely changed

hra687261: But there are indeed copies of the tests under the `multiple memories` folder were the only differences is that the tests that fail when there are multiple memories have been removed, so it seems that solution 2 is the more straightforward one.
I will do a quick cleanup of duplicate tests for now, we can open a separate issue for the updating of the repository since it is more work.

hra687261: Here's PR with 1.: https://github.com/hra687261/owi/pull/3
So I managed to update the paths of the files, it seems they were just move out of the `/proposals/x/` repo so it was easy, but there are multiple new errors that need investigating.
I think we can do that in a separate PR, what do you think @redianthus ?

redianthus: Yes, it will be better to handle this separately.

> But there are indeed copies of the tests under the multiple memories folder were the only differences is that the tests that fail when there are multiple memories have been removed, so it seems that solution 2 is the more straightforward one.

I think this is the best solution. :)

redianthus:
The failing test is actually unrelated to multi-memory but was introduced with the proposal for some reasons. Here is a patch: https://github.com/hra687261/owi/pull/4

redianthus:
Thanks!

PR #858, opened 2025-12-07
Update fuzzer

hra687261:


src/owi.ml
```diff
@@ -23,6 +23,10 @@ module Cmd_zig = Cmd_zig
 module Compile = Compile
 module Concrete_extern_func = Concrete_extern_func
 module Concrete_memory = Concrete_memory
+module Int32 = Int32
+module Float32 = Float32
+module Float64 = Float64
```
redianthus: I don't think we want to expose these modules, they should be replaced by `Concrete_{i32,i64,f32,f64}`.

test/fuzz/basic.ml
```diff
@@ -411,92 +432,113 @@ let memarg nsize =
...
 let data_mode (env : Env.t) =
-  match env.memory with
-  | Some name -> choose [ const Owi.Text.Data_passive; data_active name ]
-  | None -> const Owi.Text.Data_passive
+  match env.memories with
+  | [] -> const Owi.Text.Data.Mode.Passive
+  | memories ->
+    choose
+      ( const Owi.Text.Data.Mode.Passive
+      :: List.map (fun id -> data_active id) memories )
```
redianthus:
```ocaml
let data_mode (env : Env.t) =
  choose (const Owi.Text.Data.mode.Passive :: List.map data_active env.memories)
```

is equivalent (no need to special case the empty list

test/fuzz/gen.ml
```diff
@@ -312,16 +314,10 @@ let func env : Owi.Text.module_field gen =
...
+  Module.Field.Func { type_f; locals; body; id }
 
-let fields env : Owi.Text.module_field list gen =
-  let+ memory =
-    (* No memory management in symbolic context.
-       TODO: When implementation will be more advanced,
-       reactivate and refine instruction by instruction (not_symbolic operator). *)
-    match env.Env.conf with
-    | Concrete -> option (memory env)
-    | Symbolic -> const None
```
redianthus, line 324: IIRC I removed the symbolic mode from the fuzzer as it was not that much used (mostly because it was incomplete). If this is actually the case, maybe we could also remove this `env.conf` thing entirely?

src/owi.ml
hra687261: Done

test/fuzz/basic.ml
hra687261: Indeed, fixed!

test/fuzz/gen.ml
hra687261: I am actually unsure, I don't see a significant difference in the behavior with this change and without it (I get a few `runtime trap: unreachable executed` with both).
But I can revert it and leave it for a better investigation in the future.

test/fuzz/interprets.ml
```diff
@@ -45,7 +45,7 @@ end
 
 module Reference : INTERPRET = struct
   let parse_and_run modul =
-    let* tmp_file = Bos.OS.Dir.tmp "owi_fuzzer_official%s.wast" in
+    let* tmp_file = Bos.OS.File.tmp "owi_fuzzer_official%s.wast" in
```
redianthus: I also think it should be `.wat` and not `.wast`

hra687261: Indeed, done!

redianthus:
Yes I think that was the only thing. Should be fixed now. :)

PR #859, opened 2025-12-07
Update the testsuite subrepo and add missing instructions/types/checks needed for the tests to pass

hra687261:
The purpose of this PR is to update the test suite and track/fix the various errors that happen after this update.
This was done while I was working on #851

### DIFF
--- a/src/parser/text_parser.mly
+++ b/src/parser/text_parser.mly
@@ -1212,7 +1212,7 @@ let assert_ ==
   | ASSERT_UNLINKABLE; ~ = modul; ~ = NAME; <Assert_unlinkable>
 
 let instance ==
-  | MODULE; INSTANCE; ~ = option(id); ~ = id; <Instance>
+  | MODULE; INSTANCE; ~ = ioption(id); ~ = id; <Instance>
 
 let register ==
   | REGISTER; ~ = utf8_name; ~ = option(id); <Register>


src/compile/rewrite.ml
```diff
@@ -89,20 +129,40 @@ let rewrite_expr (assigned : Assigned.t) (locals : Text.param list)
...
-      let id = find_local id in
-      Ok (Binary.Local_get id)
+      let id = get_local_id id in
+      let* opt = find_local id in
+      begin match opt with
+      | Some (ty, init) ->
+        let has_def_value =
+          match ty with Text.Ref_type (No_null, _) -> false | _ -> true
+        in
+        if (not init) && not has_def_value then Error (`Uninitialized_local id)
```
redianthus: Maybe do this in validate_binary instead so that it is checked for binary modules too

src/compile/rewrite.ml
```diff
@@ -136,8 +196,11 @@ let rewrite_expr (assigned : Assigned.t) (locals : Text.param list)
       let+ idx = Assigned.find_global assigned id in
       Binary.Global_set idx
     | Global_get id ->
-      let+ idx = Assigned.find_global assigned id in
-      Binary.Global_get idx
+      let* idx = Assigned.find_global assigned id in
+      begin match known_globals with
+      | Some n when idx >= n -> Error (`Unknown_global (Text.Raw idx))
+      | _ -> Ok (Binary.Global_get idx)
+      end
```
redianthus: it should be checked by find_global or in typechecking phase already

src/validate/binary_validate.ml
```diff
@@ -290,244 +341,278 @@ let rec typecheck_instr (env : Env.t) (stack : stack) (instr : instr Annotated.t
...
+    let* stack = Stack.pop env.modul [ i32; Ref_type t ] stack in
     Stack.push [ i32 ] stack
   | Table_size i ->
     let* _null, _t = Env.table_type_get i env.modul in
     Stack.push [ i32 ] stack
   | Ref_is_null ->
     let* stack = Stack.pop_ref stack in
     Stack.push [ i32 ] stack
+  | Ref_as_non_null ->
+    let* stack = Stack.pop_ref stack in
```
redianthus: let* rt, stack = Stack.pop_ref stack in
Stack.push rt stack

src/compile/rewrite.ml
hra687261: Done,

src/compile/rewrite.ml
hra687261: Done. (in binary_validate)

src/validate/binary_validate.ml
hra687261: Done.

src/cmd/cmd_cfg.ml
```diff
@@ -103,6 +103,17 @@ let rec build_graph (l : Binary.expr) nodes n node edges
           (edges_to_add, 0) inds
       in
       (nodes, edges, n + 1, edges_to_add, false)
+      (* TODO: is this correct? *)
```
redianthus, line 106: I think this is fine for now. We can choose later how to encore branching on references, either say `null ~ 0` and `nonnull ~ != 0` or change the type to be a variant and have a proper type for references.

src/compile/assigned.ml
```diff
@@ -40,10 +42,12 @@ let pp fmt
      Mem names: %a@\n\
      Func names: %a@\n\
      Elem names: %a@\n\
-     Data names: %a@\n"
+     Data names: %a@\n\
+     Tag names: %a@\n"
```
redianthus, line 46: Can you explain to me which proposal is introducing tags? Is it related to exceptions?

src/compile/binary_to_text.ml
```diff
@@ -194,29 +223,40 @@ let convert_data : Binary.Data.t -> Text.Data.t = function
...
+    types
   |> Array.to_list
 
-let from_global (global : (Binary.Global.t, Text.Global.Type.t) Origin.t array)
+let from_global (global : (Binary.Global.t, Binary.Global.Type.t) Origin.t array)
   : Text.Module.Field.t list =
   Array.map
     (function
-      | Origin.Local (g : Binary.Global.t) ->
+      | Origin.Local ({ typ = mut, vt; _ } as g : Binary.Global.t) ->
```
redianthus: Maybe add `id` here to avoid the `let id = g.id in` after?

src/compile/grouped.ml
```diff
@@ -16,7 +16,9 @@ type t =
   { id : string option
   ; typ : Typedef.t Array.t
   ; function_type : func_type Array.t
-      (** Types comming from function declarations. It contains potential
+      (* TODO: "function_type" should be renamed to something like
+          "implicit_types"?  *)
```
redianthus: Please, go ahead and rename it with anything that is better. :)

src/compile/rewrite.ml
```diff
@@ -320,38 +389,76 @@ let rewrite_exports (modul : Grouped.t) (assigned : Assigned.t) :
...
+         let+ val_type = rewrite_val_type assigned val_type in
+         (m, val_type) ) )
       modul.global
   in
+  let* table =
+    array_map
+      (Origin.monadic_map ~f_local:(rewrite_table assigned)
+         ~f_imported:(fun ((l, (n, ht)) : Text.Table.Type.t) ->
+         let+ ht = rewrite_heap_type assigned ht in
+         (l, (n, ht)) ) )
```
redianthus: Maybe make `f_imported` an auxiliary function as it is done in later cases?

src/compile/rewrite.ml
```diff
@@ -320,38 +389,76 @@ let rewrite_exports (modul : Grouped.t) (assigned : Assigned.t) :
...
 let modul (modul : Grouped.t) (assigned : Assigned.t) : Binary.Module.t Result.t
     =
   Log.debug (fun m -> m "rewriting    ...");
   let* global =
     array_map
       (Origin.monadic_map ~f_local:(rewrite_global assigned)
-         ~f_imported:Result.ok )
+         ~f_imported:(fun (m, val_type) ->
+         let+ val_type = rewrite_val_type assigned val_type in
+         (m, val_type) ) )
```
redianthus: (same as next comment)

src/concrete/concrete_memory.ml
```diff
@@ -10,12 +10,15 @@ type t =
   }
 
 let init limits : t =
-  let data = Bytes.make (page_size * limits.Text.min) '\x00' in
+  let data = Bytes.make (page_size * Int64.to_int limits.Text.min) '\x00' in
```
redianthus: It works for now because we don't support `memory64` but this is going to become wrong in the future if we do, right? If this is a case, maybe adding a `TODO` comment would be nice.

src/concrete/concrete_ref.ml
```diff
@@ -17,19 +17,30 @@ end
...
+  | Extern None -> pf fmt "externref none"
   | Extern _ -> pf fmt "externref"
   | Func _ -> pf fmt "funcref"
+  | NullExn -> pf fmt "nullexnref"
+  | NullRef -> pf fmt "nullref"
 
-let null = function Text.Func_ht -> Func None | Extern_ht -> Extern None
+(* TODO: Is this the same as Symbolic_ref.null? *)
+let null = function
+  | Binary.Func_ht | NoFunc_ht | TypeUse _ -> Func None
```
redianthus, line 32: What is this `TypeUse` here?

src/concrete/concrete_table.ml
```diff
@@ -19,10 +19,10 @@ let fresh =
     incr r;
     !r
 
-let init ?label (typ : Text.Table.Type.t) : t =
+let init ?label (typ : Binary.Table.Type.t) : t =
   let limits, ((_null, heap_type) as ref_type) = typ in
   let null = Concrete_ref.null heap_type in
-  let table = Array.make limits.min null in
+  let table = Array.make (Int64.to_int limits.min) null in
```
redianthus: Does `memory64` also makes tables possibly 64bits indexed? If yes, then this is fine. If not, we should change the type (I guess it implies to have two different limits types now).

src/interpret/interpret.ml
```diff
@@ -1331,7 +1391,7 @@ struct
       let pos, stack = Stack.pop_i32 stack in
       let addr =
         let pos = I64.extend_i32_u pos in
-        let offset = I32.of_concrete offset |> I64.extend_i32_u in
+        let offset = I64.of_concrete offset in
         I64.add pos offset
```
redianthus: I believe this addition is now wrong because it can overflow (the reasong we were using an `I64` was to avoid the `I32` overflow).

src/ir/text.mli
```diff
@@ -139,16 +143,23 @@ type nonrec frelop =
...
 val pp_memarg : memarg Fmt.t
 
+(* TODO: limits should probably be in an sum type representing the
+   cases where they are i64 and i32 (maybe even when max is
+   present or not). *)
 type nonrec limits =
-  { min : int
-  ; max : int option
+  { min : Concrete_i64.t
+  ; max : Concrete_i64.t option
```
redianthus: Actually, this is a number of pages, so, it needs at most $48$ bits to be stored. Thus, I think using `int` was fine but is also still fine with memory64.

src/script/script.ml
```diff
@@ -288,7 +335,16 @@ let run ~no_exhaustion script =
...
         let+ _stack = action link_state a in
-        link_state )
+        link_state
+      | Text_module (true, _)
+      | Binary_module (true, _, _)
+      | Quoted_module (true, _) ->
+        (* TODO: differentiate between modules and module definitions in the
+           link state, ensure that we can instantiate a module from its module
+           definition, and that module definitions are not treated as "normal",
+           or instantiated module. *)
```
redianthus, line 345: The link state should only be made of module instances, definitions do not belong in there. What about adding some notion of state to the script interpreter which would remember both the previous definitions plus the link state?

src/validate/binary_validate.ml
```diff
@@ -707,34 +972,92 @@ let validate_exports modul =
...
+                     (Fmt.str "Table type is %a but init expr is of type %a"
+                        pp_ref_type rt pp_typ_list res_tys ) )
+            | _ ->
+              Error
+                (`Type_mismatch
+                   (Fmt.str "Table type is %a but init expr is of type %a"
+                      pp_ref_type rt pp_typ_list res_tys ) )
+            end
+        in
+        check_limit ~table:true limits
```
redianthus: Do you mind making this anonymous function smaller? I think defining a few auxiliary functions could help a lot.

src/validate/text_validate.ml
```diff
@@ -47,7 +117,10 @@ let modul m =
       if Hashtbl.mem seen id then Error (`Duplicate_memory id)
       else Ok (Hashtbl.add seen id ())
   in
-
+  (* let m = Text.Module.{ m with fields = List.sort Field.compare m.fields } in *)
+  (* in some tests, modules are exptected to be evaluated in an order different
+    from the one in which they were written. e.g. the last "unknown global" in
+    "test/references/global.wast". *)
```
redianthus: I guess fields of the same kind should not be reordered, but fields of different kinds should? The binary format forces to first declare all types, then all function (I don't remember the exact order, this is just for illustration purpose). I guess we should use the same order for text validation? This is somehow suggesting we should do the text validation on the `Grouped.modul` type instead...

src/compile/assigned.ml
hra687261: Yes, it is related to exceptions: 
- https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions.md#exception-handling-1
- https://webassembly.github.io/spec/core/text/modules.html#tags

src/compile/binary_to_text.ml
hra687261: Done.

src/compile/grouped.ml
hra687261: Done. Renamed to `decl_types` (i.e. types from function or tag declarations)

src/compile/rewrite.ml
hra687261: Done.

src/compile/rewrite.ml
hra687261: Done.

src/concrete/concrete_memory.ml
hra687261: I am not sure, can you explain why it would become wrong?

src/concrete/concrete_ref.ml
hra687261: `TypeUse` is a kind of `heap_type`: https://webassembly.github.io/spec/core/syntax/types.html#heap-types
Instructions like `ref.null` take a `heap_type`: https://webassembly.github.io/spec/core/syntax/instructions.html#reference-instructions
So you can write stuff like `ref.null $f` where `$f` is a function identifier, as in: https://github.com/WebAssembly/testsuite/blob/345367358f065375524498749470720d9cdd1418/br_if.wast#L667-L679
(Edit: maybe if all nulls are equal then `TypeUse $f` can be replaced by `Func_ht` when `$f` is a function and with something otherwise? That would remove the need for `Binary.heap_type`)

src/concrete/concrete_table.ml
hra687261: Yes, both tables and memories can be 64bits indexed: https://webassembly.github.io/spec/core/syntax/types.html#memory-types but they also both can take an `addrtype` argument https://webassembly.github.io/spec/core/syntax/types.html#syntax-addrtype which states whether they are indexed by 32bits or 64bits, so maybe we have to have different types of limits anyway (or at least keep the information somewhere on whether the Int64 are holding u64 or u32)

src/interpret/interpret.ml
hra687261: Indeed, I was starting to wonder if using `Z.t`s is not the right thing to do in this case, maybe not everywhere as it would be annoying to track what is a i32 and what is a i64, but at least for such operations where we need to do overflow checks, we can convert to `Z.t` then back. What do you think?

src/validate/binary_validate.ml
hra687261: Done.

src/validate/text_validate.ml
hra687261: Sorry I forgot to remove this comment, it is outdated, pretty sure I had misunderstood some error I got, that's why I removed the sorting line.

src/ir/text.mli
hra687261: You're right, it's fixed.

src/concrete/concrete_memory.ml
redianthus: Because of the int64->int conversion.

But actually, there is no reason to use an `i64` for `limits` I think. They are number of pages, so they fit in 48 bits, using `int` is fine.

src/interpret/interpret.ml
redianthus: Using `Z.t` is too annoying because here `I64` can be a concrete or a symbolic value depending on how the functor was instantiated, and in the future it can be something else (concolic etc.).

I think, we'll have to rewrite all bounds checks and be careful about overflow, this can be done while staying in the `i64` world, but it can be subtle. Defining auxiliary functions will help. Let's discuss this tomorrow.

src/concrete/concrete_memory.ml
hra687261: Indeed, fixed in: https://github.com/OCamlPro/owi/pull/859/commits/bb3cfdbe55e04235947145d31a4bbbe8ea087974

src/interpret/interpret.ml
hra687261: Something like: https://github.com/hra687261/owi/pull/5 ?

redianthus: The refactoring is quite nice.

I'm not sure I understand your overflow thing (but I read really quickly). What I was thinking was along the lines of:


```ocaml
if len >u mem_size then trap (* protect next check against underverflow *)
else if addr >u mem_size - len then trap (* same as `addr+len >u mem_size` but can not overflow *)
else OK
```

(on `I64` and using only unsigned stuff)

hra687261: My check assumes that we are working with unsigned integers, so if a + b overflows the results is necessarily <= a and <= b, because a <= max, b <= max, a + b - max = delta, a - max = delta - b, given that a <= max, then a - max <= 0 and delta - b <= 0,  therefore delta <= b (similarly for a).

The idea is simply to take the original check and add the overflow checks for the int64 additions, but I am not sure if they all can overflow. There are two which are checked `let addr = I64.add pos offset` and `let res = I64.(add addr (of_concrete const))`.

I think yours is equivalent, except the missing check for computing add `addr`. (`len` is `const` and `mem_size` is `size` in the PR by I can rename them)

~~EDIT: I think it's the other way around `mem_size - len <u addr` no? i.e. it fails when ` mem_size <u addr+len`~~

redianthus: (I think I edited my answer while you were still reading, maybe reload the page and see my latest edit :sweat_smile:)

redianthus: Final algorithm:

1. assert that `offset <=u MAXINT - const`
2. `len := offset + const`
3. `if len >=u mem_size then trap`
4. assert that `pos <=u mem_size - len`

hra687261: This shoudl be good: https://github.com/hra687261/owi/blob/8b0f02c39b851334c56cd1ef52a836c223e7f001/src/interpret/interpret.ml#L895-L912?

PR #861, opened 2025-12-08
improve symbolic driver, and make sure solver are interrupted and domain

redianthus:
are done

### DIFF
--- a/src/symbolic/solver.ml
+++ b/src/symbolic/solver.ml
@@ -27,27 +27,35 @@ let check (S (solver_module, s)) pc =
   let module Solver = (val solver_module) in
   Solver.check_set s pc
 
-let model_of_partition (S (solver_module, s)) ~partition : Smtml.Model.t =
+let model_of_path_condition (S (solver_module, s)) ~path_condition :
+  Smtml.Model.t Option.t =
+  let exception Unknown in
   let module Solver = (val solver_module) in
-  let partition =
-    List.map
-      (fun pc ->
-        match Solver.get_sat_model s pc with
-        | `Model model -> model
-        | `Unknown -> assert false
-        | `Unsat -> assert false )
-      partition
-  in
-  let model = Hashtbl.create 64 in
-  List.iter
-    (fun tbl -> Hashtbl.iter (fun sym v -> Hashtbl.add model sym v) tbl)
-    partition;
-  model
+  try
+    let sub_conditions = Symbolic_path_condition.slice path_condition in
+    let models =
+      List.map
+        (fun pc ->
+          match Solver.get_sat_model s pc with
+          | `Model model -> model
+          | `Unknown ->
+            (* it can happen if the solver is interrupted, otherwise it is an error, we raise, so the function can return an option that will be handled by the called *)
+            raise Unknown
+          | `Unsat ->
+            (* it can not happen otherwise it means we reached an unreachable branch (or added garbage to the PC and did something wrong, who knows... :-) *)
+            assert false )
+        sub_conditions
+    in
+    (* We build the new complete model by merging all "sub models" *)
+    let model = Hashtbl.create 64 in
+    List.iter (Hashtbl.iter (Hashtbl.add model)) models;
+    Some model
+  with Unknown -> None
 
-let get_sat_model (S (solver_module, s)) ~symbol_scopes ~pc =
+let model_of_set (S (solver_module, s)) ~symbol_scopes ~set =
   let module Solver = (val solver_module) in
   let symbols = Symbol_scope.only_symbols symbol_scopes in
-  Solver.get_sat_model ~symbols s pc
+  Solver.get_sat_model ~symbols s set
 
 let empty_stats = Smtml.Statistics.Map.empty
 
@@ -61,9 +69,17 @@ let interrupt_all () =
       Solver.interrupt s )
     solvers
 
-let get_all_stats () =
+let get_all_stats ~wait_for_all_domains =
   if not (Log.is_bench_enabled ()) then empty_stats
   else begin
+    (* interrupt_all is unreliable but is a best effort to try to make sure we don't wait too long on really long requests.
+     The reliable alternative would be to backup the statistics before each SMT request when in benchmark mode, but this would be too costly and lead to less accurate requests than random failures... *)
+    interrupt_all ();
+    (* we wait for all domains to terminate because:
+    - some solvers can not be interrupted and usually don't like being asked for statistics while running
+    - we already got all the previous metrics, so we don't care about waiting more... *)
+    wait_for_all_domains ();
+
     let solvers = Atomic.get instances in
     List.fold_left
       (fun stats_acc (S (solver_module, s)) ->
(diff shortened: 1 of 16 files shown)


src/symbolic/solver.ml
```diff
@@ -50,16 +50,30 @@ let empty_stats = Smtml.Statistics.Map.empty
...
 let stats_are_empty = Smtml.Statistics.Map.is_empty
 
-let get_all_stats ~finalizer () =
+let interrupt_all () =
+  let solvers = Atomic.get instances in
+  List.iter
+    (fun (S (solver_module, s)) ->
+      let module Solver = (val solver_module) in
+      Solver.interrupt s )
+    solvers
```
hra687261, line 70: I am unsure of the reliability of `Solver.interrupt`, it's also worth noting that it only exists for Z3 (in smtml but also in the solvers I think), though statistics are also only supported for Z3.
However, we should be able to use the `--bench` with other solvers without issue even if we don't get stats, but with `Solver.interrupt`  we will get an assert failure/error with the other solvers.

src/symbolic/solver.ml
```diff
@@ -50,16 +50,30 @@ let empty_stats = Smtml.Statistics.Map.empty
...
+
+let get_all_stats () =
   if not (Log.is_bench_enabled ()) then empty_stats
   else begin
-    finalizer ();
-    (* ensure that everyone has finished working *)
```
hra687261, line 57: Since `interrupt_all` is only used before we call `get_all_stats`, shouldn't the call be here so that we don't forget to call it or to wait on the domains of the solvers to stop working?

src/symbolic/symbolic_choice.ml
```diff
@@ -456,7 +460,9 @@ module Make (Thread : Thread_intf.S) = struct
           | `Unsat ->
             Atomic.set is_other_branch_unsat true;
             stop
-          | `Unknown -> assert false
+          | `Unknown ->
+            (* It can happen when the solver is interrupted *)
+            stop
```
hra687261, line 474: We should probably raise some error when it happens if the solver was not interrupted. 
Since we only use the quantifier free BV and FP theories, if a solver answers unknown (when it is not interrupted) it means it's either a bug or the decision procedure is incomplete.

src/symbolic/solver.ml
redianthus: `interrupt` has been working well according to my testing, in the sense that it correctly stops the solver and ease the end of execution. However, the state of the solver after the interrupt can be considered as "garbage".

Meaning, call `get_statistics` on it may fail with really weird errors: `Z3.Error "canceled"` when lucky, any Z3 error on more rare cases.

What I did is to simply ignore any Z3 exceptions and use empty statistics when this is failing (+ displaying a warning about the failure).

> get an assert failure/error with the other solvers.

Isn't `interrupt` a nop for other solvers?

src/symbolic/symbolic_choice.ml
redianthus: Would a simple `Log.warn` be enough?

redianthus: Or maybe, if we had a way to know if the solver was interrupted, it could be possible to do something nicer. @filipeom, what do you think of adding a `was_interrupted : bool ref` to each solver state? And set it to `true` in the `interrupt` function (before actually interrupting the solver).

src/symbolic/solver.ml
hra687261: Alright, that looks good to me.

> Isn't interrupt a nop for other solvers?

It will be once https://github.com/formalsec/smtml/pull/478 is merged :)

src/symbolic/symbolic_choice.ml
filipeom: Yeah, I think that works. The only issue is when you set an internal solver timeout and Z3 hits that timeout. The solver will return unknown and `was_interrupted` will remain false (because `interrupt` was not called directly). Also, I think Z3's `interrupt` is more like a "best effort" to stop the solver not sure if it stops it immediately.

hra687261: I would definitely prefer the second option. But, now that I think of it, if the execution stops anyway then maybe (does `stop` stop everything?), then maybe  a simple message is sufficient.
(And maybe https://github.com/formalsec/smtml/pull/479 is not necessary after all)

src/symbolic/solver.ml
redianthus: I improved the situation with comments, additional logs and stuff.

src/symbolic/symbolic_choice.ml
redianthus: `stop` is only stopping the current exploration. I left `stop` for now and added a lot of TODO+comments to say we should check for was_interrupted in the future and assert false if it was not.

src/symbolic/solver.ml
redianthus: I changed this a bit and moved as much as possible to `solver.ml`.

src/symbolic/symbolic_choice.ml
```diff
@@ -390,41 +401,41 @@ module Make (Thread : Thread_intf.S) = struct
...
+      Solver.check solver pc
     in
     return reachability
 
   let get_model_or_stop symbol =
     let* () = yield Prio.default in
     let* solver in
-    let+ thread in
-    let pc = Thread.pc thread |> Symbolic_path_condition.to_set in
+    let* thread in
```
redianthus, line 415: @krtab, I'm trying to convince myself that there was no particular reason why this was a `let+` before, could you convince me? (I could be convinced by a "I have no idea".)

src/symbolic/symbolic_choice.ml
```diff
@@ -92,21 +92,21 @@ module CoreImpl = struct
...
+        Fun.protect ~finally:callback_end (fun () ->
+          try
+            work wls sched
+              (callback ~close_work_queue:(fun () ->
+                 Work_datastructure.close sched.work_queue ) )
+          with e ->
+            let bt = Printexc.get_raw_backtrace () in
+            (* TODO: I don't understand why we are closing the work queue if one worker dies...
+                    If I remove this line, nothing seens to go wrong with the tests... *)
+            Work_datastructure.close sched.work_queue;
```
redianthus, line 108: @krtab, this part of the code is really hard to follow and I'd like to simplify it.

I noticed I don't understand why we are closing the work queue if one of the workers dies because of an exception. I can not make sense of this.

What I can say:

1. Even without closing the work queue, the execution will be noticed at the end of the execution.
2. Closing prevents other workers to keep working, but we may want them to keep working even if we got a crash for some reason. It'll not prevent error display at the end, but will allow to keep running in case of weird Z3 crashes in a single domain for instance.

Would you say we can remove this one? (Which would allows us to remove the whole `try catch`)

src/symbolic/symbolic_choice.ml
hra687261: Ok, in that case I will merge formalsec/smtml#479 :)

PR #862, opened 2025-12-11
remove bool, i8 and i16 from the FFI, in C we'll use i32 everywhere

redianthus:
instead, this is likely to be cheaper because it avoids costly concat/extract


redianthus:
Surprisingly it seems to degrade performance a lot: https://owi.zulipchat.com/#narrow/channel/486281-Benchmarks/topic/marvin/near/563301616

cc @chambart

src/lang_c/libowi/src/owi.c
```diff
@@ -16,12 +16,27 @@
     _Static_assert(sizeof(T) == 1 || sizeof(T) == 2 || sizeof(T) == 4 ||       \
                        sizeof(T) == 8,                                         \
                    "Unsupported type size!");                                  \
-    (sizeof(T) == 1)   ? (T)(owi_uint8())                                      \
-    : (sizeof(T) == 2) ? (T)(owi_uint16())                                     \
+    (sizeof(T) == 1)   ? (T)(owi_uint32())                                     \
+    : (sizeof(T) == 2) ? (T)(owi_uint32())                                     \
     : (sizeof(T) == 4) ? (T)(owi_uint32())                                     \
                        : (T)(owi_uint64());                                    \
```
filipeom, line 22: why not reduce this to: `sizeof(T) <= 4 ?  (T)(owi_uint32()) : (T)(owi_uint64())` ?

redianthus: Indeed 😅 (but this is all statically computed so it won't make a difference in practice)

PR #863, opened 2025-12-12
clean intf

redianthus:


src/symbolic/symbolic_i32.ml
```diff
@@ -69,45 +81,39 @@ let rotl e1 e2 = binop ty Rotl e1 e2
...
-    | 1l -> cond
-    | _ -> Symbolic_boolean.false_
-  end
-  | _ -> relop Ty_bool Eq e (of_concrete c)
+let eq_concrete (e : t) (c : Int32.t) : Symbolic_boolean.t =
+  relop Ty_bool Eq e (of_concrete c) |> Symbolic_boolean.of_expr
 
-let eq e1 e2 =
-  if phys_equal e1 e2 then Symbolic_boolean.true_ else relop Ty_bool Eq e1 e2
+(* TODO: why do we use `Ty_bool` in the first two cases and `ty` in the others? What is the righe choice? *)
```
redianthus, line 87: cc @filipeom, I can't make sense of this, could you explain how it works to me?

src/symbolic/symbolic_i32.ml
```diff
@@ -69,45 +81,39 @@ let rotl e1 e2 = binop ty Rotl e1 e2
...
-    | 0l -> Bool.not cond
-    | 1l -> cond
-    | _ -> Symbolic_boolean.false_
-  end
-  | _ -> relop Ty_bool Eq e (of_concrete c)
+let eq_concrete (e : t) (c : Int32.t) : Symbolic_boolean.t =
+  relop Ty_bool Eq e (of_concrete c) |> Symbolic_boolean.of_expr
 
-let eq e1 e2 =
-  if phys_equal e1 e2 then Symbolic_boolean.true_ else relop Ty_bool Eq e1 e2
```
redianthus, line 83: cc @filipeom, I removed the `phys_equal` shortcut as it did not seem used that often to me and I also thinks it would make more sense to have it in smtml, WDYT?

redianthus: See https://github.com/formalsec/smtml/pull/484

src/symbolic/symbolic_i32.ml
filipeom: Bitvectors use the default equality operator in SMTLIB (i.e., `=`). Since we pattern match on the theory to dispatch the operator we can just use `Ty_bool` for bitvectors to get the default boolean equality operator. But probably using `ty` would also work. Internally it defaults to the same operator used for `Ty_bool`.

src/symbolic/symbolic_i32.ml
filipeom: merged :smile:

PR #866, opened 2025-12-14
use Smtml.Value.default_of_type

redianthus:
To be merged once https://github.com/formalsec/smtml/pull/485 is published.


filipeom:
I think this one is good to go since https://github.com/ocaml/opam-repository/pull/29098 was merged

PR #870, opened 2025-12-23
add max path depth

### LINKED ISSUE #778: benchmarks depth of paths

S41d:
closes #778

### DIFF
--- a/src/infra/benchmark.ml
+++ b/src/infra/benchmark.ml
@@ -6,6 +6,7 @@ type stats =
   ; solver_intermediate_model_time : Mtime.Span.t Atomic.t
       (* time taken by the solver to answer model queries during the execution, for instance when concretizing through `select_i32` *)
   ; path_count : int Atomic.t
+  ; max_depth : int Atomic.t
   }
 
 let empty_stats () =
@@ -13,6 +14,7 @@ let empty_stats () =
   ; solver_final_model_time = Atomic.make Mtime.Span.min_span
   ; solver_intermediate_model_time = Atomic.make Mtime.Span.min_span
   ; path_count = Atomic.make 1
+  ; max_depth = Atomic.make 1
   }
 
 let handle_time_span atomic_span f =
@@ -32,6 +34,11 @@ let with_utime f =
     let after = (Unix.times ()).tms_utime in
     (r, Some (after -. before))
   else (f (), None)
+
+let set_max_depth stats path_depth =
+  Multicore.atomic_modify
+    (fun max_depth -> if path_depth > max_depth then path_depth else max_depth)
+    stats.max_depth
 
 let percentage ~whole ~self =
   let whole = Mtime.Span.to_float_ns whole in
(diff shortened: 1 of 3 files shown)


redianthus:
(@S41d, I think it needs to be rebased)

src/symbolic/thread.ml
```diff
@@ -101,6 +101,10 @@ struct
 
   let incr_path_count t = Atomic.incr (bench_stats t).path_count
 
+  let incr_path_depth t =
+    Benchmark.incr_path_depth t.bench_stats t.depth;
```
redianthus: Can you move this call to `add_breadcrumb` (where the thread's depth is incremented)? This way, we make sure they are the same and we don't need to have two calls in `Symbolic_choice` for this.

And then, you can call Benchmark.set_max_depth (the function I wrote before) with the depth already available there directly.

src/infra/benchmark.ml
```diff
@@ -39,6 +41,16 @@ let with_utime f =
...
 
+let incr_path_depth stats curr_depth =
+  let rec atomic_set to_add =
+    let curr = Atomic.get stats.max_depth in
+    let success = Atomic.compare_and_set stats.max_depth curr (curr + to_add) in
+    if success then () else atomic_set to_add
+  in
+  let curr_max = Atomic.get stats.max_depth in
+  if curr_depth + 1 > curr_max then atomic_set (curr_depth + 1 - curr_max)
+  else ()
```
redianthus:
```ocaml
let set_max_depth stats path_depth =
  let rec loop () = 
    let max_depth = Atomic.get stats.max_depth in
    if max_depth >= path_depth then ()
    else begin
      let success = Atomic.compare_and_set stats.max_depth max_depth path_depth in
      if success then () else loop ()
    end
  end
  loop ()
```

I found the previous version quite hard to understand compared to what it was supposed to do so I wrote a simpler one.

src/symbolic/thread.ml
S41d: So keep the function where it is but just change where it's called? or remove the function entirely?

redianthus: I think it can be removed entirely

src/infra/benchmark.ml
```diff
@@ -33,6 +35,19 @@ let with_utime f =
...
+    let max_depth = Atomic.get stats.max_depth in
+    if max_depth >= path_depth then ()
+    else begin
+      let success =
+        Atomic.compare_and_set stats.max_depth max_depth path_depth
+      in
+      if success then () else loop ()
+    end
+  in
+  loop ()
```
redianthus: We have `infra/multicore.ml` which has a helper for this now! :) (I'd like to use it everywhere so it can be replaced easily by the version that will be in the stdlib at some point and which is likely more efficient)

S41d: Oh yeah I didn't see that we had that now :), should be fine now

src/infra/benchmark.ml
```diff
@@ -33,6 +35,11 @@ let with_utime f =
     (r, Some (after -. before))
   else (f (), None)
 
+let set_max_depth stats path_depth =
```
redianthus, line 38: Where is this function called? 😅

S41d: Ah, I think I unintentionally removed it during rebase 😅

PR #871, opened 2026-01-06
[WIP] experiment with using landmarks on Owi

hra687261:
The purpose of this PR is to experiment with using a [modified version](https://github.com/hra687261/landmarks/tree/ocaml5) of the [landmarks library](https://github.com/LexiFi/landmarks) which is intended to work with parallelism.

This modified version comes with the following constraints:
1. Each domain has to run on its own physical core to be able to make proper measures.

    That is done through the [domainPC](https://github.com/hra687261/DomainPC) library which is used as a replacement for the stdlib Domain library. 
    I am not sure if there is a simple way through dune to replace an stdlib module with a another one, maybe we can cheat through an `-open` that would hide the `Domain` of the stdlib?

2. Each domain has to properly exit before the parent domain that spawned it, because `at_exit` is used to aggregate the results of the measurements.
 

The current landmarks that were added are only for testing (and so is this PR btw so its probably not worth merging).

To test this, you can run `OCAML_LANDMARKS=on,allocation dune exec -- owi c doc/src/sap/dobble.c`, which will produce (after owi's output):
```
...
Call graph '/home/hra687261/workspace/wasm/owi_landmarks/_build/install/default/bin/owi c doc/src/sap/dobble.c':
----------------------------------------------------------------------------------------------------------------
[    2.32G cycles in  4 calls ]     - 57.07% : Symbolic_choice.CoreImpl.Scheduler.run_worker
[  884.78M cycles in 43 calls ]     |   - 38.18% : Solver.check
[  184.23M cycles in  1 calls ]     |   -  7.95% : Solver.model_of_path_condition
[  798.04M cycles in  1 calls ]     - 19.65% : Cmd_sym.cmd

Note: Nodes accounting for less than 1.00% of their parent have been ignored.

Aggregated table:
----------------
                                         Name;                           Filename;    Calls;     Time; Allocated bytes; Allocated bytes major
                                         ROOT;       src/landmark_state_ocaml5.ml;        0;    4.06G; 122789048; 66476072
Symbolic_choice.CoreImpl.Scheduler.run_worker; src/symbolic/symbolic_choice.ml:95;        4;    2.32G; 10417016; 125256
                                 Solver.check;          src/symbolic/solver.ml:26;       43;  884.78M; 7895576; 55176
                                  Cmd_sym.cmd;              src/cmd/cmd_sym.ml:51;        1;  798.04M; 9972760; 8876200
               Solver.model_of_path_condition;          src/symbolic/solver.ml:30;        1;  184.23M; 536880; 0
                     Interpret.Make.exec_expr;    src/interpret/interpret.ml:1538;        1;    2.16K; 608; 0
```
Notice on the call graph that the sum of the times at the root is not 100% (not yet sure where that time is spent or if that is a measurement error) and that `Symbolic_choice.CoreImpl.Scheduler.run_worker` and `Cmd_sym.cmd` are both at the root, that is because `Cmd_sym.cmd` is run by the main domain while `Symbolic_choice.run_worker` is run by the child domain which makes landmarks see them as "independent" runs. I don't think there is a simple way to stay faithful to the real callgraph, while the time/allocation measurements stay precise for the runs of independent domains.

(cc @redianthus and @bclement-ocp if you want to take a look)

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -80,7 +80,7 @@
    (>= 1.0.3))
   (xmlm
    (>= 1.4.0))
-  (domainPC (= dev))
+  domainpc
   ;; doc
   (odoc
    (and
--- a/owi.opam
+++ b/owi.opam
@@ -42,7 +42,7 @@ depends: [
   "synchronizer" {>= "0.2"}
   "uutf" {>= "1.0.3"}
   "xmlm" {>= "1.4.0"}
-  "domainPC" {= "dev"}
+  "domainpc"
   "odoc" {>= "3.0.0" & with-doc}
   "mdx" {with-test & >= "2.1"}
   "frama-c" {>= "29.0" & with-test}
@@ -84,7 +84,7 @@ post-messages: [
   "Explore tutorials and documentation here: https://ocamlpro.github.io/owi/"
 ]
 pin-depends: [
-  ["domainPC.dev" "git+https://github.com/hra687261/domainPC.git"]
+  ["domainpc.dev" "git+https://github.com/hra687261/domainpc.git"]
   ["landmarks.1.5" "git+https://github.com/hra687261/landmarks.git#cfc017b604f2c41ac0d8841f30a0b55b0d89cbaf"]
   ["landmarks-ppx.1.5" "git+https://github.com/hra687261/landmarks.git#cfc017b604f2c41ac0d8841f30a0b55b0d89cbaf"]
 ]
--- a/src/dune
+++ b/src/dune
@@ -166,7 +166,7 @@
   mtime
   mtime.clock
   xmlm
-  domainPC
+  domainpc
   landmarks
   landmarks-ppx)
  (flags
--- a/src/symbolic/symbolic_choice.ml
+++ b/src/symbolic/symbolic_choice.ml
@@ -117,7 +117,7 @@ module CoreImpl = struct
     let spawn_worker sched wls_init ~at_worker_value ~at_worker_init
       ~at_worker_end =
       at_worker_init ();
-      DomainPC.spawn (run_worker sched wls_init ~at_worker_value ~at_worker_end)
+      Domainpc.spawn (run_worker sched wls_init ~at_worker_value ~at_worker_end)
   end
 
   module State = struct
@@ -271,7 +271,7 @@ module CoreImpl = struct
            (close_work_queue:(unit -> unit) -> 'a eval * thread -> unit)
       -> at_worker_init:(unit -> unit)
       -> at_worker_end:(unit -> unit)
-      -> unit DomainPC.t array
+      -> unit Domainpc.t array
   end = struct
     include Eval
 
--- a/src/symbolic/symbolic_driver.ml
+++ b/src/symbolic/symbolic_driver.ml
@@ -83,7 +83,7 @@ let handle_result ~exploration_strategy ~workers ~no_stop_at_failure ~no_value
     mk_callback no_stop_at_failure fail_mode res_stack path_count
   in
   let time_before = (Unix.times ()).tms_utime in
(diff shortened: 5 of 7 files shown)


src/dune
```diff
@@ -165,11 +165,14 @@
   runtime_events
   mtime
   mtime.clock
-  xmlm)
+  xmlm
+  domainPC
+  landmarks
+  landmarks-ppx)
```
redianthus, line 171: I think the `ppx` should be removed from the list of runtime dependencies.

src/dune
```diff
@@ -165,11 +165,14 @@
...
  (flags
   (:standard -open Prelude))
  (preprocess
-  (pps sedlex.ppx))
+  (pps sedlex.ppx landmarks-ppx))
```
redianthus, line 175: Is this actually needed when using the `instrumentation` (see the next lines of the dune files).

Have you tried with [this command](https://ocamlpro.github.io/owi/hacking/benchmarking.html#landmarks)? If it works, I'd prefer to only keep it as it is more flexible (as it allows to avoid the linking/preprocessing/installation of the ppx when one does not need it).

dune-project
```diff
@@ -80,6 +80,7 @@
    (>= 1.0.3))
   (xmlm
    (>= 1.4.0))
+  (domainPC (= dev))
```
redianthus:
```suggestion
  domainpc
```

owi.opam
```diff
@@ -42,6 +42,7 @@ depends: [
   "synchronizer" {>= "0.2"}
   "uutf" {>= "1.0.3"}
   "xmlm" {>= "1.4.0"}
+  "domainPC" {= "dev"}
```
redianthus: will need a `dune build` to get updated once the dune-project is fixed

src/symbolic/symbolic_choice.ml
```diff
@@ -268,7 +271,7 @@ module CoreImpl = struct
            (close_work_queue:(unit -> unit) -> 'a eval * thread -> unit)
       -> at_worker_init:(unit -> unit)
       -> at_worker_end:(unit -> unit)
-      -> unit Domain.t array
+      -> unit DomainPC.t array
```
redianthus:
```suggestion
      -> unit Domainpc.t array
```

src/symbolic/symbolic_driver.ml
```diff
@@ -83,7 +83,7 @@ let handle_result ~exploration_strategy ~workers ~no_stop_at_failure ~no_value
     mk_callback no_stop_at_failure fail_mode res_stack path_count
   in
   let time_before = (Unix.times ()).tms_utime in
-  let domains : unit Domain.t Array.t =
+  let domains : unit DomainPC.t Array.t =
```
redianthus:
```suggestion
  let domains : unit Domainpc.t Array.t =
```

src/symbolic/symbolic_driver.ml
```diff
@@ -101,7 +101,7 @@ let handle_result ~exploration_strategy ~workers ~no_stop_at_failure ~no_value
   let wait_for_all_domains () =
     Array.iter
       (fun domain ->
-        try Domain.join domain with
+        try DomainPC.join domain with
```
redianthus:
```suggestion
        try Domainpc.join domain with
```

src/dune
```diff
@@ -165,11 +165,14 @@
   runtime_events
   mtime
   mtime.clock
-  xmlm)
+  xmlm
+  domainPC
+  landmarks
```
redianthus, line 170: I'd like to avoid the landmarks dependency. IIUC, this is only used here:

> Landmark.profiling ()

Do you think we could use an environment variable instead or something else?

PR #875, opened 2026-01-16
Reduce closure creation after removing WLS reader monad

krtab:
I think this will greatly improve performances because we are now hiding much less behind closures.


redianthus:
Thanks!

PR #878, opened 2026-01-18
remove functor hell

redianthus:
cc @krtab, @filipeom and @hra687261

I'm pleased to announce that all the `{symbolic_choice,thread}_{with,without}_memory` stuff is gone! Have a look at the dune file to see how pleasant the result is :-)

I was getting really annoyed by this and decided to get rid of it. Actually, it was not that hard. There was a dependency cycle between thread, memory and symbolic choice, but it was not a real one... Thread was not really depending on memory, but only on so called "memory collections", which was included in memory. I simply moved it to its own module and it was enough to break the cycle.

While doing this change, I first thought that I could get rid of the `lift_mem`, and `project,restore` thread stuff. This is what I tried in the first commit, but a few tests were failing (collections-c stuff, that are quite good at spotting memory related bugs). I decided to bring it back to test, and now it works perfectly. Nonetheless, I miss some context about why it was introduced, and why it is done this way. In particular, I have some vague feeling that this is an ugly hack and that we could probably get rid of it. I'd happy to discuss it if any of you have some idea!

### DIFF
--- a/src/symbolic/bug.ml
+++ b/src/symbolic/bug.ml
@@ -1,10 +1,7 @@
 type t =
   { kind : [ `Trap of Result.err | `Assertion of Symbolic_boolean.t ]
   ; model : Smtml.Model.t
-      (* TODO: all that follows could be replaced by the Thread directly! *)
-  ; labels : (int * string) list
-  ; breadcrumbs : int list
-  ; symbol_scopes : Symbol_scope.t
+  ; thread : Thread.t
   }
 
 let is_trap { kind; _ } =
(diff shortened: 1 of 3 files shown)


src/symbolic/state_monad.ml
```diff
@@ -1,6 +1,20 @@
...
+    's -> ('a -> 's -> 'r) -> 'r
+
+  let return x =
+    fun s k ->
+      k x s
+
+  let bind m f =
+    fun s k ->
+      m s (fun x s -> f x s k)
+  *)
```
redianthus, line 17: cc @krtab if you want to have fun with monads

krtab: image

hra687261 (approved):
That's pretty amazing, the code is much more readable!
Does this have any effect on the performance/benchmarks?

PR #933, opened 2026-03-03
Abstract interpretation

S41d:


src/abs/abs_datastructures.ml
```diff
@@ -0,0 +1,25 @@
...
+  (*   let popped = List.take n stack in *)
+  (*   if List.length popped = n then Ok (popped, List.drop n stack) *)
+  (*   else Fmt.error_msg "not enough values to pop" *)
+
+  let pop_n : t -> int -> t * t =
+   fun stack n ->
+    let popped = List.take n stack in
+    if List.length popped = n then (popped, List.drop n stack)
+    else Fmt.failwith "not enough values to pop"
+end
```
redianthus: I guess having only `abs_stack.ml` is enough for now and will prevent this module to become too big in the future

src/abs/abs_driver.ml
```diff
@@ -201,50 +168,66 @@ let rec exec_func (state : state) (func : Func.t) =
...
   | I32_const i ->
-    let abs_i = Domain.Integer_Forward.iconst (Z.of_int32 i) state.ctx in
+    let abs_i =
+      AbsDomain.Binary_Forward.biconst ~size:size_s32 (Z.of_int32 i) state.ctx
+    in
     { state with stack = abs_i :: state.stack }
   | I64_const i ->
-    let abs_i = Domain.Integer_Forward.iconst (Z.of_int64 i) state.ctx in
+    let abs_i =
+      AbsDomain.Binary_Forward.biconst ~size:size_s32 (Z.of_int64 i) state.ctx
```
redianthus: is the `size_s32` expected or a copy-paste error?

S41d: Haha, I was wondering where did I miss up for the pretty printer to error. Int64 was supposed to be int32 (for now)

PR #954, opened 2026-04-09
Try to fix CI

filipeom:
Also try to fix opam-based installations

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -59,7 +59,7 @@
   (bos
    (>= 0.2.1))
   (cmdliner
-   (>= 2.1.0))
+   (>= 1.3.0))
   (digestif
    (>= 1.2.0))
   (domainpc
@@ -75,7 +75,7 @@
   (sedlex
    (>= 3.3))
   (smtml
-   (>= 0.22.0))
+   (>= 0.24.0))
   (symex
    (>= 0.2))
   (synchronizer
@@ -84,6 +84,8 @@
    (>= 1.0.3))
   (xmlm
    (>= 1.4.0))
+  (ocamlgraph
+   (>= 2.2.0))
   ;; doc
   (odoc
    (and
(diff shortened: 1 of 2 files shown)


filipeom:
Ah right, might need to pin a couple of things I think

filipeom:
If we dropped the constraint on cmdliner `>= 2.1.0` to `>= 1.3.0` it would build with opam because alt-ergo's latest available opam version conflicts with cmdliner `> 2.0.0`. However, it wouldn't pass a bunch of tests because cmdliner's version 2 changes the output of the CLI. I tried pinning the latest dev version `alt-ergo` and `alt-ergo-lib` because it already supports cmdliner 2, but then we have another problem: it doesn't compile with `smtml` :sweat_smile:. I'm giving up, so I'll close this. 

Let's hope we get a new version of alt-ergo soon :sweat_smile:

redianthus:
Haha, yes sorry for not telling you, I've talked with the alt-ergo people and there should be a release soon with cmdliner > 2 allowed. :D I'm waiting for it and then the CI should be back.

filipeom:
> Haha, yes sorry for not telling you, I've talked with the alt-ergo people and there should be a release soon with cmdliner > 2 allowed. :D I'm waiting for it and then the CI should be back.

Ah nice! I think it needs a dolmen release as well right? https://github.com/OCamlPro/alt-ergo/blob/7398dde07d80483f72e85a926d33d2e0de728bec/alt-ergo-lib.opam#L64-L77

redianthus:
Well, I thinks this is the `next` branch which needs the dolmen release, whereas they're going to actually to do a minor release which will not need the dolmen release IIRC

filipeom:
> Well, I thinks this is the `next` branch which needs the dolmen release, whereas they're going to actually to do a minor release which will not need the dolmen release IIRC

Ah great! I was asking because smtml also needed to be upgraded if dolmen was bumped

PR #955, opened 2026-04-10
Prepare for smtml.0.25.0

filipeom:
Use the new bitvector API that requires passing bit offsets instead of the previous non-standard byte offsets.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -75,7 +75,7 @@
   (sedlex
    (>= 3.3))
   (smtml
-   (>= 0.22.0))
+   (>= 0.25.0))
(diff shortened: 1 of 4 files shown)


redianthus:
Oh, thanks for taking care of it!

I thought it would be more invasive than this, but apparently it's not! :D

filipeom:
> I thought it would be more invasive than this, but apparently it's not! :D

It also surprised me as well. But I guess we moved almost everything into smtml, so the change was mostly contained :smiley:

redianthus:
Thanks!

PR #958, opened 2026-04-17
big AST refactoring

redianthus:
to prepare for SIMD proposal addition


redianthus:
@hra687261, this is now all good and you can use this as a base for your GC work. The fuzzer still needs to be updated, but that something that should not matter for your work :)

PR #962, opened 2026-04-21
WasmGC parsing, encoding/decoding

hra687261:
Ideally go as far as validation/type-checking


src/ir/binary.ml
```diff
@@ -132,6 +149,55 @@ let pp_func_type ppf (params, results) =
...
+  ; ids : indice list
+  ; ct : comp_type
+  }
+
+let sub_type_eq { final = f1; ct = ct1; _ } { final = f2; ct = ct2; _ } =
+  Bool.equal f1 f2 && comp_type_eq ct1 ct2
+
+let pp_sub_type fmt { final; ids; ct } =
+  Fmt.pf fmt "%a%a%a"
+    (fun fmt b -> if b && not (List.is_empty ids) then Fmt.pf fmt "final ")
```
redianthus, line 196: I'm not sure we should print `final` if there is no ids. IIRC a module can be declared as "not final" and be referenced in another one. But I need to refresh my memory on this so this may be nonsense 😅

src/ir/binary_encoder.ml
```diff
@@ -69,20 +69,17 @@ let write_char_indice buf c idx =
...
 
-let write_reftype buf nullable ht =
-  let is_null =
-    match nullable with
-    | (Text.Null : Text.nullable) -> true
-    | No_null ->
-      Buffer.add_char buf '\x64';
-      false
-  in
+let write_heaptype ?(is_null = false) buf ht =
```
redianthus: Can you make `is_null` named instead of optional? It's too easy to forget it otherwise.

src/ir/binary_encoder.ml
```diff
@@ -756,12 +893,53 @@ let encode_section buf id encode_func data =
...
+
+let encode_sub_type buf st =
+  match st with
+  | { final = true; ids = []; ct } -> encode_comp_type buf ct
+  | { final = true; ids; ct } ->
+    Buffer.add_char buf '\x4F';
+    List.iter (write_indice buf) ids;
+    encode_comp_type buf ct
+  | { final = false; ids; ct } ->
+    Buffer.add_char buf '\x50';
```
redianthus:
```suggestion
    Buffer.add_char buf (if final then '\x4F' else '\x50');
```

And then merge the two cases?

src/ir/binary_encoder.ml
hra687261: I just removed it entirely since the special case is only used in `ref_type`.

src/ir/binary.ml
hra687261: > I'm not sure we should print final if there is no ids.

We only print final if `b && not (List.is_empty ids)` (with `b` as `is_final`), so we don't print `final` if there are no ids :thinking:

redianthus:
Thanks! As discussed, I'm merging as is to avoid painful rebase in the future

PR #965, opened 2026-04-23
Check copyright header for new .ml and .mli files

### LINKED ISSUE #260: check copyright header on new files in CI

halalbit:
## What does this PR do?
Add a new CI jobs that check new committed files (.ml and .mli files) if the copyright headers is here or not.

## Why is this change needed?
This change is needed to have a uniform header for .ml and .mli files

## How to test?
You can test it locally with [act](https://github.com/nektos/act)
with this cmd : ``~/act/bin/act pull_request -W .github/workflows/check-header.yml`` after commiting files

## Issue related
Closes #260 

Disclaimer: this contribution is part of work carried out as a student at Université Paris-Cité. 
Apologies if this is inappropriate or if the contribution does not meet the project’s standards 
— any feedback is welcome.


redianthus:
Thanks! I think you can make this much simpler. First, you can probably remove the `check-headers.sh` script entirely.
In the CI, do something like:

```
missing_copyright = false

for each .ml or .mli file $f {
  grep for "Copyright © 2021-2024 OCamlPro" in file $f, if not present, print a message and set `missing_copyright` to true, otherwise do nothing
}

if missing_copyright is true, make the CI fail (you only want to do this after processing all the files so we can get a proper list of the failing files, which is more pleasant than having to run the CI many times)
```

halalbit:
## What have been done 
Before two files needed : 
``.github/check-headers.sh`` and ``.github/workflows/check-header.yml``
Now :
``github/workflows/check-header.yml`` with a simplier logic.

If you have any more suggestion or remarks feel free to tell me !

.github/workflows/check-header.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Check License Header
+        run: |
+          files=$(git diff --name-only --diff-filter=A \
+            "origin/${{ github.base_ref }}"...HEAD \
+            -- '*.ml' '*.mli')
+          if [ -z "$files" ]; then
+            echo "No new .ml or .mli files in this PR."
+            exit 0
+          fi
```
redianthus: I think we can make this even simpler and remove the check about new files. It'll also allows to 1. change the year easily and do the check on all files 2. make sure the header is not removed from pre-existing files.

You'll also have to change the `git diff` and use something which includes all files.

redianthus:
Thanks! I added a last remark and then I think this should be ready to merge.

I'll take care of updating the copyright on all existing files, don't bother with it (it'll also help to check it's actually working when the copyright is wrong or missing).

halalbit:
Done !

redianthus:
Thanks!

PR #971, opened 2026-04-27
Add ETAPS Industry Day talk

hra687261:
These lists should probably be in anti-chronological order (to make the most recent ones appear first)


redianthus:
Thanks! :D

While you're at it, could you add @S41d's talk at JFLA just before yours?

hra687261:
Done!

redianthus:
Thanks!

PR #972, opened 2026-04-28
Fix the update-testsuite workflow

hra687261:
The PR updates the testsuite and fixes the script to ensure that it runs the CI after every push, updates the right subrepo (since it changed), and only attempt to create a PR only if the PR does not already exist.


redianthus:
Thanks!

PR #980, opened 2026-05-20
update to latest frama-c

redianthus:
cc @S41d, for some reason, the E-ACSL instrumentation fails to see the macro and only sees `owi_int(string)`, thus it is failing on a call to it without a name. I went the easy way and added a name everywhere, WDYT?


S41d:
I tried playing with it a bit to see if I can make it work but I haven't found a working solution. Seems like frama-c is doing the preprocessing themselves so that's probably the culprit

PR #989, opened 2026-05-27
add a bunch of missing stuff for denot, add br_if to abs, and other fixes

S41d:
add some tests too

### DIFF
--- a/src/abstract/abstract_boolean.mli
+++ b/src/abstract/abstract_boolean.mli
@@ -2,7 +2,7 @@
 (* Copyright © 2021-2026 OCamlPro *)
 (* Written by the Owi programmers *)
 
-type t = Abstract_domain.Boolean.t
+type t = Abstract_domain.boolean
 
 val false_ : Abstract_domain.Context.t -> t
 
--- a/src/abstract/abstract_driver.ml
+++ b/src/abstract/abstract_driver.ml
@@ -189,25 +189,24 @@ module DenotFixpoint (S : DATA_STATE) = struct
       | Num_type I64 -> I64 (Abstract_i64.zero state.ctx)
       | _ -> assert false
     in
+
     let locals =
-      List.mapi (fun i (_str_opt, vt) -> (i, init_value vt)) func.locals
+      (* TODO check args position *)
+      args @ List.map (fun (_str_opt, vt) -> init_value vt) func.locals
+      |> List.mapi (fun i x -> (i, x))
       |> Abstract_locals.of_list
     in
     let fn_state =
-      { state with
-        stack = args
-      ; ctx = state.ctx
-      ; func_rt = result_type
-      ; locals
-      }
+      { state with stack = []; ctx = state.ctx; func_rt = result_type; locals }
     in
-    Log.info (fun m -> m "Func start state : %a" Abstract_state.pp fn_state);
+    Log.debug (fun m ->
+      m "Func start state : %a" (Abstract_state.pp state.ctx) fn_state );
     (* TODO: handle mapping *)
     let func_end_state, _ = eval_expr fn_state func.body.raw in
-    Log.info (fun m ->
-      m "Func end state : %a@."
-        (Fmt.option ~none:(Fmt.any "None") Abstract_state.pp)
-        func_end_state );
+    (* Log.debug (fun m -> *)
+    (*   m "Func end state : %a@." *)
+    (*     (Fmt.option ~none:(Fmt.any "None") (Abstract_state.pp func_end_state.ctx)) *)
+    (*     func_end_state ); *)
     (* We should probably copy state and join back the return values in the context here *)
     let* func_end_state in
     let stack =
@@ -222,7 +221,7 @@ module DenotFixpoint (S : DATA_STATE) = struct
     Log.debug (fun m ->
       m "#%a\t\t%a@."
         (Binary.pp_instr ~short:true)
-        instr Abstract_state.pp state );
+        instr (Abstract_state.pp ctx) state );
     match instr with
     | Call idx ->
       let func = Link_env.get_func env idx in
@@ -254,21 +253,24 @@ module DenotFixpoint (S : DATA_STATE) = struct
         | Some br_states -> List.fold_left join state br_states
         | None -> state
       in
+      let state =
+        match JumpTarget.find_opt Ret mapping with
+        | Some ret_states -> List.fold_left join state ret_states
+        | None -> state
+      in
       let mapping =
         (* TODO on peut avoir une paire de (int * map) pour ne pas avoir à decr la liste immédiatement *)
         JumpTarget.decr mapping
       in
       (Some state, mapping)
     | If_else (_, bt, expr_then, expr_else) ->
-      (* TODO: use `pop_bool` once I32.to_boolean is properly implemented *)
-      let b, stack = Stack.pop stack in
-      let cond = Abstract_value.to_boolean ctx b in
+      let b, stack = Stack.pop_bool stack ctx in
       let state_then, jt_true =
-        let> ctx, _ = (Abstract_domain.assume ctx cond, JumpTarget.empty) in
+        let> ctx, _ = (Abstract_domain.assume ctx b, JumpTarget.empty) in
         eval_instr { state with stack; ctx } (Block (None, bt, expr_then))
       in
       let state_else, jt_false =
-        let not_cond = Abstract_boolean.not ctx cond in
+        let not_cond = Abstract_boolean.not ctx b in
         let> ctx, _ = (Abstract_domain.assume ctx not_cond, JumpTarget.empty) in
         eval_instr { state with stack; ctx } (Block (None, bt, expr_else))
       in
@@ -305,7 +307,7 @@ module DenotFixpoint (S : DATA_STATE) = struct
                 join acc { state with stack } )
               { initial_state with stack = fp_stack }
               jts
-          | None -> assert false
+          | None -> state
         in
         let widened, included = widen widening_id state next_head in
         if not included then fixpoint widened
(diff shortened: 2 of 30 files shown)


src/abstract/abstract_boolean.mli
```diff
@@ -2,7 +2,7 @@
 (* Copyright © 2021-2026 OCamlPro *)
 (* Written by the Owi programmers *)
 
-type t = Abstract_domain.Boolean.t
+type t = Abstract_domain.boolean
```
redianthus, line 5: Is there actually a difference between the two or is this only cosmetic?

src/abstract/abstract_driver.ml
```diff
@@ -189,25 +189,24 @@ module DenotFixpoint (S : DATA_STATE) = struct
...
       |> Abstract_locals.of_list
     in
     let fn_state =
-      { state with
-        stack = args
-      ; ctx = state.ctx
-      ; func_rt = result_type
-      ; locals
-      }
+      { state with stack = []; ctx = state.ctx; func_rt = result_type; locals }
```
redianthus: Isn't `ctx = state.ctx` redundant with `{ state with ...` ?

src/abstract/abstract_boolean.mli
S41d: I think they're the same

src/abstract/abstract_driver.ml
```diff
@@ -189,25 +189,24 @@ module DenotFixpoint (S : DATA_STATE) = struct
...
     (* TODO: handle mapping *)
     let func_end_state, _ = eval_expr fn_state func.body.raw in
-    Log.info (fun m ->
-      m "Func end state : %a@."
-        (Fmt.option ~none:(Fmt.any "None") Abstract_state.pp)
-        func_end_state );
+    (* Log.debug (fun m -> *)
+    (*   m "Func end state : %a@." *)
+    (*     (Fmt.option ~none:(Fmt.any "None") (Abstract_state.pp func_end_state.ctx)) *)
+    (*     func_end_state ); *)
```
redianthus: Could you bring it back? (I think it's only a matter of adding a `match` before printing)

src/abstract/abstract_driver.ml
```diff
@@ -321,6 +323,19 @@ module DenotFixpoint (S : DATA_STATE) = struct
...
+    | Br_if i ->
+      let b, stack = Stack.pop_bool stack ctx in
+      let jt_if =
+        match Abstract_domain.assume ctx b with
+        | Some _ -> JumpTarget.of_list [ (I i, [ state ]) ]
+        | None -> JumpTarget.empty
+      in
+      let state =
+        match Abstract_domain.assume ctx (Abstract_boolean.not ctx b) with
+        | Some _ -> Some { state with stack }
```
redianthus: The context should added here?

src/abstract/abstract_driver.ml
```diff
@@ -321,6 +323,19 @@ module DenotFixpoint (S : DATA_STATE) = struct
       in
       fixpoint state
     | Br i -> (None, JumpTarget.of_list [ (I i, [ state ]) ])
+    | Br_if i ->
+      let b, stack = Stack.pop_bool stack ctx in
+      let jt_if =
+        match Abstract_domain.assume ctx b with
+        | Some _ -> JumpTarget.of_list [ (I i, [ state ]) ]
```
redianthus: Shouldn't the context be added to the state passed to the target?

src/cmd/cmd_drun.ml
```diff
@@ -0,0 +1,12 @@
+(* SPDX-License-Identifier: AGPL-3.0-or-later *)
+(* Copyright © 2021-2026 OCamlPro *)
+(* Written by the Owi programmers *)
+
+open Syntax
+
+let cmd ~source_file ~no_input =
+  let link_state = Link.State.empty () in
+  let+ m, link_state =
+    Compile.File.until_link ~unsafe:true ~name:None link_state source_file
```
redianthus: Can you add a `unsafe:bool` paramter to the function and in `bin/owi.ml` add `let+ unsafe in ... Cmd.drun cmd ~unsafe` ?

It'll save you a lot of time in the future as debugging something which is actually not well typed can be very frustrating

src/abstract/abstract_i32.ml
```diff
@@ -17,7 +17,9 @@ let of_boolean ctx boolean =
   let n = if Abstract_boolean.equal boolean true_ then 1 else 0 in
   Abstract_domain.Binary_Forward.biconst ~size (Z.of_int n) ctx
 
-let to_boolean _ = assert false
+let to_boolean ctx x =
+  let zero = Abstract_domain.Binary_Forward.biconst ~size Z.zero ctx in
+  Abstract_domain.Binary_Forward.beq ~size ctx x zero
```
redianthus: Something seems weird here. You are basically doing:

```c
int x = ...;
bool b = (x == 0);
```

Which seems to be the opposite of what should be done? (i.e. it is true if x!=0)

src/denot/denot_interpreter.ml
```diff
@@ -15,12 +17,15 @@ module State = struct
...
     ; locals : Value.t Locals.t
+    ; env : Concrete_extern_func.extern_func Link_env.t
+    ; envs : Concrete_extern_func.extern_func Link_env.t Dynarray.t
+    ; func_rt : Binary.val_type list
     }
 
-  let pp_map = Locals.pretty (fun fmt k v -> Fmt.pf fmt "%i->%a" k Value.pp v)
+  let pp_map = Locals.pretty (fun fmt k v -> Fmt.pf fmt "%i->%a " k Value.pp v)
 
   let pp ppf (state : t) =
```
redianthus:
```suggestion
  let pp ppf ({stack; locals; _} : t) =
```

src/abstract/abstract_i32.ml
S41d: yes 😅

src/abstract/abstract_driver.ml
```diff
@@ -189,25 +189,24 @@ module DenotFixpoint (S : DATA_STATE) = struct
...
-      ; func_rt = result_type
-      ; locals
-      }
+      { state with stack = []; ctx = state.ctx; func_rt = result_type; locals }
     in
-    Log.info (fun m -> m "Func start state : %a" Abstract_state.pp fn_state);
+    Log.debug (fun m ->
+      m "Func start state : %a" (Abstract_state.pp state.ctx) fn_state );
     (* TODO: handle mapping *)
     let func_end_state, _ = eval_expr fn_state func.body.raw in
```
redianthus, line 222: Could you do something like:

```ml
let func_end_state, _a_descriptive_name = ... in
```

Or even:

```ml
let func_end_state, (_a_descriptive_name : type_annotation) = ... in
```

In many places I wondering if there is a value that is ignored when it should not, and it's hard to know what is this value when the only information around is `_`. (There are many others cases I noticed while reading, I'll only comment them with "add info about ignored value")

src/abstract/abstract_driver.ml
```diff
@@ -254,21 +253,24 @@ module DenotFixpoint (S : DATA_STATE) = struct
...
       in
       (Some state, mapping)
     | If_else (_, bt, expr_then, expr_else) ->
-      (* TODO: use `pop_bool` once I32.to_boolean is properly implemented *)
-      let b, stack = Stack.pop stack in
-      let cond = Abstract_value.to_boolean ctx b in
+      let b, stack = Stack.pop_bool stack ctx in
       let state_then, jt_true =
-        let> ctx, _ = (Abstract_domain.assume ctx cond, JumpTarget.empty) in
+        let> ctx, _ = (Abstract_domain.assume ctx b, JumpTarget.empty) in
```
redianthus: add info about ignored value

src/abstract/abstract_driver.ml
```diff
@@ -254,21 +253,24 @@ module DenotFixpoint (S : DATA_STATE) = struct
...
         eval_instr { state with stack; ctx } (Block (None, bt, expr_then))
       in
       let state_else, jt_false =
-        let not_cond = Abstract_boolean.not ctx cond in
+        let not_cond = Abstract_boolean.not ctx b in
         let> ctx, _ = (Abstract_domain.assume ctx not_cond, JumpTarget.empty) in
```
redianthus: add info about ignored value

src/abstract/abstract_driver.ml
```diff
@@ -514,7 +530,7 @@ let expr (link_state : Abstract_extern_func.extern_func Link.State.t)
     begin fun (e : Binary.expr Annotated.t) ->
       let end_state, _ = ConcreteFixpoint.eval_expr state e.raw in
```
redianthus: add info about ignored value

src/denot/denot_interpreter.ml
```diff
@@ -109,21 +112,103 @@ module Fixpoint (DS : DATA_STATE) = struct
...
+    let locals =
+      (* TODO check args position *)
+      args @ List.map (fun (_str_opt, vt) -> init_value vt) func.locals
+      |> List.mapi (fun i x -> (i, x))
+      |> Locals.of_list
+    in
+    let fn_state = { state with stack = args; func_rt = result_type; locals } in
+    Log.debug (fun m -> m "Func start state : %a" State.pp fn_state);
+    (* TODO: handle mapping *)
+    let func_end_state, _ = eval_expr ~no_input fn_state func.body.raw in
```
redianthus: add info about ignored value

src/denot/denot_interpreter.ml
```diff
@@ -229,10 +373,23 @@ end
...
+      { stack = Stack.empty
+      ; locals = Locals.empty
+      ; env = m.env
+      ; envs
+      ; func_rt = []
+      }
+  in
+  List.iter
+    begin fun (e : Binary.expr Annotated.t) ->
+      let end_state, _ = Concrete_fixpoint.eval_expr ~no_input state e.raw in
```
redianthus: add info about ignored value

test/cram/abs/loop.wat
```diff
@@ -6,8 +6,10 @@
       (local.get $res)
       i32.const 2
       i32.add
-      (local.tee $res)
-      br 0)
+      local.set $res
+      ;; ce qui marche pas : (i32.le_s (i32.const 100) (local.get $res))
```
redianthus: We should open an issue so we don't forget to have a look at this buggy case in the future

redianthus: See #990

redianthus:
Thanks!

PR #1002, opened 2026-06-11
WasmGC validation and concrete execution

hra687261:


src/compile/assigned.ml
```diff
@@ -57,36 +59,80 @@ module Typetbl = Hashtbl.Make (struct
...
+            match field_id with
+            | Some (Text.Text n) when Hashtbl.mem tbl n ->
+              Error (`Msg "duplicate field")
+            | Some (Text.Text n) ->
+              Hashtbl.add tbl n idx;
+              Ok ()
+            | Some (Text.Raw _) | None -> Ok ()
+          in
+          Ok (idx + 1) )
+        0 fields
```
redianthus, line 81: If I understand correctly, this is `list_fold_left_i`? Maybe you could define it and use it here? The files in `compile/` are the most complicated and I try to make them as small as possible...

redianthus: Or maybe, you could use a `Dynarray` instead of an `Hashtbl`? (hum, maybe not, I don't think this is what you're doing, sorry!)

redianthus:
@hra687261, I rebased on main and force-pushed here.

hra687261:
Marking this as ready for review, since this PR is already big, I think we can merge the concrete version with the validation, the symbolic one will be added in a separate PR

src/link/link.ml
redianthus, line 1: Should be removed and we should port your changes to the new linker (env/env.ml)

src/concrete/concrete_ref.ml
```diff
@@ -18,42 +18,164 @@ module Extern = struct
...
 end
 
-type t =
+type gc_val =
+  | I32 of int32
+  | I64 of int64
+  | F32 of Float32.t
+  | F64 of Float64.t
+  | V128 of Concrete_v128.t
+  | Ref of t
```
redianthus: I feel like this is wrong.
This is basically redefining `Concrete_value.t` and I fear having this type twice is going to make things complicated in the future.
IIUC, you have to do this because otherwise, there's a cycle between ref and values, right?

What about adding a type parameter, that is :

```ocaml
and 'value gc_obj =
  { obj_id : int32
  ; type_id : int
  ; fields : 'value array
  }

and 'value struct_obj = 'value gc_obj

and 'value array_obj = 'value gc_obj

and 'value t =
  | Extern of Extern.t option
  | Func of int option
  | NullExn
  | NullRef
  | I31 of int32
  | NullI31
  | Array of 'value array_obj
  | Struct of 'value struct_obj
  | ExternAsAny of Extern.t option
```

And in `Concrete_value`, simply have `| Ref of t Ref.t` ?

src/intf/ref_intf.ml
```diff
@@ -14,13 +23,22 @@ module type T = sig
...
+  type struct_obj
+
   (* TODO; make this private and even opaque at some point *)
   type t =
     | Extern of Extern.t option
     | Func of int option
     (* TODO: Not sure about these two. *)
     | NullExn
     | NullRef
+    | I31 of int32
```
redianthus: It should be at least `i32` (where `i32` is an opaque type defined before). Otherwise, in symbolic, we won't be able to have a proper symbolic value here.

(And having a proper `i31` type would be even better)

src/intf/link_intf.ml
redianthus, line 1: should be removed too

src/concrete/concrete_ref.ml
```diff
@@ -18,42 +18,164 @@ module Extern = struct
...
+  | I31 of int32
+  | NullI31
+  | Array of array_obj
+  | Struct of struct_obj
+  | ExternAsAny of Extern.t option
+
+let any_as_extern_key : t Type.Id.t = Type.Id.make ()
+
+(* Concrete execution is not parallel, so this should be fine *)
+let obj_id_counter = ref 0l
```
redianthus:
```suggestion
let obj_id_counter = ref 0
```

I don't think an `Int32` is required here, right? And it should avoid the boxing.

src/concrete/concrete_ref.ml
```diff
@@ -18,42 +18,164 @@ module Extern = struct
...
+  | Func (Some _)
+  | Extern (Some _)
+  | I31 _ | Array _ | Struct _
+  | ExternAsAny (Some _) ->
+    false
+
+let ref_eq (r1 : t) (r2 : t) : bool =
+  if is_null r1 && is_null r2 then true
+  else if is_null r1 || is_null r2 then false
+  else
```
redianthus, line 93:
```suggestion
  if is_null r1 || is_null r2 then is_null r1 && is_null r2
  else
```

src/intf/ref_intf.ml
```diff
@@ -2,6 +2,15 @@
...
 (* Written by the Owi programmers *)
 
+(* TODO: used for conversion, can it be removed? *)
+type 'ref gc_view =
+  | GCv_i32 of int32
+  | GCv_i64 of int64
+  | GCv_f32 of Float32.t
+  | GCv_f64 of Float64.t
+  | GCv_v128 of Concrete_v128.t
+  | GCv_ref of 'ref
```
redianthus: OK, so, my previous comment rather apply here. This is `Value_intf`. I think adding a type parameter should be enough to remove the annoying cycle and this redefinition.

src/env/env.ml
```diff
@@ -197,6 +198,10 @@ module Make
     ; context : Context.t
     ; raw_names : modul StringMap.t
         (* this is used only for scripts where modules can get a $id and we have to remember them to be able to register them this way... *)
+    ; types : Binary.sub_type array IntMap.t
```
redianthus:
```ocaml
    ; types : Binary.sub_type Iarray.t IntMap.t
```

might be safer?

and same for type_groups

src/env/env.ml
```diff
@@ -942,9 +953,8 @@ module Make
       | Memory i -> Memory (rewrite_memory_instruction i)
       | Data i -> Data (rewrite_data_instruction i)
       | ( Nop | Local _ | Drop | Unreachable | Any_convert_extern
-        | Extern_convert_any | Select _ ) as i ->
+        | Extern_convert_any | Select _ | I31 _ | Struct _ | Array _ ) as i ->
```
redianthus: I feel like the struct and array case might need to be rewritten in order to update type index at some point?

src/env/env.ml
```diff
@@ -1208,9 +1227,13 @@ module Make
     | Some v -> v
     | None -> assert false
 
-  let get_types ~env:_ = assert false
+  let get_types ~env ~modul =
```
redianthus: I don't think there should be a `~modul` here, I've tried to remove this entirely in the current linker. Why do you need it?

src/env/env.ml
```diff
@@ -1208,9 +1227,13 @@ module Make
...
+    match IntMap.find_opt modul env.types with Some v -> v | None -> [||]
+
+  let get_type_groups ~env ~modul =
```
redianthus: same comment

src/ir/binary.ml
```diff
@@ -1238,7 +1238,9 @@ module Typedef = struct
 end
 
 (** Precompute a mapping from each type index to [(group_start, group_size)] for
-    its rec group. Types not in a [rec] block form singleton groups. *)
+    its rec group. Types not in a [rec] block form singleton groups.
+
+    TODO: use iarray? *)
 let compute_type_groups (type_defs : Typedef.t array) types_len =
   let groups = Array.make types_len (0, 1) in
```
redianthus, line 1245: Also, I think it should be Array.init, Array.make should be banned! 😆

src/interpret/interpret.ml
```diff
@@ -110,7 +110,14 @@ struct
...
 
-  let default_gc_val _ = assert false
+  let default_gc_val : Binary.storage_type -> Value.t = function
+    | Val_type (Num_type I32) -> I32 (I32.of_int32 0l)
+    | Val_type (Num_type I64) -> I64 (I64.of_int64 0L)
+    | Val_type (Num_type F32) -> F32 (F32.of_bits (I32.of_int32 0l))
+    | Val_type (Num_type F64) -> F64 (F64.of_bits (I64.of_int64 0L))
+    | Val_type (Num_type V128) -> V128 V128.zero
+    | Val_type (Ref_type (_, ht)) -> Ref (Ref.null ht)
+    | Pack_type _ -> I32 (I32.of_int32 0l)
```
redianthus, line 120: I think there is a similar function somewhere already used to initialize locals, but it may be of type `Binary.val_type -> Value.t`, so you could re-use it here and do something like:

```ocaml
| Val_type vt -> f vt
| Pack_type _ -> I32 ...
```

src/env/env.ml
```diff
@@ -1233,8 +1246,6 @@ module Make
       ; exported_tables
       ; exported_tags
       ; last_module
-      ; types
-      ; type_groups
```
redianthus: @hra687261, I don't understand how it can work if you don't put the new types and type_groups here?

src/env/env.ml
```diff
@@ -186,6 +198,10 @@ module Make
     ; context : Context.t
     ; raw_names : modul StringMap.t
         (* this is used only for scripts where modules can get a $id and we have to remember them to be able to register them this way... *)
+    ; types : Binary.sub_type array
+        (* table of all modules types (with the ids shifted) *)
+    ; type_groups : (int * int) array
+        (* table of all type groups (with their bound ids shifted) *)
```
redianthus: Could you make this an `Allocator.t`?

src/env/env.ml
hra687261: We do it earlier (since we need the updated env to check the types of the imports), cf.:
https://github.com/hra687261/owi/blob/7ca4dcd52a20d5de0e31c7bfd38756efc23b2f20/src/env/env.ml#L699-L706

src/env/env.ml
hra687261: We can't do that now 'cause its a single array (that needs to grow)

src/ir/binary.ml
hra687261: Why is that? (Given that the value is immutable ...)

src/env/env.ml
redianthus: Aw OK, get it, makes sense!

src/env/env.ml
redianthus: I think there's something I don't get, sorry :sweat_smile: 

When you have an `Array.t`, you can not make it grow, when you use `append` a new array is built and you have to update the IntMap anyway, right?
How would that be different with an Iarray?

src/ir/binary.ml
redianthus: That's only a nitpick, I try to avoid it as much as possible because when your type suddenly become mutable, it's very hard to debug

src/env/env.ml
hra687261: I was loosely speaking, I was referring to creating a new array with `append`. 
There is no `IntMap` anymore since we flattened it to store all the types from all the modules in one array, but you're right that the array can be immutable since we don't change it once we process all the modules. (I think we just need to rewrite the array initalization code to not use `set` on the array)

src/env/env.ml
hra687261: In the other comment we were saying that this should be an`iarray`, given that the indices go from 0 to n, it seems more suitable to use an array than a map (which is what `Allocator.t` seems to be), no?

redianthus: Yes, basically, Allocator.t handles the "shift" thing for you, when you have an allocator, it can give you what will be the next id and such, to avoid having to do the computation yourself (which usually looks like `Array.lenght old + curr_index`)

src/env/env.ml
redianthus: Yes, so I think we need a Iarray + something similar to `set` and I think that's what Allocator gives you (+ a way to handle shifting for you)

redianthus:
Let's merge and fix the 3 remaining tests later @hra687261 ! :)

PR #1005, opened 2026-06-17
add globals support in abs interpreter

S41d:


src/abstract/abstract_driver.ml
```diff
@@ -634,6 +634,30 @@ module DataAbstract_state : DATA_STATE = struct
...
+        | Some v -> v
+        | None -> (
+          let r = Link_env.get_global env i in
+          match r with
+          | Error _ -> assert false
+          | Ok v -> (
+            match v.value with
+            | I32 i -> Abstract_value.I32 (Abstract_i32.of_int32 ctx i)
+            | I64 i -> Abstract_value.I64 (Abstract_i64.of_int64 ctx i)
+            | _ -> assert false ) )
```
redianthus: Here, it might be worth to add it to the map/state, otherwise we'll recreate it many times?

redianthus:
Thanks but I suspect you also need to call the union at some point? For instance:

```wat
(if (then
  i32.const 42
  global.set 0
) (else
  i32.const 54
  global.set 0
))

global.get 0 ;; I think something is wrong here in your current implementation !
```

src/abstract/abstract_driver.ml
S41d: oh yeah forgot that

redianthus:
Sorry, I think I've found another bug. :sweat_smile: 

Because you're doing lazy initialization of the globals, if you do:

```ocaml
(it (then
  i32.const 42
  global.set 0
) (else ))

global.get 0 ;; will be {42} whereas it can be anything in case we went through the else branch!

PR #1011, opened 2026-06-23
add semi proper support for extern funcs in abstract

S41d:
I left out the handling externrefs and memory for now


src/symbolic/symbolic_extern_func.ml
```diff
@@ -2,4 +2,8 @@
 (* Copyright © 2021-2026 OCamlPro *)
 (* Written by the Owi programmers *)
 
-include Extern.Func.Make (Symbolic_value) (Symbolic_choice) (Symbolic_memory)
+include
+  Extern.Func.Make (Symbolic_value) (Symbolic_choice) (Symbolic_memory)
+    (struct
+      type t = unit
+    end)
```
redianthus: You can use the `Unit` module directly

src/ir/extern.mli
```diff
@@ -114,6 +119,9 @@ module Func : sig
   end) (Memory : sig
     (** The memory type *)
     type t
+  end) (Context : sig
+    (** The abstract context type *)
```
redianthus:
```suggestion
    (** The context type *)
```

src/concrete/concrete_extern_func.ml
```diff
@@ -2,4 +2,8 @@
 (* Copyright © 2021-2026 OCamlPro *)
 (* Written by the Owi programmers *)
 
-include Extern.Func.Make (Concrete_value) (Result) (Concrete_memory)
+include
+  Extern.Func.Make (Concrete_value) (Result) (Concrete_memory)
+    (struct
+      type t = unit
+    end)
```
redianthus: You can use the `Unit` module directly

src/symbolic/symbolic_extern_func.ml
S41d: Oh didn't know that existed

src/abstract/abstract_extern_func.ml
```diff
@@ -2,4 +2,6 @@
 (* Copyright © 2021-2026 OCamlPro *)
 (* Written by the Owi programmers *)
 
-include Extern.Func.Make (Abstract_value) (Result) (Concrete_memory)
+include
+  Extern.Func.Make (Abstract_value) (Result) (Concrete_memory)
```
redianthus: Can you define some `Abstract_memory` module? Even if everything is `assert false` inside.

Also, why are you using `Result` here? I don't think its the right choice, as there is no possible error coming from the abstract interpreter here. We may simply need `type 'a t = 'a`

src/abstract/abstract_wasm_ffi.ml
```diff
@@ -16,8 +22,9 @@ open Abstract_extern_func.Syntax
 
 let symbolic_extern_module =
   let functions =
-    [ ("i32_symbol", Extern_func (unit ^->. i32, symbol_i32))
-      (* ; ("assume", Extern_func (i32 ^->. unit, assume)) *)
+    [ ("i32_symbol", Extern_func (context ^->. i32, symbol_i32))
+    ; ("i64_symbol", Extern_func (context ^->. i64, symbol_i64))
+      (* ; ("assume", Extern_func (context ^-> i32 ^->. unit, assume)) *)
```
redianthus: I feel like `assume` should return a `context option` ? (That would be unit option in concrete/symbolic mode). What do you think?

S41d: Yeah but does that not cause typechecking errors for the wasm side ?

src/abstract/abstract_extern_func.ml
S41d: done

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -0,0 +1,291 @@
...
+    let state = eval_data state instr in
+    (Some state, None)
+  | Nop -> (Some state, None)
+  | If_else _ | Call _ | Block _ | Loop _ | Br _ | Br_if _ | Br_table _
+  | Br_on_non_null _ | Br_on_null _ ->
+    (None, Some instr)
+  | instr ->
+    Fmt.failwith "DataAbstract_state.eval_instr not implemented for %a"
+      (Binary.pp_instr ~short:true)
+      instr
```
redianthus, line 358: I don't really like the name of this file. There's something called "data" and Wasm and it's unrelated, and it's also not really a module for any kind of state, it's rather the interpreter for simple instructions, WDYT?

S41d: I agree. But i couldn't come up with a better name so I kept what we had. Is `Abstract_simple_instructions` okay for you ?

src/abstract/abstract_extern_func.ml
```diff
@@ -2,4 +2,15 @@
 (* Copyright © 2021-2026 OCamlPro *)
 (* Written by the Owi programmers *)
 
-include Extern.Func.Make (Abstract_value) (Result) (Concrete_memory)
+include
+  Extern.Func.Make
+    (Abstract_value)
+    (struct
+      type 'a t = 'a
+    end)
```
redianthus: Maybe it's time for a `Abstract_monad` module? Even if empty for now 😅

src/abstract/abstract_wasm_ffi.ml
```diff
@@ -2,22 +2,24 @@
...
+
+let assume ctx = raise (ExternFuncException (Assume ctx))
 
 let symbolic_extern_module =
   let functions =
-    [ ("i32_symbol", Extern_func (unit ^->. i32, symbol_i32))
-      (* ; ("assume", Extern_func (i32 ^->. unit, assume)) *)
+    [ ("i32_symbol", Extern_func (context ^->. i32, symbol_i32))
+    ; ("i64_symbol", Extern_func (context ^->. i64, symbol_i64))
+    ; ("assume", Extern_func (context ^-> i32 ^->. unit, assume))
```
redianthus: Hum OK, so, as we plan to use a monad later, I don't like adding Context only for the abstract world and having to fake it in others, maybe we could use the ugly exception mechanism for all of them? And make sure e.g. i32_symbol has the same type as in symbolic/concrete (i.e. unit ^-> i32) ? (sorry, I should have been clearer about suggesting to remove all of them early)

src/abstract/abstract_interpreter_simple.ml
redianthus: `Abstract_interpreter_simple` ? `Abstract_interpreter_basic` ?

src/abstract/abstract_wasm_ffi.ml
S41d: Oh okay

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -196,20 +197,78 @@ let eval_f32 ({ stack; ctx; _ } as state : Abstract_state.t) _uuid :
...
     { state with stack }
-  | Add | Sub | Mul | Div | Lt | Le | Gt | Ge ->
+  | Add | Sub | Mul | Div | Sqrt | Min | Max | Copysign ->
     let _, stack = Stack.pop2_f32 stack in
     let stack = Stack.push_f32 stack (Abstract_f32.unknown ctx) in
     { state with stack }
-  | _ -> assert false
+  | Lt | Le | Gt | Ge | Eq | Ne ->
+    let _, stack = Stack.pop2_f32 stack in
+    (* We don't have unknown for booleans :) *)
```
redianthus: Maybe we could add it then? It'll be useful I think!

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -196,20 +197,78 @@ let eval_f32 ({ stack; ctx; _ } as state : Abstract_state.t) _uuid :
   | Const f ->
     let stack = Stack.push_f32 stack (Abstract_f32.of_float32 ctx f) in
     { state with stack }
-  | Add | Sub | Mul | Div | Lt | Le | Gt | Ge ->
+  | Add | Sub | Mul | Div | Sqrt | Min | Max | Copysign ->
```
redianthus: sqrt is f32 -> f32 (it pops only one f32)

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -196,20 +197,78 @@ let eval_f32 ({ stack; ctx; _ } as state : Abstract_state.t) _uuid :
...
+    (* We don't have unknown for booleans :) *)
+    let stack = Stack.push_i32 stack (Abstract_i32.unknown ctx) in
+    { state with stack }
+  | Convert_i (_nn, _sx) ->
+    let _, stack = Stack.pop_f32 stack in
+    let v = Abstract_f32.unknown ctx in
+    let stack = Stack.push_f32 stack v in
+    { state with stack }
+  | Abs | Neg | Ceil | Floor | Trunc | Nearest | Demote_f64 ->
+    let _, stack = Stack.pop stack in
```
redianthus: Use `pop_f32` instead?

And demote_f64 should be f64 -> f32

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -196,20 +197,78 @@ let eval_f32 ({ stack; ctx; _ } as state : Abstract_state.t) _uuid :
...
     let _, stack = Stack.pop2_f32 stack in
     let stack = Stack.push_f32 stack (Abstract_f32.unknown ctx) in
     { state with stack }
-  | _ -> assert false
+  | Lt | Le | Gt | Ge | Eq | Ne ->
+    let _, stack = Stack.pop2_f32 stack in
+    (* We don't have unknown for booleans :) *)
+    let stack = Stack.push_i32 stack (Abstract_i32.unknown ctx) in
+    { state with stack }
+  | Convert_i (_nn, _sx) ->
```
redianthus: For convert_i, the type depends on `nn` see the monadic interpreter:

```ocaml
    | Convert_i (S32, S) ->
      Stack.apply_i32_f32 stack F32.convert_i32_s |> Choice.return
    | Convert_i (S32, U) ->
      Stack.apply_i32_f32 stack F32.convert_i32_u |> Choice.return
    | Convert_i (S64, S) ->
      Stack.apply_i64_f32 stack F32.convert_i64_s |> Choice.return
    | Convert_i (S64, U) ->
      Stack.apply_i64_f32 stack F32.convert_i64_u |> Choice.return
```

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -196,20 +197,78 @@ let eval_f32 ({ stack; ctx; _ } as state : Abstract_state.t) _uuid :
...
+    let stack = Stack.push_f32 stack v in
+    { state with stack }
+  | Reinterpret_i S32 ->
+    let stack = Stack.apply_i32_f32 stack (fun _ -> Abstract_f32.unknown ctx) in
+    { state with stack }
+  | Reinterpret_i S64 ->
+    let stack = Stack.apply_i64_f32 stack (fun _ -> Abstract_f32.unknown ctx) in
+    { state with stack }
+  | Load (_i, _m) ->
+    let _, stack = Stack.pop2_f32 stack in
```
redianthus: it is i32 -> f32 (you have f32 -> f32 -> f32 for now)

(the address is always an integer)

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -196,20 +197,78 @@ let eval_f32 ({ stack; ctx; _ } as state : Abstract_state.t) _uuid :
...
+    let stack = Stack.push_f32 stack (Abstract_f32.unknown ctx) in
+    { state with stack }
+  | Store (_i, _m) ->
+    let _, stack = Stack.pop2_f32 stack in
```
redianthus: It should be f32 -> i32 -> unit

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -196,20 +197,78 @@ let eval_f32 ({ stack; ctx; _ } as state : Abstract_state.t) _uuid :
...
+    { state with stack }
 
 (* TODO: handle this correctly *)
 let eval_f64 ({ stack; ctx; _ } as state : Abstract_state.t) _uuid :
   Binary.f64_instr -> _ = function
-  | Add | Sub | Mul | Div | Lt | Le | Gt | Ge ->
+  | Const _ ->
+    let stack = Stack.push_f64 stack (Abstract_f64.unknown ctx) in
+    { state with stack }
+  | Add | Sub | Mul | Div | Min | Max | Copysign | Sqrt ->
```
redianthus: move sqrt with ceil, floor etc.

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -196,20 +197,78 @@ let eval_f32 ({ stack; ctx; _ } as state : Abstract_state.t) _uuid :
...
+    let stack =
+      Stack.apply_f64_f64_f64 stack (fun _ _ -> Abstract_f64.unknown ctx)
+    in
+    { state with stack }
+  | Lt | Le | Gt | Ge | Eq | Ne ->
     let _, stack = Stack.pop2_f64 stack in
+    (* We don't have unknown for booleans :) *)
+    let stack = Stack.push_i32 stack (Abstract_i32.unknown ctx) in
```
redianthus: write and use unknown_boolean

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -196,20 +197,78 @@ let eval_f32 ({ stack; ctx; _ } as state : Abstract_state.t) _uuid :
...
+    { state with stack }
+  | Convert_i (_nn, _sx) ->
```
redianthus: type depends on nn

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -196,20 +197,78 @@ let eval_f32 ({ stack; ctx; _ } as state : Abstract_state.t) _uuid :
...
+    let _, stack = Stack.pop_f64 stack in
+    let v = Abstract_f64.unknown ctx in
+    let stack = Stack.push_f64 stack v in
+    { state with stack }
+  | Abs | Neg | Ceil | Floor | Trunc | Nearest | Promote_f32 ->
+    let _, stack = Stack.pop stack in
```
redianthus: move promote_f32 to is own branch, it has type f32 -> f64, and here you can use `pop_f32` instead of `pop`

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -196,20 +197,78 @@ let eval_f32 ({ stack; ctx; _ } as state : Abstract_state.t) _uuid :
...
+    { state with stack }
+  | Reinterpret_i S32 ->
+    let stack = Stack.apply_i32_f64 stack (fun _ -> Abstract_f64.unknown ctx) in
+    { state with stack }
+  | Reinterpret_i S64 ->
+    let stack = Stack.apply_i64_f64 stack (fun _ -> Abstract_f64.unknown ctx) in
+    { state with stack }
+  | Load (_i, _m) ->
+    let _, stack = Stack.pop2_f32 stack in
     let stack = Stack.push_f64 stack (Abstract_f64.unknown ctx) in
```
redianthus: should have type i32 -> f64

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -196,20 +197,78 @@ let eval_f32 ({ stack; ctx; _ } as state : Abstract_state.t) _uuid :
...
     { state with stack }
-  | _ -> assert false
+  | Store (_i, _m) ->
+    let _, stack = Stack.pop2_f32 stack in
```
redianthus: should have type f32 -> i32 -> unit

redianthus:
Thanks, LGTM!

PR #1035, opened 2026-07-03
Document cargo owi

krtab:


redianthus:
LGTM, but could you put this here instead: https://github.com/OCamlPro/owi/blob/main/doc/src/large/rust.md ?

redianthus:
Fix #1028

Issue #1016, opened 2026-06-29
support SIMD


redianthus:

redianthus:
A lot of it was done in #1020 because it was required. I still have to complete it.

PR #1037, opened 2026-07-03
full SIMD support (step 2)

redianthus:
Second step towards #1016


redianthus:
I'll merge this one and do the end in a follow-up PR.
@hra687261, in the last commit, I've made some changes to `Wast` and `Script`. I think it was way too complicated, and I feel like it's much better like this. I needed this to test/implement SIMD properly, because, as for floats, v128 expected results can contain nan:arith or nan:canonical. I feel like it's going to help you for the GC part also, so I'm only letting you know so you can have a look and grab the changes if this part was too annoying for you. :)

PR #1040, opened 2026-07-07
Abs checker

### LINKED ISSUE #1030: make `owi abs` display a list of alarms
It could be done this way:

1. Run the abstract interpreter and build the invariant database.
2. Pass the invariant database to an `Abstract_checker` or something, that would simply goes through the module and display on each potential bug if it was proven safe or not.

S41d:
closes #1030


src/abstract/abstract_driver.ml
```diff
@@ -173,10 +177,12 @@ module DenotFixpoint (S : DATA_STATE) = struct
     in
     fst @@ continue ctx out
 
-  let join_X (state_a, jt_a) (state_b, jt_b) =
+  let join_X (istate_a, jt_a) (istate_b, jt_b) =
```
redianthus: I'm not sure about renaming this `istate` everywhere, IMO `state` reads better 😅

src/abstract/abstract_wasm_ffi.ml
```diff
@@ -19,6 +19,10 @@ let assume condition =
     | None -> None
     | Some ctx -> Some { state with ctx } )
 
+let assert' condition = assume condition
+
+let exit _code = return ()
```
redianthus: Maybe you could bring back the module annotation to make sure you have all "required" functions? (see the leftover comment at the beginning of the file, and the annotation should be in other was_ffi files (smbolic or concrete), you can simply duplicate it here).

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -9,7 +9,9 @@ type t =
   | Unreachable
 
 let i32_can_be_zero ctx v =
-  Option.is_some @@ Abstract_domain.assume ctx (Abstract_i32.eqz ctx v)
+  match Abstract_domain.query_boolean ctx (Abstract_i32.eqz ctx v) with
+  | True | Top -> true
+  | False | Bottom -> false
```
redianthus, line 14: What is the difference? This is to avoid building a new context?

src/abstract/abstract_driver.ml
```diff
@@ -542,5 +542,8 @@ let expr (link_state : Abstract_extern_func.extern_func Link.State.t)
         | Some state, _mapping -> state )
       { state; env; envs } m.to_run
   in
+  Log.app (fun m ->
+    m "invariants : %a" Abstract_invariant.pp istate.state.invariant );
+  Abstract_checker.expr link_state m istate.state.invariant;
```
redianthus: OK so maybe it's time to move all the code in driver to `abstract_interpreter_control_flow` (or whatever) and have only the high level structure in driver as it is done in others commands? That is, call the interpreter to get the invariants, then call the checker, WDYT?

src/abstract/abstract_interpreter_simple.ml
S41d: yes, to avoid creating a copy

src/abstract/abstract_driver.ml
S41d: yeah good idea

src/abstract/abstract_checker.ml
```diff
@@ -0,0 +1,37 @@
...
+  ~(invariants : Abstract_invariant.t) ~env ~envs =
+  List.iter (check_instr ~invariants ~env ~envs) expr.raw
+
+and check_instr (instr : Binary.instr Annotated.t)
+  ~(invariants : Abstract_invariant.t) ~env ~envs =
+  if Abstract_invariant.can_divide_by_zero invariants ~uuid:instr.uuid then
+    Log.app (fun m ->
+      m "Possible division by zero for expression:(uuid: %i) %a" instr.uuid
+        (Binary.pp_instr ~short:true)
+        instr.raw );
```
redianthus: Ah yes, you should do this only on operations that might divide by zero! (div_s, div_u, rem_s, rem_u)

It'll probably better to write a small helper for each kind of error, like, `check_div_by_zero` (i.e. one helper per invariant kind), and call it on each instruction that might trigger it.

src/abstract/abstract_driver.ml
S41d: state.state just looks weird imo. but okay I'll change it back

S41d:
woops. I don't actually export the functions in ffi 😅

src/abstract/abstract_checker.ml
```diff
@@ -0,0 +1,61 @@
+(* SPDX-License-Identifier: AGPL-3.0-or-later *)
+(* Copyright © 2021-2026 OCamlPro *)
+(* Written by the Owi programmers *)
+
+let check_can_divide_by_zero invariants uuid pp instr =
+  if Abstract_invariant.can_divide_by_zero invariants ~uuid then
+    Log.app (fun m ->
+      m "Possible division by zero for expression:(uuid: %i) %a" uuid pp instr )
```
redianthus, line 8: I think it is also interesting to print something in the `else` case, to make sure we managed to prove some stuff.

What about using `Logs.warn` when something is not proven and `Logs.info` when it is proven?

src/abstract/abstract_checker.ml
```diff
@@ -0,0 +1,61 @@
...
+
+let check_i32 ~uuid ~invariants : Binary.i32_instr -> unit = function
+  | (Div_s | Div_u | Rem_s | Rem_u) as instr ->
+    check_can_divide_by_zero invariants uuid Binary.pp_i32_instr instr
+  | _ -> ()
```
redianthus: Could you make this explicit? Otherwise we might forget to update the code here, if a new instruction that might divide by zero is added. (same for all other functions)

src/abstract/abstract_checker.ml
```diff
@@ -0,0 +1,61 @@
...
+    | Wasm { func; idx } ->
+      let env = Dynarray.get envs idx in
+      check_expr ~invariants ~env ~envs func.body
+    | Extern _ -> ()
+    end
+  | I32 i32_instr -> check_i32 ~uuid:instr.uuid ~invariants i32_instr
+  | I64 i64_instr -> check_i64 ~uuid:instr.uuid ~invariants i64_instr
+  | F32 f32_instr -> check_f32 ~uuid:instr.uuid ~invariants f32_instr
+  | F64 f64_instr -> check_f64 ~uuid:instr.uuid ~invariants f64_instr
+  | _ -> ()
```
redianthus: Can you also make this one explicit?

src/abstract/abstract_driver.ml
```diff
@@ -2,525 +2,11 @@
...
-        (* TODO handle this properly *)
-        match ConcreteFixpoint.eval_expr ~env ~envs state e with
-        | None, _mapping -> state
-        | Some state, _mapping -> state )
-      state m.to_run
-  in
-
-  state.invariant
+  let abs_state = Abstract_state.empty env () in
+  eval_exprs m.to_run abs_state env envs
```
redianthus: I would remove the `open` and simply qualify `Abstract_interpreter_control_flow.eval_exprs` here, this is used only once so it's not that annoying having to write it and it'll make things easier to follow.

src/abstract/abstract_checker.ml
```diff
@@ -0,0 +1,61 @@
+(* SPDX-License-Identifier: AGPL-3.0-or-later *)
```
redianthus, line 1: Adding a `.mli` for this module would be nice too

src/abstract/abstract_wasm_ffi.ml
```diff
@@ -8,27 +8,93 @@ open Abstract_monad
...
    displayed *)
-let symbol_i32 () = fold_state (fun { ctx; _ } -> Abstract_i32.unknown ctx)
+module M :
+  Wasm_ffi_intf.S0
+    with type 'a t := 'a Abstract_monad.t
+     and type memory := Abstract_memory.t
+     and type i32 := Abstract_i32.t
+     and type i64 := Abstract_i64.t
+     and type f32 := Abstract_f32.t
+     and type f64 := Abstract_f64.t = struct
```
redianthus: Thanks a lot for making this change everywhere in the codebase!

src/abstract/abstract_wasm_ffi.ml
```diff
@@ -8,27 +8,93 @@ open Abstract_monad
...
-  map_state (fun ({ ctx; _ } as state) ->
-    let condition = Abstract_i32.to_boolean ctx condition in
-    match Abstract_domain.assume ctx condition with
-    | None -> None
-    | Some ctx -> Some { state with ctx } )
+  let assert' condition = assume condition
 
+  let exit _code = return ()
+
+  let abort () = return ()
```
redianthus: Is this correct? I expect `exit` and `abort` to return bottom or something like this? That is, after executing one of them, nothing should be analyzed further.

src/abstract/abstract_wasm_ffi.ml
```diff
@@ -8,27 +8,93 @@ open Abstract_monad
...
+
+  let symbol_i32 () = fold_state (fun { ctx; _ } -> Abstract_i32.unknown ctx)
+
+  let symbol_i64 () = fold_state (fun { ctx; _ } -> Abstract_i64.unknown ctx)
+
+  let symbol_f32 () = fold_state (fun { ctx; _ } -> Abstract_f32.unknown ctx)
+
+  let symbol_f64 () = fold_state (fun { ctx; _ } -> Abstract_f64.unknown ctx)
+
+  let symbol_v128 () = assert false
```
redianthus:
```suggestion
  let symbol_v128 () =
    (* TODO *)
    assert false
```

src/abstract/abstract_wasm_ffi.ml
```diff
@@ -8,27 +8,93 @@ open Abstract_monad
...
+    let* gt =
+      fold_state (fun { ctx; _ } ->
+        Abstract_i32.gt_s ctx v lo |> Abstract_i32.of_boolean ctx )
+    in
+    let* lt =
+      fold_state (fun { ctx; _ } ->
+        Abstract_i32.lt_s ctx v hi |> Abstract_i32.of_boolean ctx )
+    in
+    let* () = assume gt in
+    let+ () = assume lt in
```
redianthus, line 59: It's quite nice to be able to write it like this! :D

src/abstract/abstract_wasm_ffi.ml
```diff
@@ -8,27 +8,93 @@ open Abstract_monad
...
+  let symbol_range lo hi =
+    let* v = fold_state (fun { ctx; _ } -> Abstract_i32.unknown ctx) in
+    let* gt =
+      fold_state (fun { ctx; _ } ->
+        Abstract_i32.gt_s ctx v lo |> Abstract_i32.of_boolean ctx )
+    in
+    let* lt =
+      fold_state (fun { ctx; _ } ->
+        Abstract_i32.lt_s ctx v hi |> Abstract_i32.of_boolean ctx )
+    in
```
redianthus, line 58: Maybe define it (`lt`) after `assume gt` check, so it's not created if not needed?

src/abstract/abstract_wasm_ffi.ml
```diff
@@ -8,27 +8,93 @@ open Abstract_monad
...
+    let* () = assume gt in
+    let+ () = assume lt in
+    v
+
+  let alloc _m _base _size =
+    Fmt.epr "TODO : implement alloc";
```
redianthus:
```suggestion
    Log.warn (fun m -> m "TODO : implement alloc");
```

src/abstract/abstract_wasm_ffi.ml
```diff
@@ -8,27 +8,93 @@ open Abstract_monad
...
+    fold_state (fun { ctx; _ } -> Abstract_i32.unknown ctx)
+
+  let free _m _ptr =
+    Fmt.epr "TODO : implement free";
```
redianthus:
```suggestion
    Log.warn (fun m -> m "TODO : implement free");
```

src/cmd/cmd_abs.ml
```diff
@@ -13,10 +13,8 @@ let cmd ~source_file ~unsafe =
   let+ m, link_state =
     Compile.File.until_link ~unsafe ~name:None link_state source_file
   in
-  let (_abstract_invariant : Abstract_invariant.t) =
-    Abstract_driver.expr link_state m
-  in
-  ()
+  let state = Abstract_driver.expr link_state m in
+  Abstract_checker.check_module link_state m state.invariant
```
redianthus, line 17: I expect `abstract_driver` to take care of most of this (i.e. calling abstract_checker), it should be something like:

```ocaml
  Abstract_driver.expr link_state m
```

And then in `driver.ml`:

```ocaml
  let expr link_state m =
     ...
     let state = ... in
     Abstract_checker.check_module link_state m state.invariant
```

src/cmd/cmd_sym.ml
```diff
@@ -24,7 +24,8 @@ let run_file ~parameters ~source_file =
 
   let* abstract_invariant =
     if parameters.generate_abstract_invariant then
-      Cmd_abs.from_binary m ~unsafe:true
+      let+ state = Cmd_abs.from_binary m ~unsafe:true in
```
redianthus: Ah, yes, I called `cmd_abs` as a hack here, it should probably not be the case anymore, sorry about that (but now I understand your change in cmd_abs)

src/abstract/abstract_wasm_ffi.ml
S41d: I fixed it right at the same time as you were reviewing. it's set to `map_state(fun _ -> None)` now

redianthus:
Thanks!

PR #1044, opened 2026-07-17
Some fixes and tests for the abstract interpreter

S41d:


test/cram/abs/invariant_gen1.t
```diff
@@ -5,3 +5,2098 @@
   }
   owi: [ERROR] Reached problem!
   [13]
+  $ owi abs ./invariant_gen1.wat -vv
```
redianthus: The output is too big 😅 Can you remove the `-vv`?

S41d: Oh yeah i was testing stuff I forgot to remove this

S41d:
There seems to be a problem. I don't know if it comes from br_table or it's some other existing problem that resurfaced here. I'll add a simpler test for br_table next week, it'll be more clear

S41d:
The problem was our handling of `br 0` in blocks 😅. It seems okay now

test/cram/abs/invariant_gen1.t
```diff
@@ -5,2098 +5,5 @@
...
-                 locals : i32 [--..--];
-                 i32 [--..--];
-                 i32 [--..--];
-                 i32 {30};
-                 i32 {0; 1};
-                 i32 [0x2710..0x7FFFFFFF];
-                 i32 {0x2710}
-  
-  owi: [DEBUG] jt            :  
-  owi: [INFO] Passed division by zero check for expression:(uuid: 24) i32.div_s
```
redianthus: Hum sorry, maybe put back `-v` only? (and not `-vv`) It's nice to see what was proven, WDYT?

PR #1054, opened 2026-07-24
add non recursif function call support in abs

S41d:


test/cram/abs/call.t/run.t
```diff
@@ -0,0 +1,300 @@
...
+  owi: [INFO] running instr : drop
+  owi: [DEBUG] jt after (drop) :  
+  owi: [DEBUG] after call(start): callee state : 
+                 context: Context{id=1, <empty>}
+                 stack  : 
+                 locals : 
+  
+  owi: [DEBUG] jt after (call 1) :  
+
+  $ owi abs rec.wat -vv
```
redianthus:
```suggestion
  $ owi abs rec.wat -v
```

test/cram/abs/call.t/run.t
```diff
@@ -0,0 +1,300 @@
+  $ owi abs non_rec.wat -vv
```
redianthus:
```suggestion
  $ owi abs non_rec.wat -v
```

src/abstract/abstract_interpreter_control_flow.ml
```diff
@@ -311,8 +311,12 @@ module DenotFixpoint (S : DATA_STATE) = struct
     in
     loop state JumpTarget.empty expr.raw
 
-  and eval_func ({ abs_state; _ } as state : interpreter_state)
+  and eval_func ({ abs_state; _ } as state : interpreter_state) idx
     (func : Binary.Func.t) =
+    if List.mem idx abs_state.call_stack then begin
+      Log.err (fun m -> m "Recursive function call. Unsupported");
+      exit 1
```
redianthus: Maybe raise an exception and catch it at the driver (or whatever now that driver does not exist anymore) level ?
So that, we can still use `--generate-abstract-invariant` on symbolic even on function with recursive functions. When you catch the exception, you can simply return an empty abstract_invariant. WDYT? (Ideally, that would not be an exception and we would use a proper error monad, but as this is only temporary, I think it'll do the trick).

S41d: hmm okay

PR #1056, opened 2026-07-25
More stuff for abs script

S41d:
With this, I can successfully run `test/cram/script/passing/localfun.wast` and `test/cram/script/passing/42.wast`. There's a lot of tests with recursive functions and/or `assert_trap`s so we can't run them yet


redianthus:
@S41d, it seems a rebase is needed here.

PR #1058, opened 2026-07-27
new linker implementation

redianthus:


redianthus:
Three things for the future:

- return the env in case of failure, this will allow to pass some script tests related to linking, this is very annoying to do... we don't really care about analyzing things after a failure, so this does not really matter for now
- bring back support for alloc/free and double-free, heap overflow etc. this can actually be done by storing the chunks in the symbolic monad instead of inside the memory, then, the symbolic_memory can simply update the monadic state, and the external functions do not have to return the runtime env, they can simply modify the monadic state! this shouldn't be too hard to do, but left for later as this PR is already way too big
- fix globals in the abstract interpreter, I removed the fold and now the results are wrong but it should be easy to fix

PR #1061, opened 2026-08-04
Progress bar 

### LINKED ISSUE #458: add an option to display progress
for instance, the number of current tasks in the wq and the number of finished tasks?

### LINKED ISSUE #575: make a shiny terminal progress interface
something like AFL does

ArkadySkv:
# Add `--progress` flag with dynamic progress bar for symbolic execution

This PR adds a `--progress` flag to the `owi sym` command, displaying a live‑updating progress bar during symbolic execution. The progress bar shows the number of completed tasks, total tasks, and a visual progress indicator.

## OCaml Version
- **OCaml**: 5.4.0 (same as project's main branch)
- **Dependencies**: No new dependencies **added**

## Motivation

Currently, there is no way to see the progress of symbolic execution without enabling `--debug`, which floods the output and significantly slows down execution. This feature provides a lightweight, user‑friendly progress indicator that updates in place.

Fixes: #458, #575

### New Module: `src/infra/progress.ml`
- Implements a dynamic progress bar with:
  - `#`‑filled bar showing completion percentage
  - Task counters (`completed / total`)
  - Carriage return for in‑place updates
  - Fallback to plain logging when stdout is not a TTY
  - "Waiting for tasks..." message when total is unknown

### Modified Files

| File | Changes |
|------|---------|
| `src/bin/owi.ml` | Added `--progress` CLI flag, included in `symbolic_parameters` |
| `src/symbolic/symbolic_parameters.ml` | Added `progress : bool` field |
| `src/symbolic/symbolic_parameters.mli` | Added `progress : bool` to interface |
| `src/symbolic/symbolic_driver.ml` | Integrated progress reporter with execution loop (task counting, reporting, finishing) |
| `src/symbolic/symbolic_driver.mli` | Added `progress:bool` to `run` signature |
| `src/cmd/cmd_sym.ml` | Pass `~progress` from parameters to driver |
| `src/cmd/cmd_iso.ml` | Pass `~progress` from parameters to driver |
| `src/cmd/cmd_iso.mli` | Added `progress:bool` to `cmd` signature |
| `src/owi.mli` | Updated `Symbolic_parameters` record definition |
| `src/dune` | Added `progress.ml` to build system |

### Test File: `example.wat`

```wat
(module
  (func $classify (param $x i32) (result i32)
    (if (i32.eq (local.get $x) (i32.const 0))
      (then (return (i32.const 42))))
    (if (i32.eq (local.get $x) (i32.const 1))
      (then (return (i32.const 99))))
    (return (i32.const 0)))
  (export "classify" (func $classify)))
```
Command & Output
```bash
./_build/default/src/bin/owi.exe sym --progress --invoke-with-symbols example.wat --entry-point classify
```

The progress bar updates as symbolic execution explores each path:

```text
[###################                                         ]  33% (1/3 tasks)
[########################                                    ]  40% (2/5 tasks)
[####################################                        ]  60% (3/5 tasks)

All OK!
```
Expected Behavior

TTY output: Live‑updating progress bar with # characters, percentage, and task counts.

Non‑TTY output (redirected to file or pipe): Plain log messages like Progress: 3 / 5 tasks completed.

Unknown total: Shows "Waiting for tasks..." until the first task is counted.

Usage
```bash
owi sym --progress [OTHER_OPTIONS] FILE
```
Examples
```bash
# Symbolic execution with progress bar
owi sym --progress example.wat

# With symbolic arguments
owi sym --progress --invoke-with-symbols example.wat --entry-point classify

# With verbose logging (progress bar still works)
owi sym --progress --verbosity=debug example.wat
```

Implementation Details
The progress bar is implemented using:

Atomic counters for thread‑safe updates across workers.

Logs.app for output (follows Owi's logging system).

Carriage return (\r) for in‑place terminal updates.

Unix.isatty to detect TTY vs. non‑TTY output.

Tasks are counted at two points:

Total tasks incremented when a new task is pushed (Yield).

Completed tasks incremented when a task finishes (Ok or Error).

The reporter is enabled only when --progress is passed and output is a TTY.

Future Work
Add --progress support to other subcommands (owi c, owi c++, owi rust, owi go, owi zig).

Add customizable progress bar style (e.g., spinner for unknown total).

Related Issues
Closes #458 – add an option to display progress

Closes #575 – make a shiny terminal progress interface (partial)

Checklist
☑ Build passes (dune build @all)
☑ Tests pass (dune runtest) – except pre‑existing failures
☑ Code follows Owi style and uses the prelude library
☑ Documentation updated (usage examples)
☑ --progress flag appears in --help

### DIFF
--- a/src/infra/progress.ml
+++ b/src/infra/progress.ml
@@ -1,0 +1,63 @@
+(* SPDX-License-Identifier: AGPL-3.0-or-later *)
+(* Copyright © 2021-2026 OCamlPro *)
+(* Written by the Owi programmers *)
+
+type t = {
+  total : int Atomic.t;
+  completed : int Atomic.t;
+  mutable enabled : bool;
+}
+
+let create () =
+  { total = Atomic.make 0; completed = Atomic.make 0; enabled = false }
+
+let enable t = t.enabled <- true
+
+let increment_total t =
+  Atomic.incr t.total
+
+let increment_completed t =
+  Atomic.incr t.completed
+
+let get_total t =
+  Atomic.get t.total
+
+let get_completed t =
+  Atomic.get t.completed
+
+let bar_width = 80
+
+let render_bar ~total ~completed =
+  let width = bar_width - 20 in
+  let percent =
+    if total = 0 then 0.0
+    else (float_of_int completed /. float_of_int total) *. 100.0
+  in
+  let filled = int_of_float ((percent /. 100.0) *. float_of_int width) in
+  let bar =
+    Fmt.str "%s%s" (String.make filled '#') (String.make (width - filled) ' ')
+  in
+  Logs.app (fun m ->
+      m "\r[%s] %3.0f%% (%d/%d tasks)" bar percent completed total)
+
+let report t =
+  if t.enabled then
+    let total = get_total t in
+    let completed = get_completed t in
+    if total = 0 then
+      Logs.app (fun m -> m "Waiting for tasks...")
+    else if Unix.isatty Unix.stdout then
+      render_bar ~total ~completed
+    else
+      Logs.app (fun m -> m "Progress: %d / %d tasks completed" completed total)
+
+let finish t =
+  if t.enabled then
+    let total = get_total t in
+    let completed = get_completed t in
+    if Unix.isatty Unix.stdout then (
+      (* The final bar was already printed by the last call to report.
+         Just move to a new line so subsequent logs appear cleanly. *)
+      Logs.app (fun m -> m "")
+    ) else
+      Logs.app (fun m -> m "Done: %d / %d tasks completed" completed total)
--- a/src/symbolic/symbolic_driver.ml
+++ b/src/symbolic/symbolic_driver.ml
@@ -55,7 +55,7 @@ let compute_number_of_workers workers =
 let run ~exploration_strategy ~workers ~no_worker_isolation ~no_stop_at_failure
   ~no_value ~no_assert_failure_expression_printing ~deterministic_result_order
   ~fail_mode ~workspace ~seed ~solver ~model_format ~model_out_file
-  ~with_breadcrumbs ~run_time (to_run : unit Symbolic_choice.t) =
+  ~with_breadcrumbs ~progress ~run_time (to_run : unit Symbolic_choice.t) =
   (* Various initializations *)
   let bug_stack = Bugs.make () in
   let path_count = Atomic.make 0 in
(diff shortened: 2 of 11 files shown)


redianthus:
Hi,

Thanks for contributing to the project.
There are way too many useless information in the PR description (e.g. `OCaml: 5.4.0 (same as project's main branch)`, why should I care about that ?!).
Can you rewrite it (yourself, not something generated) so I can understand what was actually done?

I also encourage you to join [Owi's Zulip](https://owi.zulipchat.com) to discuss your various PRs.

src/infra/progress.ml
```diff
@@ -0,0 +1,63 @@
...
+  let percent =
+    if total = 0 then 0.0
+    else (float_of_int completed /. float_of_int total) *. 100.0
+  in
+  let filled = int_of_float ((percent /. 100.0) *. float_of_int width) in
+  let bar =
+    Fmt.str "%s%s" (String.make filled '#') (String.make (width - filled) ' ')
+  in
+  Logs.app (fun m ->
+      m "\r[%s] %3.0f%% (%d/%d tasks)" bar percent completed total)
```
redianthus, line 41: Isn't this spamming the terminal way too much? Shouldn't it be printed in another way?

src/symbolic/symbolic_driver.ml
```diff
@@ -100,15 +113,26 @@ let run ~exploration_strategy ~workers ~no_worker_isolation ~no_stop_at_failure
...
-        at_bug bug
-      | Yield (prio, f) -> write_back (prio, f)
+        (match progress_reporter with
+          | Some p -> Progress.increment_completed p; Progress.report p
+          | None -> ());
+          at_bug bug
+      | Yield (prio, f) ->
+        (match progress_reporter with
+          | Some p -> Progress.increment_total p
+          | None -> ());
```
redianthus, line 131: Why are you incrementing on Yield? What do tasks exactly correspond to?

src/symbolic/symbolic_driver.ml
```diff
@@ -100,15 +113,26 @@ let run ~exploration_strategy ~workers ~no_worker_isolation ~no_stop_at_failure
...
+          write_back (prio, f)
       | Choice (m1, m2) ->
-        at_schedulable m1 write_back;
-        at_schedulable m2 write_back
+          at_schedulable m1 write_back;
+          at_schedulable m2 write_back
```
redianthus, line 135: Why aren't you incrementing here?

src/infra/progress.ml
ArkadySkv: The current implementation uses Logs.app with a carriage return (\r) to update the same line in place. In a TTY, this does not spam the terminal – it overwrites the previous line. The progress bar updates only when a task completes (or when a new task is pushed), which is typically a few times per second at most. If the rate is too high, we could throttle updates (update every 10 completed tasks). Another alternative would be to use a terminal library like progress or fancy-progress to handle rendering more elegantly. I'm happy to adjust the approach if you have a preferred method.

src/symbolic/symbolic_driver.ml
ArkadySkv: Yield corresponds to a task being suspended and re‑scheduled. In the work‑stealing scheduler, each Yield pushes a new work item (task) onto the queue. The total task count is the number of such items that will be processed. We increment on Yield to reflect that a new task is enqueued. The completed tasks are counted when a task finishes (Ok or Error). This provides a rough measure of progress.

src/symbolic/symbolic_driver.ml
ArkadySkv: Choice represents a branch in the symbolic execution tree. It does not enqueue a new task; it creates two alternatives that are explored separately. Since the total number of tasks is not known in advance, we chose not to count Choice as a separate task.    instead, the two branches are explored as part of the same scheduling process. If we counted Choice as a task, we would need to also count it as "completed" when one branch finishes, which would be more complex. Our approach treats the work queue as the source of truth for tasks.

PR #1060, opened 2026-08-03
Update testsuite subrepo (2026-08-03)

hra687261:
Automated update check for `test/cram/script/reference/`.

PR #1064, opened 2026-08-18
Bump smtml version to 0.30.0, Frama-C to 33.0 & fix CI

hra687261:


hra687261:
Merging since the CI passes, fixes some testsuite errors I missed in https://github.com/OCamlPro/owi/pull/1060

PR #1067, opened 2026-08-20
Abs gui

S41d:
For now it's only for the abstract interpreter but it can easily be done for concrete too, not so sure about symbolic though.
Right now we show the state of the context, the call stack, the stack, locals, and globals (which aren't yet exported from the ocaml side) after each instruction. 

There are special cases for showing joins and widenings, like so:
image

And as you asked, I also added syntax highlighting and support for showing warnings.
image

With the gui, I found two bugs in loops. First one is fixed in this pr, I'm gonna leave the second one for later


trace-viewer/prism/prism.js
redianthus, line 1: Could you use an URL in the HTML file so that the JS/CSS files for Prism are downloaded by the user's browser and not stored in the repo?

trace-viewer/app.js
redianthus, line 1: I feel like this (and everything in `trace-viewer`) could be its own project, would you be willing to move it to https://github.com/OCamlPro/owi-trace-viewer ?

dune-project
```diff
@@ -79,6 +79,8 @@
    (>= 1.0.3))
   (xmlm
    (>= 1.4.0))
+  (yojson
+   (>= 2.0.0))
```
redianthus, line 83: Could you add yojson to shell.nix too?

src/abstract/abstract_interpreter_simple.ml
```diff
@@ -13,9 +13,15 @@ let i32_can_be_zero ctx v =
   | True | Top -> true
   | False | Bottom -> false
 
+let add_divide_by_zero_invariant invariant ~uuid ~possible =
+  Abstract_invariant.add_divide_by_zero_invariant invariant ~uuid ~possible;
+  if possible then
+    Abstract_trace.record_wasm_warning ~instr_id:uuid
+      ~message:"Possible division by zero"
```
redianthus, line 20: Can you move the `Abstract_trace.record_wasm_warning` call to `Abstract_invariant.add_divide_by_zero_invariant` and remove this function?

PR #1070, opened 2026-08-27
Fix remaining WasmGC tests

hra687261:


src/env/env_linker.ml
```diff
@@ -243,13 +247,23 @@ end = struct
...
-        (address, table) :: link_state.tables
+        match init with
+        | None -> Ok table
+        | Some expr ->
+          let* r =
+            Constexpr_eval.ref_expr ctx ~get_const_type ~get_const_func
+              ~get_const_global expr.Annotated.raw
+          in
+          let len = Int32.of_int (Table.size table) in
+          Ok (Table.fill table 0l len r)
```
redianthus: Indeed, that was a bug, thanks!
But I think it should be solved differently.
I'm trying to remove the `Constexpr_eval` as much as possible.
The good news is that the right fix is quite easy to do !

You should actually add the code to initialize the table in `link_state.initialization_code`, this is some code that is run to initialize a module.

You should do something like:

```ocaml
initialization_code @ expr.Annotated.raw @ [
  Binary.(Simple (I32 (Const 0l);
  Binary.(Table Fill)
]
```

(this is wrong and just some pseudo-code so you get the idea)

One thing to be careful about is that you should put this at the end of the initialization code !

hra687261: Alright, done in: https://github.com/OCamlPro/owi/pull/1070/changes/6c040bccf54fe71ae0473bffcf3a924fdd0136c5, seems to work.

> One thing to be careful about is that you should put this at the end of the initialization code !

What do you mean? Are you referring to the order of the instructions in the stack being inverted?

redianthus: Yes, what you did is correct.

PR #1071, opened 2026-08-31
Update testsuite subrepo (2026-08-31)

hra687261:
Automated update check for `test/cram/script/reference/`.

### DIFF
--- a/src/abstract/abstract_checker.ml
+++ b/src/abstract/abstract_checker.ml
@@ -11,8 +11,8 @@ let check_can_divide_by_zero invariants uuid pp instr =
       m "Passed division by zero check for expression:(uuid: %i) %a" uuid pp
         instr )
 
-let check_i32 ~uuid ~invariants : Binary.i32_instr -> unit = function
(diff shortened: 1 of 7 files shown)


redianthus:
@hra687261, I think it failed because it was before I merged #1070. It should pass now, can you rebase so we can check it's OK?

hra687261:
That's probably it. 
Rebasing caused the PR to be closed, the only difference was in the `update-testsuite.yml` (but I haven't changed it in a while, maybe someone else did?), so the testsuite was apparently not updated and I assume that since there were no changes with main, the PR was closed ...

redianthus:
Yes, I don't really understand why it was closed and I can't reopen it :sweat_smile: We can simply wait next week that it gets updated again?

hra687261:
From what I can tell, the testuite was not updated, but in that case the PR should not have been created in the first place ... unless I messed up somewhere during the rebase. 
Yes, we'll see next week, if there was an update, we should get it then anyway.

PR #1072, opened 2026-08-31
Fix `script abstract`

S41d:
This is the same as #1056. I'll close that one


src/abstract/abstract_interpreter_control_flow.ml
```diff
@@ -274,7 +276,10 @@ module DenotFixpoint (S : module type of Abstract_interpreter_simple) = struct
 
   and eval_func ({ abs_state; _ } as state : Abstract_interpreter_state.t) idx
     (func : Binary.Func.t) =
-    if List.mem idx abs_state.call_stack then raise RecursiveFunctionCall;
+    let nb_recursive_calls =
+      List.length (List.filter (( = ) idx) abs_state.call_stack)
```
redianthus: With a `List.fold_left` you can avoid traversing the list twice

src/abstract/abstract_stack.ml
```diff
@@ -122,8 +122,7 @@ let pop_bool s ctx =
   | Value.I32 n -> (Value.I32.to_boolean ctx n, tl)
   | _ -> assert false
 
-let pop_n s n =
-  (List.filteri (fun i _hd -> i < n) s, List.filteri (fun i _hd -> i >= n) s)
+let pop_n s n = (List.take n s, List.drop n s)
```
redianthus, line 125: Nice! Could you update the code in datastructure/stack.ml too if it does the same?

src/abstract/abstract_i64.ml
```diff
@@ -80,7 +80,9 @@ let gt_s ctx x1 x2 = lt_s ctx x2 x1
 let gt_u ctx x1 x2 = lt_u ctx x2 x1
 
 let shl ctx x1 x2 =
-  let flags = Operator.Flags.Bshl.pack ~nsw:true ~nuw:true in
+  let mask = of_int64 ctx 63L in
+  let x2 = and_ ctx x2 mask in
```
redianthus, line 84: Is this necessary?

S41d: yes, it didn't work without it. 

for the following example : 
```wast
(module
  (func (export "shl") (param $x i32) (param $y i32) (result i32) (i32.shl (local.get $x) (local.get $y)))
)
(assert_return (invoke "shl" (i32.const 1) (i32.const 32)) (i32.const 1))
```
I was getting 0 instead of 1

redianthus:
Thanks!

PR #1081, opened 2026-09-02
reorganize subcommands

redianthus:


redianthus:
The Nix build is failing but this is only because of TinyGo currently being broken in Nix, should be fixed soon.

PR #1092, opened 2026-09-14
Update testsuite subrepo (2026-09-14)

hra687261:
Automated update check for `test/cram/script/reference/`.


redianthus:
I think the error is only due to a new version of the Rust compiler or Z3, I promoted the tests in #1094, if you rebase it should be fine.

hra687261:
Yeah, that seems to be the case, the rebase should do it

redianthus:
Yes, looks like your fix is right! :) A `dune runtest`/`dune promote` is missing and that should be good to go then

hra687261:
Yes, I though the test failure was introduced after the rebase, apparently there was one before (the CI and the testsuite take too long to run tho :/ do you think something can be done about that?)

redianthus:
Yes, I can probably try to update some stuff in Nix so we can benefit from the cache again and go back to 4 minutes of CI, but I need a new release of Codex first and I'm not sure when that'll happen (hopefully quite soon)

hra687261:
Merging, the other CIs are apparently failing due to change in the generated counter example

